PR #7, opened 2026-08-20
Add sender alias selection

atomashevic:
Adds a From picker backed by Gmail send-as aliases. New messages use the default alias, while replies prefer the address that received the original message.

image

Tested with `make validate`.


huacnlee:
Thank you.

PR #10, opened 2026-08-20
Add Alt-A to open Switch accounts

### LINKED ISSUE #9: Switch accounts from the keyboard with Alt+1…9
The account switcher already works: the sidebar address and
**Switch account...** both call `switchToIndex`. There is no key
that does the same thing.

**Proposal**

`Alt+1` through `Alt+9` select the mailbox in that position — the
same order the switcher already lists. A number past the last
account does nothing.

These are modified keys, so they can live in `list` and `reader`
without colliding with a draft or a query being typed. They do not
collide with anything in `keys/Keymap.js` today.

The help sheet should show one row, not nine: `Alt+1…9` /
`Switch account`.

**Not this**

Opening the switcher from the keyboard. The numbered keys are the
switch, the same way `g,i` is the inbox rather than a mailbox
picker.

serg3k:
## Summary

- Closes #9. `Alt+1` through `Alt+9` pick the mailbox in that position — the same order the switcher already lists. A number past the last account does nothing, and the keys stay out of a draft or a search field.
- The help sheet shows one row (`Alt+1…9` / Switch account) and scrolls, so the new line is reachable. Docs no longer list multiple accounts as future work.
- `restore.sh` is the inverse of `install.sh`: it puts the marketplace clone back after a development symlink.

## Test plan

- [ ] Two or more accounts: `Alt+1` / `Alt+2` land on the same mailboxes, in the same order, as the sidebar switcher
- [ ] `Alt+9` with fewer than nine accounts does nothing
- [ ] Typing in a reply or the search field: `Alt+2` does not switch
- [ ] `?` shows **Going → Alt+1…9 / Switch account**, and the sheet scrolls if the window is short
- [ ] After a switch, `j`/`k` still move in the new mailbox

Made with [Cursor](https://cursor.com)


serg3k:
@huacnlee this is ready for a review when you have a moment — I have been using it against a couple of mailboxes and it matches the switcher order.
What could be also added is reordering accounts in Settings. I would keep it separate though

huacnlee:
Thanks for this — the keyboard part was clean, and `Instantiator` + a table-driven `display` was the right instinct. I've pushed four commits onto the branch rather than send it back with a list; here is what changed and why.

**`Alt+1…9` → `Alt+A`, opening the switcher.** Nine top-level keys is more keyboard than switching mailboxes is worth — it isn't a frequent action, and it spent a whole namespace on it. `Alt+A` opens the account switcher instead, and the switcher is now a list the keyboard walks: `j`/`k` move and wrap, `Enter` or `o` takes one, `Esc` leaves. The cursor rests on the mailbox you're already in, so the first `j` is one step away from it. Modified rather than bare, because the bare letters are the scarce thing here — `a` on its own still means reply-all, with a test that says so.

That took the machinery the nine keys needed with it: `triggered` carries an id again and not a sequence, `display` is gone, and so is `accountIndexFor` and its second copy of the `Alt+` prefix.

**The switcher answers its own keys, and that is worth reading before you touch it.** The first attempt routed `j`/`k` through `KeyRouter` with `survivesOverlay`, per the rule in AGENTS.md. It didn't work, and the reason inverts the rule: an open `QQC.Popup` takes every key before the shortcut map sees it — `focus` true or false, bare or modified — so inside a popup it's the `KeyRouter` binding that looks live and never runs. `tests/qml/tst_popup_keys.qml` asserts both halves, and AGENTS.md and `docs/KEYS.md` now carry the exception beside the rule it breaks. The only actual decision in it — that this cursor wraps where the message list clamps — is `Model.wrappedIndex`, where the node tests reach it.

**Two things the scrolling `ShortcutHelp` introduced.** An interactive `Flickable` accepts the press itself, so the dismiss `MouseArea` underneath stopped seeing clicks on the sheet — in exactly the tall list the `Flickable` was added for, and the file's own comment promises "or a click puts it away". A `TapHandler` puts it back; a flick scrolls and doesn't close. And it scrolled by mouse only, on the one screen in the window that *is* the keyboard reference — it has a scrollbar now, and `j`/`k` move it.

**`restore.sh` is out of this branch.** It's a good script and it doesn't belong in a keyboard-shortcut PR: it moves directories around under a user's `~/.config`, restarts the shell and re-enables the plugin, and nobody reviewing "add a key" should be approving that by the way. Please do send it on its own — it'll get read properly there.

Two notes that aren't yours to fix. `make test-shell` is red on this branch because `main` carries a 283 KB `preview.png` that trips `test_source.sh`'s own 128 KB ceiling; that's a `main` problem and I'm fixing it separately. And `switchToIndex` counts half-configured account rows, so "the second mailbox" can be one still being set up — pre-existing, same as the switcher menu, so I left it alone.

PR #11, opened 2026-08-21
Answer an invitation, and get off a list, without leaving the window

huacnlee:
A Google Calendar invitation is not a special kind of mail: it is an ordinary
one with a `text/calendar` part beside the HTML. So this reads the part rather
than the sender, and works for Outlook, Fastmail and a self-hosted CalDAV
server without knowing that any of them exist.

Answering one is a mail too — an RFC 5546 REPLY addressed to the organiser,
carrying the same UID with this account's ATTENDEE line rewritten. That is the
whole protocol, and it is why RSVP needs **no calendar API, no second OAuth
scope, and works the same on IMAP as on Gmail**.

## The times are the part worth being careful about

There is no timezone database here, so a `TZID` is resolved through the
`VTIMEZONE` the sender ships in the same file — which is what the standard asks
a sender to include for exactly this reason. The yearly rules are read, the
year's transitions computed, and the one behind the meeting's wall clock wins.

January is EST because the transition that put it there was in the *previous*
November. A search inside the event's own year is an hour wrong every winter,
so the search covers the years either side; there is a test for both directions.

A zone that arrives with no VTIMEZONE is **not converted at all**: the
organiser's clock is kept and the zone named beside it. A conversion nothing
backs is worse than an honest `(Antarctica/Troll)`.

## Unsubscribing

Three different amounts of work wearing one word, and `Unsubscribe.plan` picks
between them so no view branches on a header:

| Sender offers | What happens | Label |
|---|---|---|
| `List-Unsubscribe-Post` + an https URL | one POST, never leaves the window | `Unsubscribe` |
| a `mailto:` | a message to the address they nominated | `Unsubscribe` |
| anything else | their page opens in a browser | `Unsubscribe...` |

RFC 8058 is the fast path: a sender that sends `List-Unsubscribe-Post` has
promised one POST is enough. **Without that line a POST would be a guess at
somebody else's API**, so the honest thing is to open the page — with a label
that says so before it is pressed.

Whether a URL may be fetched at all is decided by `Html.imageSourceKind`, the
same judgement that decides whether a message may load a picture, for the same
reason. Loopback, private ranges, link-local, single-label and `.local` names
are refused; one-click is https only, because the request itself carries the
fact that this address reads this mail.

## Also here

- `buildRawMessage` grows a calendar part: `multipart/alternative`, so a client
  that understands the reply does not also show it as a file to open.
- The invitation and the unsubscribe offer are kept in the body cache beside
  the body. Neither changes once fetched, so a message opened before shows its
  card at the same moment it shows its text.
- An answer that is sent is written back into the cached invitation. The
  organiser's document still says `NEEDS-ACTION`, and a message reopened
  tomorrow would otherwise show its buttons unanswered — after the answer had
  been sent, and had worked.
- The two reader notices become one `ReaderNotice`. Three can be up at once,
  and each anchoring to whichever of the others happened to be visible is a
  rule that has to be rewritten every time a fourth arrives.
- Four glyphs on the existing 16px grid: `calendar`, `video`, `pin`, `people`.
  `IconTextButton` gains a `selected` state — never the only sign of one, so
  the chosen answer also carries a check.

## Verification

`make test-js` (13 files, two of them new), `test_qml_names.py`,
`test_qml_text_format.py`, `test_service_source.sh`, `test_install.sh` and
`test_transport.sh` all pass.

**Not verified here:** `make qml-check` and `make test-qml` need the Qt the
plugin runs on, which the machine this was written on does not have. Please run
`make validate` on Arch before merging.

`tests/test_source.sh` fails on `preview.png` at 283 KB against its 128 KB
ceiling. **That is on `main` already**, from 8e5bd8d, and is left alone —
verified by stashing this branch. Which of the two gives way is not this
branch's call.


huacnlee:
Rebased onto `main` after #7 (sender alias selection) landed. The overlap was real rather than textual — that PR rewrote `buildRawMessage`'s From line and `send()`, which are both on this path — so the two are now composed rather than merely merged:

- `buildRawMessage` keeps #7's `fromHeader(values.from, values.fromName)` and gains the calendar part beside it.
- **An RSVP now answers as the address the invitation arrived at**, not the account's primary one. A mailbox with aliases is invited as one of them and the ATTENDEE line names that one — so looking for an existing answer under the primary address finds nothing, and sending one from it answers for somebody the organiser never invited. `receivedAsAddress` resolves it through `Api.preferredSendAs` over the message's To and Cc, which is #7's own rule read the other way round.
- The `mailto:` unsubscribe path sends from the same address, for the same reason: a list that only ever knew an alias has no reason to honour a request from anywhere else.
- `summarize` gaining `cc` in #7 is what makes that lookup complete.

#7's four new `test_source.sh` rules were checked by hand, because the `preview.png` size rule aborts the script before it reaches them — all four still hold. Everything else is green; `make validate` on Arch is still the outstanding check.

PR #20, opened 2026-08-21
Leave the noisy categories out of Unread instead of asking for Primary

### LINKED ISSUE #19: Unread is empty when Gmail category tabs are off
The Unread mailbox (and the badge, which uses the same query) is

```
in:inbox is:unread category:primary
```

That is Gmail's Primary *tab*, not "unread mail in the inbox".
The comment in `providers/Gmail.js` says why: category tabs leave
the INBOX label on Promotions and the rest, so a plain
`in:inbox is:unread` came back as newsletters.

Accounts that do not use those tabs never get `CATEGORY_PERSONAL`.
Google Workspace often has them disabled; a personal account on
Important / Unread / Starred first is the same. Inbox still shows
unread rows. Unread and the badge stay at zero.

**Proposal**

Keep Primary only when the account actually categorises mail that
way. If a Primary unread search returns nothing and the inbox has
unread mail, fall back to `in:inbox is:unread`. Or drop the
category scope and let people who want Primary type it in search.

IMAP is already `folder:INBOX UNSEEN` and is not this bug.

serg3k:
## Summary

Closes #19. The Unread mailbox and the badge asked for `in:inbox is:unread category:primary`, and on a real account that matched nothing while the inbox held 201 unread messages. They now ask for `in:inbox is:unread -category:promotions -category:social -category:forums`: the same mail, named by what it leaves out.

Primary is the one category whose search word does not name its label. The other four read `promotions`, `social`, `forums` and `updates` against `CATEGORY_PROMOTIONS` and the rest; Primary's label is `CATEGORY_PERSONAL`, and `primary` is the web UI's name for the tab. The API's `q` promises only "most of" the web search syntax and has never documented `category:` at all, so that one word was the one with nothing holding it up.

## Why a negation rather than a second probe

Which way the query fails is the whole argument. A positive scope that stops matching empties the mailbox and the badge — unread mail piling up with nothing left to say so. A negation that stops matching widens to every unread message in the inbox: noisier, and nothing worse than noisier. That covers the account whose owner turned Smart features off, where Gmail stops applying the category labels entirely and there is no Primary left to ask for, without the app having to detect that it happened.

Turning the category tabs off was never the cause. That setting is display only — Gmail goes on applying the labels and `category:` goes on matching — so a probe built to detect "this account has no tabs" would have been reading a distinction that is not there.

And counting could not have carried that judgement anyway: `Primary = 0, unscoped > 0` reads the same whether the account does not categorise at all or Primary is simply clear and the backlog sits in Promotions. Falling back on the second case moves the badge to exactly the mail the scope exists to keep off it, on a mailbox somebody keeps up with, every time Primary reaches zero.

Updates stays in. Receipts, deliveries and GitHub's notifications land there, and they are mail somebody is waiting on.

## Invariants

- One query per unread count, as before. `refreshCounts` and `effectiveQuery` are untouched, so the cache key does not move and the badge cannot flip between two definitions of "unread".
- IMAP and HEY are unchanged: `folder:INBOX UNSEEN` and `box:imbox unseen`.
- The provider still owns what "unread" means; the account does not branch on a provider id.

## Verification

- `make validate` — node tests, source regressions, `qmllint`, `omarchy plugin validate`.
- `tests/test_provider.js` pins the new query, that it is not a positive category filter, and that Updates is not excluded.
- Not yet run against a live mailbox. Worth confirming once on the account that reported `0`: `in:inbox is:unread category:personal` and the four `category:` counts — non-zero `personal` confirms `primary` was the failing word, all four at zero says the account is not categorising at all. The change is correct either way; the counts only say which of the two it was.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,14 +222,7 @@ and the client itself are console-only; there is no CLI for them.
 | `F5` | Check for mail |
 | `Ctrl+?` | Every shortcut |
 
-Search takes Gmail's own operator syntax straight through — `from:jane`,
-`has:attachment`, `older_than:7d`. The Unread mailbox prefers Primary, because
-category tabs do not remove the `INBOX` label and an unscoped unread search
-comes back as the promotional backlog. Accounts that do not use those tabs
-(Workspace, or an inbox type without Primary) never get that label, so Unread
-falls back to `in:inbox is:unread` when Primary is empty and the inbox is not.
-Right-click any row in the list for archive,
-trash, spam, star and read/unread without leaving the keyboard cursor behind.
+Search takes Gmail's own operator syntax straight through — `from:jane`, `has:attachment`, `older_than:7d`. The Unread mailbox leaves Promotions, Social and Forums out rather than asking for Primary: Gmail's categories do not remove the `INBOX` label, so an unread filter without that exclusion comes back as the whole promotional backlog rather than the mail you have not read — while one that asks for Primary comes back empty on any account where Gmail is not applying the category labels, which is unread mail with nothing left to say so. Updates stays in, because receipts, deliveries and notifications land there. Right-click any row in the list for archive, trash, spam, star and read/unread without leaving the keyboard cursor behind.
 
 ## What it does not do
 
--- a/README.md
+++ b/README.md
@@ -223,10 +223,12 @@ and the client itself are console-only; there is no CLI for them.
 | `Ctrl+?` | Every shortcut |
 
 Search takes Gmail's own operator syntax straight through — `from:jane`,
-`has:attachment`, `older_than:7d`. The Unread mailbox is scoped to Primary:
-category tabs do not remove the `INBOX` label, so an unread filter without that
-scope returns the whole promotional backlog rather than the mail you have not
-read. Right-click any row in the list for archive,
+`has:attachment`, `older_than:7d`. The Unread mailbox prefers Primary, because
+category tabs do not remove the `INBOX` label and an unscoped unread search
+comes back as the promotional backlog. Accounts that do not use those tabs
+(Workspace, or an inbox type without Primary) never get that label, so Unread
+falls back to `in:inbox is:unread` when Primary is empty and the inbox is not.
+Right-click any row in the list for archive,
 trash, spam, star and read/unread without leaving the keyboard cursor behind.
 
 ## What it does not do
--- a/account/MailAccount.qml
+++ b/account/MailAccount.qml
@@ -308,8 +308,17 @@ Item {
   // The provider decides what a mailbox and a typed search amount to: Gmail's
   // are search operators, IMAP's name a folder. Opaque from here on — it is
   // handed back to the client that produced it, and used as a cache key.
-  readonly property string effectiveQuery: rawQuery !== "" ? rawQuery
-    : Provider.query(providerId, mailboxKey, searchQuery, defaultQuery)
+  // After a Primary unread probe comes back empty, this holds the inbox
+  // unread query instead. Empty means the provider's own unread query.
+  property string unreadQueryUsed: ""
+  readonly property string effectiveQuery: {
+    if (rawQuery !== "") return rawQuery
+    var text = String(searchQuery || "").replace(/^\s+|\s+$/g, "")
+    if (text !== "")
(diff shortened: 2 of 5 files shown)


serg3k:
@huacnlee ready for a human review — this is the empty Unread mailbox on accounts that do not use Gmail's category tabs (#19).

huacnlee:
The change reads well — the provider decides whether a second probe is worth making, and the account never branches on a provider id. What I would like to settle before merging is the premise.

#19 says accounts without category tabs never get `CATEGORY_PERSONAL`. On the maintainer's machine **both** Gmail accounts have that label, including a personal one that certainly has tabs on, and both still come back empty:

```
in:inbox is:unread category:primary   ->  0
in:inbox is:unread                    ->  201
```

Two readings, wanting opposite fixes:

- **The API does not honour `category:primary`.** `primary` is the web UI's name for the tab; the label is `CATEGORY_PERSONAL`. Then the fix is one word, not a second probe per poll.
- **Primary really is empty there.** The 201 would be GitHub notifications and newsletters, which land in Updates and Promotions. Then `0` is the correct answer, and falling back moves the badge to exactly the backlog the Primary scope was added to keep off it.

A minute in the app's search box settles it, since that goes down the same API path as the badge:

```
in:inbox is:unread category:primary
in:inbox is:unread category:personal
```

Worth knowing either way: `Primary = 0 and unscoped > 0` also matches "tabs on, Primary clear, unread all in Promotions", which `unreadQueryAfterCounts` cannot tell apart from "no tabs" — so on a mailbox somebody keeps up with, the badge would flip to the promotional backlog exactly when Primary reaches zero. The test plan covers tabs-with-unread and no-tabs, but not that one.

serg3k:
I do not think `category:personal` solves this. It is still Gmail’s Primary/Personal category, not “all unread inbox mail”, so it skips unread distribution lists, newsletters, Updates, Promotions, Forums, and the rest.

The dangerous case is:

```text
in:inbox is:unread category:primary -> 0
in:inbox is:unread                  -> non-zero
```

That can mean either:

1. category tabs are off, so Unread should be all unread inbox mail; or
2. category tabs are on, Primary is clear, and the unread mail is in Promotions/Updates/etc.

Those need opposite behavior, and the counts alone cannot distinguish them. I think the second case should be treated as a real zero. The app should not silently fall back to all inbox unread there.

So the safer fix is probably an explicit Gmail account option: “Unread shows Primary only” vs “Unread shows all unread Inbox mail”. Default can remain Primary-only to preserve the existing noise-filtering behavior, while Workspace/no-tabs accounts can opt into all unread inbox mail.

huacnlee:
Thanks for chasing this down — the symptom you found is real, and the digging settled the cause. I've pushed a different fix to this branch; the title and description are re-derived from it.

What turned up:

- **Turning the category tabs off is display only.** Gmail goes on applying the `CATEGORY_*` labels and `category:` goes on matching, so a probe built to detect "this account has no tabs" would be reading a distinction that is not there. What does stop the labels being applied is Smart features being switched off, which is a different setting and not what #19 describes.
- **Primary is the one category whose search word does not name its label.** The other four read `promotions`, `social`, `forums`, `updates` against `CATEGORY_PROMOTIONS` and the rest; Primary's label is `CATEGORY_PERSONAL`, and `primary` is the web UI's name for the tab. The API's `q` promises only "most of" the web syntax and has never documented `category:` at all — so that one word was the one with nothing holding it up.

So the scope is now named by what it leaves out:

```
in:inbox is:unread -category:promotions -category:social -category:forums
```

The reason it is a negation rather than a narrower scope is which way it fails. A positive scope that stops matching empties the mailbox and the badge — unread mail piling up with nothing left to say so. A negation that stops matching widens to every unread message in the inbox: noisier, and nothing worse. That covers the Smart-features-off account too, without the app having to detect that it happened.

You were right that counting cannot tell "this account does not categorise" apart from "Primary is clear and the backlog is in Promotions" — that is exactly why I did not want the fallback, and it is also why the setting you suggested is not needed: with the query correct there are no longer two behaviours to choose between. One request per count, as before, and `refreshCounts` is untouched.

Updates stays in — receipts, deliveries and GitHub's notifications land there.

`make validate` passes. It has not been run against a live mailbox yet; if you have the account from #19 to hand, `in:inbox is:unread category:personal` plus the four `category:` counts would say which of the two it was, though the change holds either way.

PR #21, opened 2026-08-21
app: Persist resizable mail panes

pixelpush-io:
## Summary

- Adds draggable resizing for the mailbox sidebar in the main mail window.
- Persists both sidebar and message-list widths in the existing `window.json` preferences path.
- Clamps pane widths against the current window so the list and reader keep usable minimum sizes.
- Makes the list/reader splitter easier to discover with hover feedback while preserving double-click reset.

## Notes

The service already owns window preferences such as sidebar collapsed state, body zoom, plain text, and remote-image policy. This extends that same state surface with `sidebarWidth` and `listWidth`, keeping `0` as the default/proportional sentinel so existing installs continue to use the current layout until the user drags a divider.

The sidebar handle overlays the rail edge instead of taking layout space away from the message list. The existing list/reader handle still occupies its hit lane so the reader toolbar boundary continues to meet the visible divider.

## Verification

- `make validate`

### DIFF
--- a/App.qml
+++ b/App.qml
@@ -75,8 +75,11 @@ Item {
   // service holds it because it is written to disk: a size somebody reached for
   // is theirs until they change it, not until they close the window.
   readonly property real bodyZoom: service ? service.bodyZoom : 1.0
-  // 0 means "proportional"; anything else is a width somebody dragged to.
-  property real listWidth: 0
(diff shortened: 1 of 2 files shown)


huacnlee:
Close no response PR, if you want continue, reopen again.

PR #25, opened 2026-08-22
Never adopt a legacy token that could belong to another mailbox

tmn73:
mise ~/.config/mise/config.toml tools: gh@2.97.0
## What happens today

Two mailboxes, one Cloud OAuth client, which `Credentials.sharedClient` makes the
normal case: the second account borrows the client the first one set up. Switch to
the first account and it shows the second account's inbox, under its own name, and
its cache fills with the other mailbox's mail.

The keyring is not at fault. It ends up holding one refresh token under both keys:

```
account=one@gmail.com     -> 9755eb23fbb06764
account=two@company.com   -> 9755eb23fbb06764
```

## Why

`secret-tool` matches any item carrying *at least* the attributes it was asked for,
so leaving `account` out is a wildcard over it rather than a filter.
`legacyKeyringAttributes` therefore matches the account-keyed entries as well as the
nameless one it exists for, and `mayAdoptLegacyToken` narrows the claim to the first
account without preventing it. `tests/test_credentials.js` already notes that an
empty attribute value is a wildcard; an omitted one is too.

Nothing fails loudly from there. The stage 1 read hands back the sibling's token,
the refresh succeeds, `OAuth.parseTokenResponse` carries the same refresh token
through because Google returned no new one, and `acceptToken` writes it back under
*this* account's key. Both mailboxes now hold one token.

The trigger only has to happen once: an `invalid_grant` clears the exact key, and
the next restore finds nothing of its own and falls through to stage 1.

**Which entry the wildcard answers with is not specified**, which is worth stating
because it makes this intermittent rather than reliable. Against a throwaway service
name, with a legacy entry and a named one both present, `secret-tool lookup` returned
the *legacy* one under both insertion orders — while the install this was found on
returned the sibling's. The shape that is reliably wrong is the third one: with only
a named entry left, the legacy read has nothing but another mailbox's token to find,
and takes it.

| the keyring holds | the wildcard lookup answers with | counted | after this change |
|---|---|---|---|
| one nameless entry | its own token | 1 / 1 / 0 | adopted |
| a nameless and a named one | either — unspecified | 2 / 2 / 1 | refused |
| one named entry | **another mailbox's token** | 1 / 1 / 1 | refused |

## The change

The legacy read looks before it leaps. `secret-tool search --all` on the same
attributes reports what is really there, and the token is asked for only when that
report is one match with no named account in it. Everything else refuses.

The stand-in an account with no address yet is stored under stays adoptable: it has
an `account` attribute without having a name, and a lone one means there is no other
entry on that client to take, which with `mayAdoptLegacyToken` keeps the claim to
the first account. Nothing writes such an entry any more, since `storeRefreshToken`
holds an unnamed token in memory until the profile read names it, so this is for
installs from before that hold.

Two details that shaped the code:

- The report crosses two streams. `secret-tool` prints the record with `g_print` and
  its attributes with `g_printerr`, buffered independently, so which record an
  attribute line belongs to is not answerable from outside the process. Counting is,
  and the decision needs no more. It fails closed twice over: a stderr that does not
  account for every match names nothing, because an attribute that was not read is
  not an attribute that is absent, and a search that did not exit cleanly is refused.
- The counting is line by line rather than over a collected buffer. A search keyed on
  the client alone loads every matching mailbox's token and stdout carries all of
  them, so nothing here holds a secret it has no use for.

The exact lookups are untouched, on `secret-tool lookup` and its first line.

## Verified

`make test-js`, `make test-shell`, `make qml-check` and `omarchy plugin validate .`
all pass. `make test-qml` fails on this machine on a clean `main` too, so it is an
environment problem rather than this diff.

Against a real two-account keyring, counted with the same predicates the QML uses:
`matches 2, attributed 2, named 2`, refused. Before the change the same keyring
handed back a token.

Reviewed against a throwaway service name rather than the real entries, which
confirmed the three things this rests on: the wildcard really does match a named
entry, `search --all` really does put records on stdout and attributes on stderr
with the two streams unaligned, and a `SplitParser` on **stderr** really does
deliver lines — that last one because if it did not, `attributed` would stay 0,
`attributed !== matches` would refuse every shape including the lone nameless one,
and the migration this exists to protect would fail silently for everybody upgrading
from the single-account version.

## If a mailbox is already affected

This stops the adoption; it does not undo one that already happened. An install
where two accounts already share a token keeps both keyring entries pointing at it
after this merges, so the symptom — one account showing the other's inbox — survives
the upgrade and will look like the fix did not work.

The remedy is per account and manual: sign out of the mailbox showing the wrong mail
and sign in again. That writes a token of its own under its own key, and with this
change nothing will hand it a sibling's again.

## Related, deliberately not in this diff

Both are the same wildcard, elsewhere, and both need a call on migration semantics
rather than a mechanical fix:

- At stage 3, `renamedAttributesToClear` gets `renamedLegacyKeyringAttributes`, with
  no `account`. That `secret-tool clear` is a wildcard as well, so it would remove
  every `omarchy-gmail` entry for the client, including a second account's that has
  not migrated yet.
- Adopting at stage 1 never clears what it adopted, and `clearStoredToken` only
  clears the exact account key, so on a single-account install signing out and
  restarting re-adopts the legacy entry and signs the user back in.


huacnlee:
Thank you.

PR #33, opened 2026-08-23
Ask a folder for a window of itself rather than for all of it

kbyjoel:
Hi,

This PR comes out of a problem I ran into syncing a mailbox with more than 9k messages: the Inbox would only ever show the curl error `a value or data field grew larger than allowed`, while Unread and the other folders kept working.

Full disclosure: this PR was entirely generated with Claude Code. I have read through it, but I have not tested it against a real account. I am posting it in the hope that the code is sound, or that it at least helps toward a fix.

-----

An IMAP account whose INBOX holds more than about ten thousand messages opened the Inbox on "A value or data field grew larger than allowed" and no mail at all, while Unread, Flagged and search kept working on the same account.

That sentence is libcurl's. curl is the transport, and it assembles one ping-pong response line in a 64 KiB buffer (`DYN_PINGPPONG_CMD` in lib/curlx/dynbuf.h); a line that outgrows it fails the whole request with CURLE_TOO_LARGE, exit code 100. A server answers SEARCH on a single untagged line however many messages matched, so `UID SEARCH ALL` against a large mailbox is exactly that line: ten thousand UIDs at six octets each is already past the buffer. `responseError` maps curl's documented exit codes and passes anything else through, so what reached the panel was curl's own words. Unread, Flagged and a text search escaped it only because few messages match them.

The invariant is that nothing may ask a folder a question whose answer is unbounded. `searchPlan` cuts the mailbox into windows of message sequence numbers, which RFC 9051 6.4.9 makes expressible before a single UID is known: a bare sequence set inside UID SEARCH selects by sequence number while the numbers that come back are still UIDs. Windows hold 4096 messages, because a UID is at most ten digits and a separator, so 4096 of them behind "* SEARCH " is 45065 octets.

Without criteria the folder's own order is the answer, so one window is the page and nothing but the page crosses the wire — where a listing used to download every identifier in the mailbox to display twenty-five of them. With criteria the server has to be asked, so the mailbox is walked whole, one window at a time and all of them on one connection; the windows arrive as several SEARCH lines and `parseSearch` already read every one of them as a single answer.

The newest window ends at "*" rather than at the count STATUS reported, so mail delivered between the two commands is in the list rather than past the end of it, and the page is taken from the front of what came back rather than assumed to be all of it.

What replaces it: the size of a folder stops being something the client derives by downloading every UID in it and becomes something it asks for. A listing therefore costs one `STATUS` before the search — one command against a mailbox nothing has selected, which is why it is its own request rather than a section of the next one. That is one extra connection per page load, paid back immediately on any folder large enough for the old listing to be slow and on every folder large enough for it to fail.

Rejected: raising curl's buffer, which is a compile-time constant in libcurl and not reachable from a config file; and windowing only the criteria-free listing, which leaves Unread and Flagged one large mailbox away from the same error.

Verification: `make test-js` and `make test-shell` pass, with new cases in tests/test_imap.js covering the newest page, a later page, the oldest page, a page past the end, an empty folder, a folder smaller than one window, and the multi-window ladder a criteria search produces. `make qml-check` exits 0 with the same three pre-existing warnings on ImapClient.qml as before the change, and `omarchy plugin validate .` passes. `make test-qml` could not be run here: qmltestrunner is not installed on this machine. Not verified against a mailbox of the failing size; the reproduction is a single curl request,
`curl --url imaps://host/INBOX --user ... --request "UID SEARCH ALL"`, which exits 100 on such an account and exits 0 for every window this change asks for instead.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -250,6 +250,14 @@ key. What matters while working:
   the end of one response into the middle of the next — which is also how a
   message body could forge a response of its own. Base64 keeps one character
   per octet, so counting characters is counting octets.
+- **A response line has a ceiling, and it is curl's.** libcurl assembles one
+  ping-pong line in a 64 KiB buffer (`DYN_PINGPPONG_CMD`) and abandons the whole
+  request with `CURLE_TOO_LARGE` when a line outgrows it — which `UID SEARCH
+  ALL` does on any mailbox past about ten thousand messages, because a server
+  answers SEARCH on one line however many matched. Nothing may ask a folder a
+  question whose answer is unbounded: `searchPlan` cuts the mailbox into windows
+  of message sequence numbers first, which is what a listing pays a `STATUS`
+  for.
 - `BODY.PEEK`, never `BODY`. Reading a list must not mark the mailbox seen, and
   that is the most common way a hand-rolled IMAP client ruins a mailbox.
 - `UID EXPUNGE`, never bare `EXPUNGE`: the latter removes every `\Deleted`
--- a/providers/ImapClient.qml
+++ b/providers/ImapClient.qml
@@ -200,9 +200,12 @@ Item {
 
   // ---------------------------------------------------------------- reads
 
-  // IMAP has no page token: SEARCH answers with every matching UID at once. So
-  // the "token" is an offset into that answer, and the count it reports is
-  // exact rather than the estimate Gmail returns.
+  // IMAP has no page token, so the "token" is an offset and the page is a
+  // window of the folder rather than a slice of a reply that carried all of it.
+  // How big the folder is decides what may be asked of it — a SEARCH the size
+  // of a mailbox is a response line curl refuses to hold — so the size is asked
+  // for first. STATUS is one command against a mailbox nothing has selected,
+  // which is why it is its own request rather than a section of the next one.
   //
   // Newest first, which is the order the list is read in and the reverse of the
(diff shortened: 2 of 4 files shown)


huacnlee (changes requested):
Thanks for working on this. Limiting each SEARCH response is the right direction, but the current sequence-number windows can return incorrect pages.

1. A filtered search uses several commands such as `UID SEARCH 1:4096 UNSEEN` and then `UID SEARCH 4097:8192 UNSEEN`. If another client expunges a message between these commands, later sequence numbers move down. The fixed next window can then skip or duplicate a message. IMAP allows EXPUNGE responses during UID SEARCH, but this code does not adjust the remaining windows.

2. A non-first page is calculated from the earlier STATUS count. For example, count 1000 and offset 25 produces `951:975`. If a new message arrives before SEARCH, the correct second page is now `952:976`, so the current command misses one message and includes one older message. Only the newest page uses `*`; older pages are still shifted.

Please change the paging design so its boundaries remain correct when messages are added or expunged between STATUS and SEARCH commands, and add tests for both mailbox changes.

huacnlee:
Thanks again for identifying and investigating the large-mailbox curl failure. This has now been resolved by #36, which uses a stable UID snapshot and bounded UID search ranges to avoid both the response-line limit and the sequence-number races noted in review. Closing this PR as superseded by #36.

PR #34, opened 2026-08-24
Add account-scoped calendars, reliable Google sessions, delayed sending, contacts, attachments, and bar previews

eliasfaltin:
## Summary

- Add native week and month calendars for Google Calendar and CalDAV, including source management, event details, recurrence, event creation, keyboard navigation, and themed colors.
- Scope Google Calendar display and event creation to the active mail account while retaining shared CalDAV calendars.
- Cache account-scoped calendar ranges, retain cached events during source failures, and show actionable Google Calendar setup errors.
- Place Calendar at the bottom of the mailbox sidebar and preserve the current mailbox when switching accounts.
- Recover saved Google sessions after transient refresh failures and retain the Calendar grant after reauthorization.
- Add configurable delayed sending without blocking mailbox navigation or account switching. Ctrl+Z and the toast button undo the pending send.
- Add recipient suggestions, attachment-preserving forwards, and clickable received attachments for Gmail and IMAP.
- Move focus from the compose subject to the body when Tab is pressed.
- Add bar previews for unread mail and upcoming events while restoring bar-icon window toggling.
- Improve rich-message sizing, table layout, image dimensions, and system-theme integration.

## Architecture and invariants

Calendar providers normalize Google JSON and CalDAV iCalendar data into one event model. Google calendars reuse Gmail OAuth. CalDAV credentials remain in GNOME Keyring.

The main calendar follows the active mail account. Its Google source and event composer are account-scoped. CalDAV sources remain shared because they have no mail-account owner.

The bar preview aggregates enabled sources in a separate controller and cache. Calendar caches are bounded, private, atomic, and keyed by account context and date range.

Calendar grid borders and Today styling come from reactive Omarchy theme tokens passed down by `App.qml`. Today uses the quieter selected accent fill, with bold date text as a non-color state indicator.

Google request deadlines abort through one completion path. Event creation cannot remain busy forever. Transient token-refresh failures retain and retry the saved session.

Refresh tokens use a versioned Calendar grant marker. Existing Gmail-only grants request incremental consent instead of appearing signed in and failing Calendar requests with 403. The keyring writer removes the matching account's unversioned item before storing the current grant because libsecret otherwise replaces its secret without adopting the new attribute.

Inline VTIMEZONE data remains authoritative. A TZID without its definition preserves the organizer's wall clock, stays marked unresolved, and no longer depends on the host timezone.

A queued send remains owned by its originating mailbox. The service keeps that mailbox reachable while other navigation continues. Undo switches back before restoring the parked draft.

Received attachments use the provider's existing fetch path. The opener validates base64, removes path traversal, writes a private 0600 runtime file, and calls xdg-open without a shell.

## Validation

`make validate`

- JavaScript, Python, and shell tests pass, including OAuth recovery, versioned keyring storage, attachment opening, delayed-send routing, and Calendar theme regressions.
- 63 QML tests pass with zero failures.
- Calendar parsing passes under Europe/Stockholm, Asia/Singapore, and UTC.
- `qmllint`, `omarchy plugin validate .`, and `git diff --check` complete successfully.


huacnlee (dismissed):
Please fix these issues before merging:\n\n1. The Google calendar GET timeout completes the request twice. abort() already triggers the DONE handler, which calls failSource() and advances the source queue. The timer then calls failSource() again, so it can skip a calendar or clear a new request. Please let one path finish the request.\n\n2. Google event creation has no timeout. If the POST never receives a response, creatingEvent stays true forever and the user cannot create another event.\n\n3. Switching accounts during the undo-send delay hides the Undo action, but the old account timer still sends the message. Please keep the pending send reachable, or block/handle account switching until the delay ends.\n\n4. Existing Gmail refresh tokens do not have the new Calendar scope. Refresh still marks the account signed in, but calendar requests fail with 403 and there is no re-consent path. Please handle the scope upgrade for existing accounts.\n\n5. Calendar timezone parsing depends on the machine timezone when TZID is present without VTIMEZONE. This makes events appear at the wrong instant and makes the test fail outside Europe/Stockholm. For example, TZ=Asia/Singapore node tests/test_calendar_feed.js fails by 6 hours.\n\nAlso, please rename the PR. The project does not use scope prefixes such as [codex], and the title should name the machinery and outcome.

huacnlee (approved):
The requested fixes are included in 840471f. Full validation passes, including 54 QML tests.

eliasfaltin:
Follow-up complete in 17daf80 and 620039e. Queued sends no longer own the keyboard or block account switching. Ctrl+Z remains globally reachable from mail and Calendar. The update also adds subject-to-body Tab focus, clickable Gmail and IMAP attachments, bounded Google requests, Calendar scope migration, deterministic unresolved TZID handling, and upstream transient session recovery. Full validation passes with 63 QML tests.

PR #38, opened 2026-08-25
Compose from a mailto, pick From across mailboxes, and attach files

eliasfaltin:
## What changed

- **Compose from mailto links.** Omamail handles `mailto:` links from the desktop and message bodies. The desktop handler summons the existing window instead of toggling it, and To, Cc, Bcc, subject, and body fill the new draft.
- **Choose From across connected mailboxes.** The From menu combines every signed-in account with its usable send-as aliases and switches the sending account without discarding the draft. Replies and forwards remain on the mailbox that received the original.
- **Use Bcc in every draft.** Bcc can be exposed beside Cc and uses the same recipient suggestions and submission path.
- **Attach files through picker, drop, or paste.** Compose accepts an out-of-process file picker, drag and drop, copied files, clipboard images, and clipboard PDFs. The helper enforces a 20 MiB per-file limit and cleans up clipboard-owned temporary files.
- **Send attachments through each provider's real interface.** Outgoing files become base64 MIME parts for Gmail and IMAP, while HEY receives paths through repeatable `--attach` arguments. Attachment-only messages and forwarded attachments are supported.
- **Register the desktop mailto handler.** First installation claims the default; later starts refresh the desktop entry without replacing a default the user chose afterward.
- **Restore an open window after service restart.** Window-open state persists and the service asks the shell to reopen it after restarting.

## Release Notes

- Open Omamail compose from desktop and in-message `mailto:` links.
- Choose the sending address from any connected mailbox without losing the draft.
- Use Bcc with recipient suggestions in every draft.
- Attach files by choosing, dropping, or pasting them, including screenshots and PDFs.
- Send attachment-only messages and preserve forwarded attachments through Gmail, IMAP, and HEY.
- Register Omamail as the desktop mailto handler without overriding a later user choice.
- Restore the Omamail window automatically when its service restarts while the window was open.

## Architecture and invariants

- `message/Mailto.js` is the only parser for external mailto input. Header fields lose CR and LF before they can become message headers; the body retains line breaks.
- `compose/Senders.js` owns the provider-neutral rules for combining accounts and aliases. Provider differences remain below `MailAccount`.
- Attachment bytes cross the helper boundary as one JSON response and are embedded in the raw MIME message. Only HEY receives attachment paths because its CLI owns the final composition request.
- A new message may switch sending accounts, while a reply or forward may not cross the account boundary of its original message.
- Mailto registration is refreshed on startup, but an existing user-selected default handler is not overwritten.

## Verification

- `make validate`
- Mailto parser and desktop-handler tests, including percent decoding, header newline removal, and summon semantics
- Sender aggregation and account-switching tests
- Attachment helper tests for file reads, clipboard images and PDFs, copied file URIs, picker cancellation, size limits, JSON escaping, and owned-file cleanup
- MIME and HEY command tests for attachments and Bcc
- Offscreen compose test confirming the From menu opens and lists identities
- Manual verification in the running Omarchy shell for `xdg-open mailto:you@example.com`, paste, drop, and Attach


huacnlee:
Next time, please 1 PR just do 1 thing, to help me to review. Thanks.

PR #39, opened 2026-08-25
Read Fastmail through its JMAP API

### LINKED ISSUE #37: Read Fastmail over JMAP rather than IMAP
Fastmail is already reachable here — `Imap.js` names it in the provider summary, and it works. But IMAP is a compatibility layer over Fastmail's real data model, and the provider descriptor has to say so: `threads: false`, `labels: false`, `web: false`. Those three are false because IMAP cannot express what the server actually holds, not because Fastmail lacks them.

JMAP is that data model directly. I would like to add it as a fourth provider, and I am offering to do the work.

## What it changes

| | `imap` today | a `jmap` provider |
| --- | --- | --- |
| `threads` | `false` — reconstructed from `References` | `true` — server-assigned `threadId` |
| `labels` | `false` — one folder per message | `true` — `mailboxIds` is a set |
| `web` / `webBox` | `false` — no address to open | `true` — Fastmail messages and searches have URLs |
| `search` | `TEXT` substring scan | structured server-side filters on Fastmail's index |
| `batch` | sequential `STORE` / `COPY` / `EXPUNGE` | one `Email/set` carrying every id |
| `spam` | `false` | `false` — unchanged, there is still no junk verb worth honouring |

Every one of those switches already exists in `Registry.capabilities`. They are simply unlit for IMAP.

## What I measured first

Against a live Fastmail account with a read-only API token, before proposing anything:

One POST carried `Mailbox/get`, then `Email/query`, then `Email/get` consuming the query by back-reference, then `Thread/get` consuming *that* by a nested back-reference on `/list/*/threadId`. Four method calls, thirty envelopes with thread ids, 12KB, a single round trip.

`Email/query` reported `canCalculateChanges: true`, so `Email/changes` drives incremental sync rather than polling and diffing.

Two findings that shape the design:

The session's `apiUrl` was `https://phl.api.fastmail.com/jmap/api/`, not `api.fastmail.com` — and `downloadUrl`, `uploadUrl` and `eventSourceUrl` are region-scoped the same way. A provider that hardcodes the host past the initial session request would work for some accounts and fail for others.

A token scoped to read-only mail advertises only `urn:ietf:params:jmap:core` and `urn:ietf:params:jmap:mail`. `urn:ietf:params:jmap:submission` is absent, so such an account cannot send at all.

## Five things I would rather you decide than guess

**Is it `jmap` or `fastmail`?** The registry holds two named services and one generic protocol. I lean `jmap`, ordered after the named services and before `imap`, with Fastmail as the headline in `SUMMARY` — mirroring how `Imap.js` already names Fastmail without being called Fastmail. That leaves Stalwart and Cyrus reachable later. The cost is that `webMessageUrl` then has to recognise a Fastmail session rather than simply knowing it is one.

**Static capabilities, or narrowed from the session?** `Registry.define` reads capabilities from a static descriptor, but JMAP publishes them per session — that is much of what makes it a protocol rather than a service. `Imap.js` already has the shape of an answer, declaring `archive: true` as "the ceiling, not the guarantee" and refining it per account from what `LIST` reported. I would do the same: declare the ceiling statically, narrow it from the session document at connect. If you would rather the descriptor stayed literally true, that is a different design and worth saying now.

**Compose the message resource, or parse RFC 822?** A JMAP client could satisfy the Gmail-shaped resource lazily by downloading each message's raw blob and handing it to `Message.parseRfc822`. I would rather compose it from the structured `Email` object the way `HeyClient.toMessage` does — "same shape, no parse" — because parsing blobs throws away the round-trip advantage that justifies the provider at all.

**What auth kind is a bearer token?** It behaves like the `password` flow: a secret the user supplies once, into the keyring. But the setup page would then say the wrong word. I am happy to reuse the password machinery with token wording, or add a `token` kind if you would prefer the distinction be real.

**Does the first version send?** I would ship `send: false`. Sending means `EmailSubmission`, an identity lookup and a write-scoped token, and the capability system means the compose button is simply not drawn. That also lets the first version be reviewed against a read-only token. `EmailSubmission` would follow separately.

## Scope

Roughly 1,500 lines, all below the provider boundary, touching nothing that branches on a provider id:

- `providers/Jmap.js` — descriptor, capabilities, mailboxes, filter mapping
- `providers/JmapApi.js` — method calls, back-references, `Email/changes`, normalisation
- `providers/JmapClient.qml` — transport, modelled on `GmailApiClient.qml`
- `providers/JmapAuth.qml` and `components/JmapSetupPage.qml` — token to keyring over stdin
- `tests/test_jmap.js`, plus `Registry.js` and `Makefile` registration

One caveat I would rather state up front: the account I can test against is a recent migration, so it has no multi-message threads and nothing filed in two mailboxes at once. `threadId` and `mailboxIds` are populated, but I cannot yet *demonstrate* `threads` or `labels` end to end, and I would not declare a capability I have only reasoned about. I will either test against a mailbox with real history or build fixtures before claiming either.

Happy to be told this does not fit, or that you would rather own it yourself.

ryanmcmillan:
Closes #37.

## What changed

- Add Fastmail to the mailbox chooser with its official mark and a token setup page.
- Store the account-scoped API token in GNOME Keyring over stdin, separate from an IMAP password for the same address.
- Read the Fastmail session document, then use every regional API and attachment address it supplies.
- Fetch mailboxes and pages with `Mailbox/get` plus a batched `Email/query` and back-referenced `Email/get`, normalize structured Email objects into the common message resource, and use Fastmail's indexed text search.
- Carry unread counts and every mailbox membership through the existing account, cache, sidebar and bar-badge paths.
- Narrow labels from the session's RFC 8621 `maxMailboxesPerEmail` value instead of assuming every JMAP account supports multiple memberships.

## Boundaries

This contribution deliberately remains a reader even when it is given a broader token. Archive, trash, read-state changes, star, batch and send controls are absent from rows, readers, menus, shortcuts and help; the client contains no `Email/set` or Email Submission path.

Thread IDs are normalized but `threads` remains false. A server ID alone does not provide conversation rendering or thread-wide action semantics, so that capability is not claimed here.

The live Fastmail account reports `maxMailboxesPerEmail: 1000`, which makes labels available, but it is a folders-mode account and cannot produce a live multi-mailbox message. Multiple membership is therefore verified structurally with fixtures rather than claimed as a live demonstration.

## Architecture and safety

Provider-specific session, protocol, authentication and normalization stay below the provider boundary. Shared account and view code asks only about `readOnly` and capabilities; the only provider-ID branches in `MailAccount` are the auth/client construction loaders.

The token never enters plugin settings or a command line. Requests use a `Timer` that aborts `XMLHttpRequest` after the deadline because QML provides no working timeout property. Only the initial Fastmail session address is fixed; API and download hosts come from the validated session document.

## Verification

- `make validate`
- 90 QML tests passed, 0 failed or skipped
- Live Fastmail read matrix: mailbox list, envelope list, paging, unread badge, message open and server-side search
- Automated coverage checks hidden row and reader controls, menu/compose source invariants, disabled keyboard paths and a broader session that must still refuse writes
- JMAP fixtures cover regional endpoints, batched query back-references, structured message normalization, multi-mailbox membership, mailbox-limit narrowing, attachments and server-capped paging

Implementation assistance: Claude Code and Codex.


huacnlee (changes requested):
Thanks for the thoughtful contribution and the work put into the JMAP implementation. After researching the current JMAP ecosystem, I don't want to introduce this feature into Omamail at this stage.

JMAP is a mature open standard and Fastmail uses it successfully in its own products, but it has not yet gained broad adoption across mainstream email providers and clients. Most general-purpose desktop and mobile mail clients still rely on IMAP and SMTP, while the current third-party JMAP client ecosystem remains comparatively small. That makes a separate JMAP protocol stack a future-facing investment rather than a capability Omamail currently needs.

Fastmail is already supported through Omamail's generic IMAP and SMTP provider, with reading, archive, trash, read state, star, reply, compose, and send. This PR adds a Fastmail-specific JMAP path but currently provides a read-only mailbox, so it does not close a necessary product gap and users needing normal mail actions would still have to configure Fastmail through IMAP.

For those reasons, I don't think the additional authentication, protocol, normalization, setup, and maintenance surface is justified today. This is not a rejection of JMAP as a technology; we can reconsider it if JMAP becomes broadly adopted or if Omamail has a concrete need for a generic JMAP provider with complete core mail actions and a clear advantage over the existing IMAP/SMTP path. For the current scope, please close this PR.

ryanmcmillan:
Understood. I'll close this Fastmail-only, read-only version. If I revisit JMAP here, it will be as a generic provider with conforming-server discovery, the core write and submission actions users already receive through IMAP/SMTP, and measured synchronization or performance advantages. Thanks for spelling out the acceptance bar.

PR #40, opened 2026-08-25
Rebuild a message for reading instead of stripping the sender's layout

alberthild:
A message now opens in a reading view Omamail builds itself, with the sender's original layout and plain text available from the same toolbar.

Reading mode rebuilds a new document from the parsed sender tree instead of filtering the sender's layout. It carries only text, checked links and prepared image data, plus bounded numeric dimensions for images. Every output element starts with an empty attribute list, so sender classes, colours, layout, styles, backgrounds, and unchecked resource-bearing attributes have no path into the rebuilt document.

It preserves headings, paragraphs, blockquotes, emphasis, code, rules, safe links, data tables, inline avatars, social-icon rows, and bounded sender-specified image sizes. Small interface-like images keep both dimensions; large images keep their width and derive their height from the source, so narrowing the reading column cannot distort their aspect ratio. Lists use one fixed 26px indent rather than stacking Qt's marker indent with another margin. Layout tables become semantic blocks; short layout rows remain lines. The reading column is measured from the active font and bounded to a readable measure, with a clearer gap below the message header. Application-owned content such as the Event card keeps the full panel content width.

Reader, Original, and Plain form one segmented toggle in the message toolbar and align with Open Web. At narrow widths, the complete mode group moves together to a second line.

Remote images remain blocked until the reader opts in. Once allowed, Omamail fetches each approved public URL silently through curl without following redirects, with time and size limits, validates its media type, and gives Qt a data URI only after the bytes are ready. There is no loading banner to move the body, and Qt never sees a pending remote source that could draw a misleading broken-image placeholder. All three readings otherwise come from one parse, so switching modes neither fetches nor reparses the message.

## Release Notes

- Open messages in a cleaner Reader view while retaining Original and Plain views in a segmented toolbar control.
- Improve newsletter, notification, list, avatar, icon-row, and event-card layout in the reader.
- Keep large images at their original aspect ratio and preserve the intended size of inline images.
- Load approved remote images without a broken-image placeholder or a banner that moves the message body.
- Fall back to plain text when a rich message is too heavy to render safely.

## Security fixes

- HTML `background` is treated as a resource-bearing attribute and refused before colour handling.
- Rebuilt text is escaped at the single serialization boundary.
- Style attributes are decoded before declarations are split, closing encoded `url(...)` bypasses.
- Excessively deep inline chains stop opening elements before they can multiply the output document.
- Image sources are validated only on images, `data:` URLs are restricted by media type, and hosts are checked across decoding forms.
- Remote image downloads refuse redirects and non-image responses and have explicit connection, request, and byte limits.

## Verification

- Node, source, image-fetch transport, shell, and install suites pass.
- 78 QML tests pass.
- `qmllint` exits successfully.
- `omarchy plugin validate .` passes.
- `git diff --check` passes.

`make validate` reaches the real unsubscribe transport harness, where this environment returns a proxy `502` instead of the harness's local `200`/`302`; its remaining QML, lint, and plugin-validation gates were run separately and passed.

## Known limitations

A blocked image in reading mode shows its alt text or `[image]`, but is not a clickable marker as it is in plain text. `<dl>` becomes plain paragraphs. Heading inference is size-only and limited to two levels.


huacnlee:
Thank you for the thoughtful Reader mode work — the rebuilt mail body is substantially easier to read, and the structural sanitization and performance work give this a strong foundation. I added the final toolbar treatment, kept application-owned cards at their original width, preserved bounded inline-image dimensions for avatars and social-icon rows, resolved the conflicts with main, and pushed those follow-ups to this branch.

PR #41, opened 2026-08-25
Edit and delete calendar events

vitorwilson:
Events were create-only. The calendar detail now edits and deletes them, with the rules those operations need enforced before any credential moves.

**What you can do**

- Edit... / Delete... on an event, drawn only where the operation is real. Google writes go to PATCH/DELETE on the item id — with `singleEvents=true`, changing or removing one occurrence of a series does what Google Calendar does. CalDAV writes go to PUT/DELETE on the event's own `href`, through the new `calendar-delete.sh`.
- Edit an all-day event as the dates it spans: the composer shows first/last day instead of time fields, and the write stays `DTSTART;VALUE=DATE` with an exclusive `DTEND` in ICS, `start.date`/`end.date` in Google's body. A title-only edit can no longer turn an all-day event into midnight-to-midnight times.
- Deletes confirm first, naming the event, through one dialog.

**Where it refuses**

- A recurring CalDAV event is one ICS holding a rule, its exceptions and its exclusions; this client does not re-serialize that state, so it refuses — and an occurrence carrying a RECURRENCE-ID answers the same way (parsed or not), because its `href` is the series' shared file.
- A read-only calendar draws no Edit/Delete, refuses writes in the controller, and is not offered in the creation picker. Synthesized Google sources are writable even when an older config persisted their former read-only stamp.
- A server-written `href` is resolved against its collection and accepted only on the same scheme, host and port — judged before the keyring is touched — and carries no raw whitespace, since the address becomes one quoted line of the transport's curl config.

**What a write preserves**

- A CalDAV PUT rewrites only the editable properties in the matching standalone VEVENT. Organizer, attendees, alarms, timezone definitions and server extension fields survive unchanged.
- Calendar cache schema changes invalidate ranges that lack the Google item id or original CalDAV resource instead of restoring events that cannot be edited safely.
- Create, update and delete share one write gate. A form becomes pending only when no older event write owns the controller, so an old completion cannot close a newer form.

**Invariants**: every decision lives in `.pragma library` JS with node tests (`caldavEventUrl`, `writeRefusal`, date-only serialization and CalDAV resource preservation); every new request gets the Timer-plus-abort deadline Qt's XHR makes necessary; a failed write surfaces on the view's banner; a completion is answered only by the form that started the write.

One drive-by, predating this branch: `ComposeView.submit` never checked it was open, so `Ctrl+Return` in the event form could re-queue a draft parked for the undo window and send it twice. One guard closes it.

**Verification**: `make test` and `make validate` pass on the rebased branch on the Qt the plugin runs on — node, shell and 81 offscreen QML tests, qmllint, and `omarchy plugin validate`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -27,6 +27,7 @@ three directories away from the client that calls it.
 | `providers/` | Everything that differs between mail services: a description per provider, the registry over them, the protocol each speaks, and the pair of objects — signs in, fetches — that each needs. |
 | `account/` | One mailbox and the list of them. `MailAccount.qml`, `Accounts.js`, and the rules in `Model.js` about what a list does after an action. |
 | `cache/` | What a query result and a message body are kept in, and the two objects that keep them. |
+| `calendar/` | The calendars an account serves and their events: the sources in `Sources.js`, the rules in `Calendar.js`, the controller that reads and writes them, and the range cache. |
 | `message/` | A message's own content: parsing it (`Message.js`) and making it safe to draw (`Html.js`). |
 | `components/` | Views. They draw what they are given and decide nothing. |
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -275,6 +275,11 @@ key. What matters while working:
   is world-readable.
 - Anything that could carry a credential passes through `OAuth.redact` before
   it can reach a label.
+- **CalDAV credentials go only to the configured source's own origin.** A
+  server-written event `href` is resolved against the collection and refused
+  when its scheme, host or port differs — `Calendar.caldavEventUrl` is the
+  one place that decides, and the controller judges it before the keyring is
+  touched.
 - **Every request that crosses the network is given up on eventually, and in
   QML that costs a `Timer`.** Qt's QML `XMLHttpRequest` has no `timeout` and no
   `ontimeout`: the properties do not exist, and assigning one reads back
--- a/App.qml
+++ b/App.qml
@@ -646,6 +646,27 @@ Item {
     settingsVisible = true
   }
 
+  // A delete asks first, and asks naming the target. Only the confirmation
+  // reaches the controller, with the event the dialog named.
+  function requestEventDelete(sourceId, event) {
+    if (!event) return
+    confirmDeleteDialog.openFor({
+      kind: "event",
+      name: String(event.summary || "Untitled event"),
+      message: "This event will be permanently deleted.",
+      sourceId: String(sourceId || ""),
+      event: event
+    })
+  }
+
+  function confirmDelete(request) {
+    if (!service) return
+    if (request.kind === "event" && request.event) {
+      service.calendarController.deleteEvent(request.sourceId, request.event)
+      calendarView.closeDetail()
+    }
+  }
+
   FloatingWindow {
     id: window
     visible: root.opened
@@ -1160,6 +1181,8 @@ Item {
             onCreateAt: function(startMs) { eventComposer.beginAt(startMs) }
             onCopyRequested: function(text) { root.copyText(text) }
             onOpenRequested: function(url) { Qt.openUrlExternally(url) }
+            onEditRequested: function(sourceId, event) { eventComposer.beginEdit(sourceId, event) }
+            onDeleteRequested: function(sourceId, event) { root.requestEventDelete(sourceId, event) }
           }
         }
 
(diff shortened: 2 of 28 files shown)


huacnlee (changes requested):
Thanks for the substantial work here. 

I don't think Google Tasks belongs in Omamail's product scope, so I can't accept this direction as part of the PR.

Omamail is intentionally a mail app. Calendar was added because calendar invitations arrive through email and need a native place to be understood and acted on; that exception should not turn the app into a general productivity suite. Adding Tasks creates a third top-level product area, expands navigation, keyboard contexts, caching and API machinery, and forces every existing Google account through another OAuth grant upgrade for functionality unrelated to reading and handling mail. That makes the app broader and heavier without supporting its core purpose.

Please remove the Google Tasks feature, including its top-level view, key bindings, cache/controller, Tasks OAuth scope and grant migration. The calendar event edit/delete work is independently useful and can be submitted as a focused PR after that separation, where we can review its behavior and safety on its own.

huacnlee:
I found three additional blockers in the calendar edit/delete portion:

1. **CalDAV credentials can be sent to a different host.** `Calendar.caldavEventUrl` accepts any absolute HTTPS `href` returned by the CalDAV REPORT. `CalendarController` then sends the configured username and password to that URL through `calendar-write.sh` / `calendar-delete.sh`. A server-controlled `href` such as `https://other.example/event.ics` therefore receives the source calendar's credentials. Resolve the href against the collection, then require the resulting scheme, host, and port to match the configured CalDAV source before looking up or sending credentials. Add a regression test that refuses a cross-origin absolute href.

2. **Editing an all-day event converts it into a timed event.** The detail currently offers Edit for all writable events. `beginEdit` turns an all-day date into local `00:00` fields, and `updateEvent` always serializes `DTSTART` / `DTEND` as UTC date-times and Google `start` / `end` as `dateTime`. Saving even a title-only change therefore destroys the all-day representation. Either model all-day edits and preserve `VALUE=DATE` / Google `date`, or do not offer Edit for all-day events until that is supported. Deleting an all-day event can remain available.

3. **Read-only CalDAV sources still advertise writes.** `CalendarEventDetail.canWrite` checks the provider, href and recurrence, but not `source.readOnly`. These calendars show Edit... and Delete... even though the operation cannot be performed. The capability should be refused before drawing the controls and again in `CalendarController` before starting a write.

If the Tasks implementation is retained despite the product-scope feedback, it also needs pagination: the task-list request uses Google's default page size and the task request caps at 100, while neither path consumes `nextPageToken`. Accounts beyond either page are silently shown as complete snapshots with data missing.

vitorwilson:
Thanks for the careful read. The Tasks feature is out of this branch entirely — the top-level view, the key bindings, the cache/controller, and the `tasks` scope with its grant-marker migration never happened here. The PR is now only calendar event edit/delete, rebased onto current main. (Tasks stays on a local branch for my own use, where I also fixed the pagination you noted; moot for this PR.)

On the three blockers:

1. **Cross-origin credentials** — `Calendar.caldavEventUrl` now resolves the href against the collection and requires scheme, host and port to match the configured source (scheme-relative hrefs included; an explicit `:443` still matches the default). The controller judges the resolved address in `startCaldavWrite` *before* the keyring lookup runs, so a refused address never has a password read for it, and the detail draws no buttons for one. Regression tests cover the cross-origin, host-suffix, port-mismatch and scheme-relative cases.

2. **All-day edits** — modeled rather than withheld. `updateEvent` takes its shape from the event being edited: an all-day event is written back as `DTSTART;VALUE=DATE` with an exclusive `DTEND`, and `start.date`/`end.date` to Google, so a title-only change keeps it all-day. The composer's edit mode shows first/last day instead of the time fields and computes the exclusive end by date-field arithmetic a daylight-saving boundary cannot shift.

3. **Read-only sources** — refused in both places. One wrinkle worth naming: synthesized Google sources were stamped `readOnly: true` with no consumer, so gating on the flag as it stood would have hidden Google's own writes. The stamp is dropped and `readOnly` now means "refuses writes" for every kind; `Calendar.writeRefusal` answers in create/update/delete, the detail draws no Edit/Delete, and the composer offers only writable calendars.

A self-review before this push found four more issues, each fixed in its own commit:

- A modified occurrence of a recurring CalDAV event (RECURRENCE-ID, no RRULE of its own) passed the recurring-write refusal while its `href` named the series' shared file — Delete would have removed the whole series. Refused now, including a RECURRENCE-ID too malformed to parse.
- A server-written href carrying a raw line break survived the base64 transport and landed in the quoted URL line of the curl config, where the break writes more options — re-opening the credential leak and worse. Refused in `caldavEventUrl` (raw whitespace is never a URL) and again in the write scripts before curl is invoked.
- Delete failures were silent: nothing listened for `eventDeleted`. Failures now land on the calendar view's banner, and a write completion is answered only by the edit that started it (cancel-then-reopen no longer eats the newer form).
- Pre-existing one-liner: `ComposeView.submit` never checked `opened`, so `Ctrl+Return` pressed in the event form could re-queue a draft parked for the undo window and send it twice.

`make test` and `make validate` pass locally.

PR #44, opened 2026-08-26
Revert the squashed mailto and attachment merge

huacnlee:
Reverts #43 so the contributor's original commits can be merged without squashing away their authorship. The feature will be restored immediately in a follow-up PR using the original PR 38 commit chain plus the reviewed mailto fix.


huacnlee:
Closed because main was restored directly with force-with-lease as requested; no revert commit is needed.

PR #46, opened 2026-08-26
Require complete release notes in every pull request

huacnlee:
PR descriptions now have one required section: `## Release Notes`. Its bullets must cover every functional change in the final diff, not repeat only the PR title or select a few highlights. Other description sections remain unrestricted.

Release preparation reads that section from every merged PR and checks it against the release diff. The rule remains documentation only; this PR does not change the release workflow.

## Release Notes

- Require every PR to list all functional changes in its `## Release Notes` section so release preparation can collect complete notes from PR descriptions.

## Verification

- `git diff --check`


huacnlee:
Temporarily closed while the original contributor PR is re-landed with the repository's one-commit-per-PR squash policy.

PR #48, opened 2026-08-26
Stream search rows from cache and bounded IMAP UID windows

alberthild:
## Summary

- paint conservative matches from every cached query immediately, then merge live provider metadata without replacing or duplicating visible rows
- stream interactive IMAP search through newest-first UID windows from a stable ceiling, with bounded response lines and small concurrent header batches
- keep a rotating `Searching server` state inside the query field until both the server listing and outstanding metadata reads are complete
- preserve the shared progressive client interface for Gmail, HEY, and IMAP while retaining exact non-interactive IMAP reads

## Verification

- `make test`
- `make qml-check`
- `git diff --check origin/main...stream-search-results`


huacnlee (changes requested):
The two-stage approach is the right shape — painting cached matches first and merging live rows as windows settle is exactly what the search UX needed, and the AGENTS.md/SPEC updates keep the UID-window invariant documented. A few things need to change before this can go in.

### Search results can be wrong, not just late

These three interact, so they're best fixed together.

1. **An empty server answer deletes the rows we just painted.** `account/MailAccount.qml:684` — `if (finalPage.ids.length === 0 && liveSummaries.length === 0) { if (!append) root.messages = [] }`. When the server returns at least one id the cached rows survive (merged via `applySummaries(..., cachedSearch)`); when it returns zero they're wiped. Search `unicorn`, a cached Sent row matches locally and paints, `UID SEARCH ... TEXT "unicorn"` over INBOX finds nothing — the row appears, then vanishes and the list says "Nothing matches that search". The two branches disagree about whether a cached row counts as a result; please make them agree.

2. **The result set is a union that can never shrink, and it gets persisted.** `account/MailAccount.qml:691` — `applySummaries` merges `cachedSearch` into the live answer rather than letting the server replace it, and the merged list is then written back under the query's cache key at `:693`. Nothing in this path ever removes an id. Repro: search `invoice`, trash one of the hits (`rememberList` at `:1198` only rewrites the *current* cacheKey, so other cached queries keep the row), search `invoice` again — the trashed message paints from cache, survives the live answer, and is re-persisted. On IMAP, opening it gives "That message is no longer in the mailbox".

3. **The local fallback isn't scoped to the mailbox being searched.** `cache/Cache.js:211` — `searchSummaries` scans every cached query, so Trash, Spam, Sent and label pages are offered as results of an Inbox search. Gmail's own search excludes trash/spam, and IMAP's `folder:INBOX TEXT "..."` only searches INBOX, so these are rows the provider would never return. The `-`/`:` guard in `localSearchTerms` is a good call for operator syntax, but there's no equivalent guard on scope.

### Interactive IMAP search can cost hundreds of logins

`providers/ImapClient.qml:270` — `searchWindow` steps down 4096 **UID numbers** per iteration, not 4096 messages, and the walk only stops early when the page fills. For any term matching fewer than `limit` messages it runs all the way to UID 1, and each iteration is a separate `root.run`, i.e. its own curl invocation with its own TLS handshake, LOGIN and SELECT (see the comment at `:68`). A folder whose highest UID is 200,000 costs 49 sequential logins; a long-lived INBOX archived down to a few hundred messages but with a highest UID near a million costs roughly 240. The old path did the same total SEARCH work as N commands on *one* connection.

The comment's claim that "a UID range of width 4096 can contain at most 4096 messages" bounds the response line, which is the curl constraint we care about — but it doesn't bound the number of round trips, and a sparse range costs a full connection just to learn it's empty. Two options that keep the streaming property: cap the walk and fall back to `uidListCommand()` + `searchCommands()` after the first window or two, or window over a UID snapshot fetched in the background so the ranges track messages instead of UID numbers.

Related: each settled window starts its own `getMessages`, which opens up to `streamedSummaryConcurrency` more connections, so several curl processes can be live at once. Dovecot defaults to 10 concurrent IMAP connections per user/IP — worth a shared ceiling rather than a per-call one.

### Failure mid-walk skips results silently

`providers/ImapClient.qml:281` — `finish(found, searchError, nextUid > 0)` calls `callback(null, error)`, so `finalPage` is null and `finishIfReady` (`MailAccount.qml:673`) bails out *before* it reaches `root.nextPageToken = finalPage.nextPageToken`. But `idsArrived` (`:729`) has already set `nextPageToken` to the provisional `"25"` from the last progress event, because `searchPage` returns `nextOffset = String(start + limit)` whenever `hasUnscanned` is true. The user ends up with 12 rows, an error banner, and an enabled "Load more" that pages from offset 25 — matches 13-24 are unreachable. The `found` list is discarded too, even though those matches are valid.

### Smaller things

- `providers/GmailApiClient.qml:219` — `progress([payload])` fires once per message and Gmail's `getMessages` issues all 25 requests in parallel, so `paintPayloads` runs 25 times per page. Each run does two `mergeSearchResults` passes (sort included) over the whole merged list and reassigns `root.messages`, rebuilding every delegate in `MessageList`'s `Repeater` (`components/MessageList.qml:45`). Coalescing on a short timer, or batching the way IMAP's `streamedSummaryBatch` does, gets the same visible benefit for much less churn.
- `account/MailAccount.qml:761` — `fetchSummaries` now writes the whole accumulated list on every page including appends, where previously only page one was cached. There's no per-query row cap, only `MAX_QUERIES = 12`, and `Cache.serialize`/`Cache.prune` run on the GUI thread (`cache/CacheStore.qml:113`). Fifteen "Load more" clicks in a large mailbox stores ~375 summaries for that one query; twelve such queries put the store into the megabytes, rewritten whole 800ms after every list change — against a file whose own header comment calls that "the wrong shape". A per-query cap would keep the local-search win without the growth.
- `providers/ImapProtocol.js:389` — `UID FETCH * (UID)` is undefined on an empty folder. Dovecot answers OK with no FETCH row, which `ImapClient.qml:266` handles correctly, but a server that rejects an out-of-range `*` with BAD will make `Imap.isFailure` turn an empty folder into a search error instead of an empty result. `UID FETCH *:* (UID)` is well-defined in both cases.

Two notes that aren't defects: `listRefreshed()` (`MailAccount.qml:386`) has no connections anywhere in the repo, so the extra emissions here are inert — fine to keep, just so you know they aren't doing anything yet. And `progress` being accepted but never called by `GmailApiClient.listMessages` / `HeyClient.listMessages` degrades safely (ids only stream on IMAP), which reads deliberate.

`make test-js`, `make test-shell` and `tests/test_source.sh` all pass on the branch — the issues above are behavioural rather than anything the suite covers, so a couple of them would be worth new cases in `tests/test_model.js` and `tests/test_imap.js`.

alberthild:
Thanks for the detailed review — I have pushed a follow-up in d1cfe7d that addresses the feedback.

The update:

- makes live provider results authoritative over cache-first search previews, including empty results and partial metadata failures;
- scopes cached search candidates per provider and caps persisted summaries per query;
- bounds IMAP search connection churn by streaming the newest UID window first and reusing a snapshot-backed connection for older windows;
- batches Gmail progress updates and propagates partial metadata errors consistently across providers;
- invalidates unsafe continuation offsets after membership-changing actions and failed page-one reads;
- serializes optimistic actions with active list streams, deferred navigation, cross-query revalidation, and queued automatic mark-read operations so stale snapshots cannot be persisted.

I also added regression coverage for the cache, model, provider, IMAP, paging, and source invariants. The local verification is green: all JavaScript and shell tests, qmllint, source checks, and 81/81 QML runtime tests pass. A separate adversarial review pass found no remaining blocker, high, or medium findings.

The branch is ready for another look. Thanks again!

PR #56, opened 2026-08-27
Mark the current time on the week grid

RyanRusnak:
## Summary

- draw the current time across today in the visible week, with a bead on the day column and a matching label on the time rail
- keep the marker, events, elastic hour range, and click-to-create slots on the same local wall-clock geometry through daylight-saving transitions
- keep the minute clock in the calendar controller and run it only while the week view is visible
- keep placement and time formatting in calendar/Calendar.js so node tests exercise the decisions without a compositor

## Invariants

- no marker is drawn outside today, the visible week, or the drawn hour range
- the marker is declared after events so it crosses a meeting in progress
- inherited urgent and background colors carry the marker and label
- the timer uses triggeredOnStart and stops outside the visible week view

## Verification

- node tests/test_calendar_feed.js, including spring-forward and fall-back cases
- make test-js
- python3 tests/test_qml_names.py
- python3 tests/test_qml_text_format.py
- bash tests/test_source.sh
- git diff --check

The local qmllint installation could not start because its Homebrew Qt build references an unavailable ICU 74 library. The full shell suite also encountered the existing intermittent attachment runtime-directory harness failure.

### DIFF
--- a/calendar/Calendar.js
+++ b/calendar/Calendar.js
@@ -681,6 +681,35 @@ function eventHeight(event, day, hourHeight) {
   return Math.max(Number(hourHeight) * 0.42, (end - start) / 3600000 * Number(hourHeight))
 }
 
+// Where "now" sits in a day column, or -1 when it does not belong on the grid:
+// another day, or an hour outside the range the week view drew. The caller
+// draws nothing on -1 rather than clamping to an edge, because a line pinned to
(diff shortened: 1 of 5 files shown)


RyanRusnak:
weekview

This app is great. I just had claude add a "now" line to better see where I am in the day.

Issue #49, opened 2026-08-27
Sent Emails: email are not sent and nowhere to be found


marcotroisi:
I've had two occurrences of emails that I've sent, where not only the message was not sent, but it is nowhere to be found.

There was no visible error when I pressed Send, and I'm unsure if logs exist anywhere for me to investigate. I'd be happy to investigate myself and even post a PR with a resolution.

The most pressing issue is that some messages may get swallowed up and simply disappear if something goes wrong. The less pressing, longer term issue is that a better handling of _Sent_ vs _Sending_ messages would greatly improve the user experience.

marcotroisi:
Further investigation has revealed what the actual problem is: sent emails (sent via SMTP) are not added to "Sent" folder via IMAP, as most email clients do. So the emails are sent, but they just disappear from my own history.

This [PR](https://github.com/huacnlee/omamail/pull/60) would fix that.

lukaseppler:
Related follow-up: #79.

#49 / #60 are the success path (SMTP accepted, no IMAP APPEND to Sent). On iCloud I hit the **failure** path: SMTP timed out (`curl` 28, "The mail server did not answer in time"), the composer stayed parked/hidden, and the only copy of the message was gone — not in compose, Drafts, Sent, or Outbox.

That looks like the original "swallowed" report here, but it is a different code path from missing Sent copies.

huacnlee:
Fixed by merged #103 (`39e4b59818f9609417903f3eaf848a5f6371d72c`), included in v0.8.1: after SMTP accepts the message, Omamail appends the submitted message to the server's SPECIAL-USE Sent folder with the Seen flag. If the server has no Sent folder or saving the copy fails, the send reports a warning rather than silently claiming the copy was saved. #60 remains open and unmerged; it is not the integrated fix cited here.

The failure-path follow-up mentioned above is also addressed by merged #88 (`64095e164c9d8af30d3cb573355d3961e67e28c3`, included in v0.8.0): a failed send restores the parked composer and recovery state, including when a newer draft must be preserved.

Verified at `6568bc9bc7797693d6cb087d89e5301c5d85f2bf` (the subsequent main change `85eb6c2` does not change these paths), using Node v26.7.0 and Qt 6.11.2:
- IMAP protocol and mail-transport tests pass.
- 17 compose/failure QML checks pass.
[...]

PR #60, opened 2026-08-28
Finish SMTP delivery before saving the Sent copy

eliasfaltin:
## Why

SMTP acceptance must finish a send immediately. A later Sent-copy failure must never restore a delivered draft or encourage another send.

## Fix

`ImapClient` reports delivery before saving the Sent copy. Copy failures reach the mailbox as warnings. Gmail and Outlook SMTP hosts skip APPEND because they file submissions themselves.

Folder discovery requests SPECIAL-USE attributes when supported. Discovery retains upstream's generation checks and treats an empty listing as authoritative. Generic IMAP saves the exact SMTP message with the Seen flag through upstream's shared, OAuth-aware append transport.

## Blast radius

Draft saves and OAuth credentials retain upstream's shared append implementation. SMTP completion and background filing have separate request handles. Cancelling a completed send cannot cancel its Sent copy.

## Testing

- Focused IMAP send QML tests: 7 passed, including cancellation after SMTP acceptance.
- Full QML suite: 588 passed, plus 4 native HTTP and 64 sidebar checks.
- IMAP protocol tests and all shell tests outside the existing AI bridge failure passed.
- `qmllint` and `omarchy plugin validate .` passed.
- `make validate` reaches an existing failure in the unchanged AI bridge suite. The same 10 failures reproduce on upstream main at `77db4b1`.

Latest upstream main is merged without rewriting branch history.


marcotroisi:
This [has caught me](https://github.com/huacnlee/omamail/issues/49) before. Really looking forward to seeing this PR merged and the problem fixed.

lukaseppler:
This PR covers the success path (SMTP accepted → APPEND to Sent). The timeout/error path is separate: after a failed send the parked composer is not restored, so the draft can disappear entirely. Filed as #79.

karamble (commented):
Reproduced the bug and verified this branch fixes it, on a self-hosted Postfix + Dovecot mailbox reached over IMAP 993 and SMTP 465, with the server logs and the maildir as the evidence rather than the client's own display.

Before, on `main`: a send is accepted by SMTP and no copy reaches the server's Sent folder. The newest file in the account's `Maildir/.Sent/cur` was two and a half months old, and the IMAP sessions in the seconds after a send are `in=55`, `in=68`, `in=186` byte status polls — an order of magnitude too small to be an APPEND. That matches the code: the only `imap-append` in `providers/ImapClient.qml` is in `saveDraft`, and `sendMessage` returns as soon as SMTP does.

With this branch, the same send produces an IMAP session of `in=329 out=599` — the 275-byte message plus APPEND overhead — and a new file in `Maildir/.Sent/cur` whose content is byte-for-byte the message that was submitted. Its maildir filename ends `:2,S`, so `upload-flags = "seen"` arrives as `\Seen` on disk and the copy is not shown as unread. Two consecutive sends produced two appends and two files; nothing was duplicated and the SMTP path is unchanged.

Applied on top of current `main` rather than the branch as-is: the branch is based on `Version 0.6.0`, and `tests/qml/imports/Quickshell/Io/Process.qml` has since gained the same four lines upstream, so the commit cherry-picks onto `main` cleanly at 5 files and 160 insertions. Worth a rebase before merge, but nothing conflicts.

Gates on that branch, all passing: `make test-js`, `make test-shell`, `make test-qml` (131 passed, including this PR's `ImapSend::test_smtp_success_appends_a_seen_copy_to_sent` and `ImapSend::test_a_failed_sent_copy_does_not_claim_delivery_failed`), and `make validate`. `qmllint` exits zero with the usual unresolved `qs.Ui` / `qs.Commons` imports.

Two design points I checked rather than assumed. An append failure is reported as a warning and does not become a send error, which is right: SMTP has already accepted the message and a retry would deliver it twice. And the bytes appended are the same message that was submitted, so the Sent copy cannot disagree with what the recipient received.

This is the fix for the missing Sent copies described in #49. #79 is the failure path and is not covered here.

patrykbart:
I ran this against three Gmail mailboxes while working on something else, and hit one case worth knowing before it lands.

The APPEND is unconditional, and `imap.gmail.com` and `outlook.office365.com` are both shipped presets whose SMTP files a Sent copy on the server already. So those mailboxes get the message in Sent twice.

Gmail's own de-duplication does not absorb it, which is the part that surprised me: Gmail matches on `Message-ID`, and `buildRawMessage` emits none — #84. So the second copy arrives as a second message rather than as the same one recognised.

What I ended up with, on top of this branch:

```js
// providers/ImapProtocol.js
var SENT_FILED_BY_SERVER = [
  "smtp.gmail.com",
  "smtp-mail.outlook.com",
  "smtp.office365.com"
]

function serverFilesSentCopy(settings) {
  var host = normalizeSettings(settings).smtpHost
  if (host === "") return false
  for (var i = 0; i < SENT_FILED_BY_SERVER.length; i++) {
    if (SENT_FILED_BY_SERVER[i] === host) return true
  }
  return false
}
```

and an early return in `sendMessage` before the Sent folder is resolved.

Asked of the resolved SMTP host rather than of the preset that was picked, because a Workspace or Microsoft 365 mailbox on a custom domain is configured by hand as often as it is chosen from the list and submits through the same servers either way. `smtp.office365.com` has no preset of its own — the Outlook row names the address personal accounts are given.

The default stays "append". A server nobody has established this about keeps the copy this branch saves: a duplicate in Sent is visible and recoverable, and a missing sent message is the fault the APPEND exists to prevent.

Mozilla hit the same thing and reached the same answer — [bug 1427619](https://bugzilla.mozilla.org/show_bug.cgi?id=1427619), "Don't save outgoing messages to Gmail/Outlook Sent folder by default, because gmail/outlook forces it at the server".

Happy to open this as a follow-up PR once yours is in, or to hand you the patch to fold in here — whichever you prefer. It is your change and I did not want to open a competing one.

karamble:
The duplicate is #84: the message this client builds carries no `Message-ID`, so Gmail cannot tell your appended copy from the one it filed itself and keeps both. The problem is that the email client doesn't attach date and message-id to the emails.

I have that fix open as #85. With it applied on top of this branch, the same send produces one copy in `[Gmail]/Sent Mail` instead of two.

| | copies in Sent Mail |
|---|---|
| this branch | 2 |
| this branch + #85 | 1 |

The second copy on this branch has no `Message-ID` and no `Date` at all — it is the raw APPEND — while Gmail's own carries both, which is why they do not match.

Outlook and Office 365 I have no account for, so your allowlist may still be needed there.

huacnlee:
@eliasfaltin I pushed a follow-up review commit rather than merging this yet, so you can inspect the resulting design and behavior.

The main change is that SMTP acceptance now completes delivery immediately. Saving the Sent copy runs afterward as best effort, so an APPEND timeout or failure cannot reopen an already delivered draft and encourage a duplicate send.

I also added these safeguards:

- Wait for folder discovery before reading the Sent folder.
- Use `LIST "" "*" RETURN (SPECIAL-USE)` when the server advertises SPECIAL-USE, as specified by [RFC 6154](https://www.rfc-editor.org/rfc/rfc6154.html).
- Skip APPEND for Gmail and Outlook SMTP hosts because those services file SMTP submissions themselves; [Google explicitly documents that Gmail automatically copies SMTP-sent messages and clients should not save them again](https://support.google.com/mail/answer/78892?hl=en).
- Route background Sent-copy warnings through `MailAccount`, covering compose, calendar replies, and mail unsubscribe sends consistently.
- Verify that generic IMAP still appends the exact SMTP message bytes with the Seen flag.

`QT_QPA_PLATFORMTHEME= make validate` passes, including 222 QML tests. An independent review found no concrete issues. Please check the follow-up commit and the revised PR description; I have left the PR open and unmerged.

huacnlee:
Thanks, and thanks to everyone who tested. This landed on main via #103 (SMTP completes the send, the Sent copy is appended afterwards as best effort and reports a warning), and #174 moved that into the Rust backend's imap.send. #85 added the Message-ID that made Gmail keep two copies. None of the files this branch touches exist any more, so closing as superseded. Two small pieces from here are not on main and would be welcome as a fresh change: skipping APPEND for the Gmail and Outlook SMTP hosts, and asking LIST for RETURN (SPECIAL-USE).

PR #62, opened 2026-08-28
Refuse CR/LF in remote-image URLs to close a curl config injection

jzuijlek:
image-fetch.sh builds a curl config with `url = "<url>"`, where the escape helper neutralises the quote and the backslash but not a line break. A CR or LF in the URL therefore ends the url line and turns whatever follows into another curl option — `output = <path>` writes the fetched (attacker-served) bytes to a path of the sender's choosing.

The value is a mail  src, written by a stranger. It is classified up in Html.js through `imageSourceKind`, whose normaliser strips CR/LF/HT before reading the host — so a URL that reads as a public host to that gate can still carry a raw newline by the time it reaches this script, and the scheme check here does not catch it because the value still begins with https://.

Refuse a URL that spans lines before any config is built, the same guard calendar-write.sh and calendar-delete.sh already apply to their fields. Adds a regression test that a line-broken URL is refused and builds no config.


huacnlee:
Thanks

PR #63, opened 2026-08-29
Write from an alias, reach a STARTTLS server, and pick a recipient from the address book

bchurch95:
An IMAP mailbox can now write from more than one address, reach a server that speaks STARTTLS, keep its password across a restart, choose whether new mail is announced, and fill a recipient from an address book rather than from memory.

## Send-as aliases

An IMAP account has no settings endpoint to ask, so the addresses it may write from are typed on the setup page and stored with the account. They arrive in the composer's From menu in the same shape Gmail's send-as settings do, so nothing above the provider boundary has to know which kind of mailbox it is looking at, and the envelope sender is resolved from the From header that was actually chosen rather than from the account's username.

One alias may be marked the default, which is the address a new message opens on. `account/Aliases.js` is the one place that reads the field and the one place that writes it back — the setup page re-formats the list every time it opens, so a parser that disagreed with the formatter would rewrite the user's own text under them. It reads quoted display names, which is what keeps `"Lee, Jason" <j@example.com>` one alias rather than two halves of one.

## STARTTLS

Ports 143 and 587 are the upgrade ports, and dialling them as implicit TLS is why iCloud's submission host failed its handshake. They are named as a list rather than inferred from "not 993", because a server offering implicit TLS on a port a hosting panel picked has no STARTTLS to negotiate and would simply stop connecting. The plaintext scheme is never a plaintext session: `mail-transport.sh` adds `ssl-reqd` to every non-loopback `imap://` and `smtp://`, so a server that turns out not to offer the upgrade is refused rather than spoken to in the clear. The loopback exemption is recognised by literal text down there, so a host is lowercased once on the way in.

## Notifications

New mail can be announced or not, from the settings page. What counts as new is answered by the ids already seen, and — for an unread message that was never on the cached page — by a floor: the mailbox's own newest timestamp at the moment notifications were primed. Server clock against server clock, set once and never raised, so a machine whose clock runs ahead does not go silent for the session and a message that arrives out of order is still announced.

## Contacts

A picker beside the recipient fields, filled from Thunderbird's collected addresses, a `contacts.json` or `contacts.vcf` under the config directory, and the addresses already in the message cache. Everything it reads is a file this program did not write, so every address passes one gate before it can be offered as somebody to send mail to; a bulk send is not a room full of contacts, and a display name the parser filled in from a local part is not a name anybody wrote.

## Secrets

`secret-tool lookup` writes its value with no trailing newline, so a reader splitting on one never fired and a stored password read back as no password at all — the mailbox asked to be signed in to again on every restart, and a CalDAV source reported itself as having none. All three lookups now read the whole output when the process is done with it, and strip the pipe's newline and nothing else: a password may legitimately end in a space, and trimming it would authenticate as a different string.

## Sending is not retried

A dropped TLS handshake is worth a second attempt; a message the server already took is not. curl's `--retry-all-errors` cannot tell them apart, because by the time the transfer failed it had already been sent — so the retry is the transport's own, on the three exit codes that mean the command never left this machine (6, 7, 35). A rejected password is not retried at all, which is what iCloud and Gmail lock an app password for.

## Verification

`make validate` — node tests, source regressions, the QML tests, `qmllint` and `omarchy plugin validate` — passes on a clean checkout of this branch, on Omarchy with Qt 6.

New coverage: `tests/test_aliases.js` holds the alias format and its round trip, including that a display name carrying a comma survives being written back twice; `tests/test_secrets.js` holds what may be stripped from a keyring value; `tests/test_transport.sh` holds that a send failing mid-transfer or on a timeout is attempted exactly once while a handshake failure is attempted three times; `tests/qml/tst_popup_toggles.qml` holds that a control can put its own popup away; `tests/test_contacts.py` holds the shapes the harvesters are handed. `tests/test_model.js` and `tests/test_imap.js` cover the notification floor and the scheme chosen per port.

Tested live against iCloud IMAP and SMTP with several send-as aliases.

## Release Notes

- An IMAP mailbox can be given send-as aliases, set on its setup page and offered in the composer's From menu. One may be marked the default, which is the address a new message opens on.
- IMAP and SMTP servers that use STARTTLS — iCloud's submission host on port 587, and any server on 143 — now connect. The upgrade is required, never optional.
- A saved IMAP password survives a restart, and a CalDAV calendar no longer asks for a password it already has.
- New-mail notifications can be turned off, on the settings page under Notifications.
- A Contacts button beside the recipient fields fills To, Cc or Bcc from Thunderbird's address book, from a `contacts.json` or `contacts.vcf` in the config directory, or from addresses already seen in the mailbox.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ node_modules/
 # repository; see AGENTS.md for what is kept and why.
 docs/design/
 docs/superpowers/
+
+# Python leaves these beside the harvester scripts.
+__pycache__/
--- a/Makefile
+++ b/Makefile
@@ -67,6 +67,7 @@ test-js:
 	node tests/test_senders.js
 	node tests/test_oauth.js
 	node tests/test_credentials.js
+	node tests/test_secrets.js
 	node tests/test_gmail_api.js
 	node tests/test_message.js
 	node tests/test_calendar.js
@@ -82,6 +83,7 @@ test-js:
 	node tests/test_model.js
 	node tests/test_keymap.js
 	node tests/test_accounts.js
+	node tests/test_aliases.js
 	node tests/test_menu.js
 	node tests/test_provider.js
 	node tests/test_imap.js
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,7 @@ QML_FILES := Service.qml BarWidget.qml App.qml \
 	components/ReaderSkeleton.qml \
 	components/ComposeView.qml \
 	components/RecipientSuggestions.qml \
+	components/ContactsPicker.qml \
 	components/UndoSendToast.qml \
 	components/DraftSavedToast.qml \
 	components/SearchBar.qml \
--- a/Service.qml
+++ b/Service.qml
@@ -58,6 +58,7 @@ Item {
     settings ? settings.undoSendSeconds : Outbox.DEFAULT_DELAY_SECONDS)
   readonly property bool alwaysRenderHeavyMessages: Html.alwaysRenderHeavyMessages(
     settings ? settings.heavyMessageRendering : null)
+  readonly property bool notifyNewMail: String(settings ? settings.notifyNewMail : "On") !== "Off"
 
   // Thunderbird and Betterbird keep both explicit and learned addresses in
   // their local profile. The helper reads those databases without modifying
@@ -105,6 +106,10 @@ Item {
 
   function setAlwaysRenderHeavyMessages(value) {
     persistSetting("heavyMessageRendering", Html.heavyMessageRendering(value))
+  }
+
+  function setNotifyNewMail(value) {
+    persistSetting("notifyNewMail", value ? "On" : "Off")
   }
 
   // ---------------------------------------------------------- the accounts
--- a/account/Accounts.js
+++ b/account/Accounts.js
@@ -102,6 +102,39 @@ function portOr(value, fallback) {
   return port
 }
 
+function normalizeAliases(value) {
+  if (!value) return []
+  var rawList = []
+  if (Array.isArray(value)) {
+    rawList = value
+  } else if (typeof value === "string") {
+    rawList = value.split(/[\r\n,]+/)
+  }
+  var out = []
+  var seen = {}
+  for (var i = 0; i < rawList.length; i++) {
+    var item = rawList[i]
+    var email = ""
+    var displayName = ""
+    if (typeof item === "string") {
+      email = trimmed(item).toLowerCase()
+    } else if (item && typeof item === "object") {
+      email = trimmed(item.email || item.sendAsEmail).toLowerCase()
+      displayName = trimmed(item.displayName || item.name)
+    }
+    if (!email || !isValidEmail(email)) continue
+    if (seen[email]) continue
+    seen[email] = true
+    out.push({
+      email: email,
+      displayName: displayName,
+      isPrimary: false,
+      isDefault: false
+    })
+  }
+  return out
+}
+
 function makeImapSettings(raw) {
   var values = raw || {}
   return {
@@ -110,6 +143,7 @@ function makeImapSettings(raw) {
     smtpHost: trimmed(values.smtpHost),
     smtpPort: portOr(values.smtpPort, 465),
     username: trimmed(values.username),
+    aliases: normalizeAliases(values.aliases),
     insecure: values.insecure === true
   }
 }
--- a/account/Accounts.js
+++ b/account/Accounts.js
@@ -116,11 +116,30 @@ function normalizeAliases(value) {
     var item = rawList[i]
     var email = ""
     var displayName = ""
+    var isDef = false
     if (typeof item === "string") {
-      email = trimmed(item).toLowerCase()
+      var str = trimmed(item)
+      if (/\(default\)|\[default\]/i.test(str)) {
+        isDef = true
+        str = str.replace(/\(default\)|\[default\]/gi, "").trim()
+      } else if (str.endsWith("*")) {
+        isDef = true
+        str = str.replace(/\*+$/, "").trim()
+      } else if (str.startsWith("*")) {
+        isDef = true
+        str = str.replace(/^\*+/, "").trim()
+      }
+      var matchAngle = str.match(/^(.*?)\s*<([^\s@]+@[^>]+)>\s*$/)
+      if (matchAngle) {
+        displayName = trimmed(matchAngle[1])
+        email = trimmed(matchAngle[2]).toLowerCase()
+      } else {
+        email = trimmed(str).toLowerCase()
+      }
     } else if (item && typeof item === "object") {
       email = trimmed(item.email || item.sendAsEmail).toLowerCase()
       displayName = trimmed(item.displayName || item.name)
+      isDef = item.isDefault === true || item.default === true || item.defaultFrom === true
     }
     if (!email || !isValidEmail(email)) continue
     if (seen[email]) continue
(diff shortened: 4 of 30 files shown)


huacnlee (changes requested):
The keyring diagnosis is right and the fix is the correct one — `secret-tool lookup` writes no trailing newline, so `SplitParser`'s `onRead` never fires. The STARTTLS problem is real too. But the branch has grown well past what the description covers, and several of the changes are not safe to merge as they stand.

## Scope

The description lists three changes; the branch carries eight commits, including a contacts harvester and picker (~370 lines), a notifications toggle, and transport retry. AGENTS.md asks for the title and description to be re-derived from the final `base...HEAD` diff every time a commit is added, and this is exactly the case it describes. It also needs a `## Release Notes` section.

The commit subjects and the PR title carry `feat(imap):` / `fix(smtp):` prefixes. AGENTS.md forbids scope prefixes here and says so at length, including a note that it has been re-litigated once already: a title is the imperative outcome with nothing in front of it.

I would rather see this as three pull requests — aliases + STARTTLS, the keyring fix, and the contacts harvester — because the first two need discussion and the keyring fix is ready now apart from one missing call site.

## Blocking

**`scripts/mail-transport.sh:187` — `--retry 2 --retry-all-errors` can send a message twice.** The flags go on the single shared curl invocation, which also serves `mode = "smtp"` and `imap-append`. If the server accepts DATA and returns 250 but the response is lost — a reset, or `max-time = 60` elapsing during the final read — curl exits 28/55/56 and re-runs MAIL FROM / RCPT / DATA, and the recipient gets two copies while the panel reports one send. `imap-append` has the same shape and produces duplicate drafts. Sending is not idempotent and must not be blanket-retried.

`--retry-all-errors` also retries permanent failures. A wrong app password (curl exit 67) becomes three LOGIN attempts per operation; with a 120-second refresh interval that triples the failed-auth rate against iCloud and Gmail, both of which throttle and lock on repeated failures. Certificate refusals and the new `ssl-reqd` failures are retried for nothing and add seconds before the user sees an error. `--retry-delay 1` is a fixed delay, not the backoff the commit subject claims. Restrict the retry to the read-only `imap` mode, or to connect/handshake-stage failures only.

**`providers/ImapProtocol.js:321` and `:336` — inferring TLS from the port number breaks non-standard ports.** `port === 993 ? imaps : imap` and `port === 465 ? smtps : smtp` fix 143 and 587, but any server offering *implicit* TLS on another port now gets a plaintext scheme plus `ssl-reqd`, so curl opens cleartext and issues STARTTLS that the server does not offer, and the account stops connecting. The port does not determine the encryption mode; an explicit "SSL/TLS or STARTTLS" setting does, which is what every other mail client exposes.

This also moves an invariant out of JS. "A plaintext scheme is only ever emitted for loopback" no longer holds in `ImapProtocol.js` at all, and cleartext credentials are now prevented solely by the `case "$url" in … ssl-reqd` blocks in the transport, which every future transport branch has to repeat correctly. The two loopback lists have already drifted: `isLoopback` accepts `::1`, but `smtp://::1:*` / `imap://::1:*` cannot match a real URL, since an IPv6 URL is bracketed (`smtp://[::1]:465`).

**`providers/ImapProtocol.js:177` and `account/Accounts.js:105` — the alias parser exists twice.** `parseAliases` and `normalizeAliases` are the same ~65 lines, down to a second copy of the email regex — and the two validators are not identical (`EMAIL_PATTERN` vs `isValidEmail`). Two parsers that must agree will stop agreeing, and the symptom is an alias the setup page accepts being silently dropped when the account reloads. One implementation.

**`providers/ImapProtocol.js:239` — the format/parse round trip corrupts aliases.** `formatAliases` joins with `", "` while `parseAliases` splits on `/[\r\n,]+/`, and quoting is not honoured. `"Lee, Jason" <j@icloud.com>` comes back with `displayName` of `Jason"` and `Lee, ` gone; the settings page re-runs the round trip every time it opens, so the corruption is persistent. Either honour quoted display names or drop the comma as a separator and take one alias per line.

**`account/Model.js:602` — the arrival filter drops real new mail.** It compares `summary.date` — an `INTERNALDATE` or a sender-written `Date:` header — against the local clock at `MailAccount` creation. If the local clock runs ahead of the server's (no NTP, a wrong RTC after boot, a resumed VM), every arrival is `time < minTime` and notifications stop silently for the whole session. `parseInternalDate` also defaults a missing offset to `+0000`, which can shift a timestamp up to 12 hours into the past, and a summary with no parsable date is now dropped rather than notified. `notificationsPrimed` already prevented the false positives this targets; this only adds false negatives. If a baseline is needed, seed it from the primed snapshot's newest timestamp rather than `Date.now()`, and allow a tolerance window.

**`components/ComposeView.qml:1354` — the picker is never closed on teardown.** `fromMenu` is closed explicitly in `clearCurrentDraft()` (line 112) and at line 213 because `ComposeView` is hidden rather than destroyed and a `QQC.Popup` lives in the window overlay. `contactsPicker` has no such call, so sending, discarding, or starting a new draft with it open leaves it on screen over the message list. Add `contactsPicker.close()` next to `fromMenu.close()`.

**`calendar/CalendarController.qml:578` — the keyring fix misses the third call site.** The CalDAV `passwordLookup` still uses `SplitParser` with `splitMarker: "\n"` on `secret-tool lookup`, so `onRead` never fires, `handlePassword("")` runs from `onExited`, and the source fails with "Add this calendar's password in Settings" even though the password is stored. `eventPasswordLookup` at line 673 is already on the `StdioCollector` pattern, so the file is now inconsistent with itself.

**`tests/test_recipients.js:31` — coverage went down.** The new `append` assertions were written over the existing `accept("ja", { name: "", email: "jane@example.com" })` case rather than beside it, removing the only test of the `name === ""` branch of `address()`. Please restore it.

## Repository conventions

`providers/ImapClient.qml:566` — `getSendAs()` now holds about thirty lines of merge, dedupe and default-resolution logic. AGENTS.md: everything that parses, formats or decides lives in a `.js` so the node tests can reach it without a compositor; QML holds no logic worth testing. This belongs in `Imap.js` with tests.

`formatAliases` builds a display string for the settings field, and `Imap.js` is the protocol and nothing else. Formatting for the setup page belongs with the settings shape.

The new code carries almost no comments. This repository explains why a thing is the way it is, at length, and the new files read as if they came from a different codebase.

## Smaller

`providers/ImapAuth.qml:259` and `providers/AuthManager.qml:782` — `String(...).trim()` changes more than the reported bug. The old path passed the value untrimmed, so a stored password with a leading or trailing space now authenticates as a different string and surfaces as "wrong password" with no way to correct it. Strip a single trailing newline instead.

`scripts/contact-suggestions.py:97` — `omamail_cache_records` harvests `to`, `cc` and `bcc` of every cached message, so one 50-recipient bulk mail permanently adds 50 entries, along with every spam sender and the user's own address, with no cap. `Message.parseAddress` fills `name` from the local part when there is no display name, so the picker will offer `noreply <noreply@example.com>` and insert it into the To field.

`scripts/contact-suggestions.py:72` — `parse_vcf` wraps the whole read-and-loop in one `try/except Exception: pass`, so the first property line without a colon discards every remaining vCard in the file. vCard folding is also unhandled: `line.strip()` destroys the leading space marking a continuation, and exported files from Google Contacts and iOS fold routinely.

`components/ContactsPicker.qml:2` — `QtQuick.Controls` is imported twice.

## Verification

`make validate` passes on the branch (92 QML tests, JS and Python suites green). I confirmed the alias round-trip corruption and the duplicate parser by running both modules through `tests/load.js`, and read the three call sites above against `main`. The transport retry and TLS findings are read from the diff, not reproduced against a live server.

PR #65, opened 2026-08-30
Sign a new draft with the mailbox's own signature

maartendra:
A signature is set per mailbox on the settings page, under Writing. It is placed under a new message and above the quoted text in a reply, so a sign-off stays next to the words it signs rather than stranded below a screen of somebody else's message. It is sent as typed — no `-- ` line is added in front of it, because a client that adds one turns a signature into two decisions, and the line is one keystroke away for anybody who wants it.

## Why it sits on the account

Two mailboxes are two identities, and one sign-off under both is wrong for whichever it was not written for. So the field is on the account rather than in the window preferences, beside the label — the other thing on an account the user chose and the server did not. Nothing about it is secret; the keyring still holds what is.

A signature is also the one field here that legitimately contains newlines, and `accounts.json` crosses a line-oriented pipe on the way to disk. The existing JSON escaping is what keeps a two-line sign-off from truncating the account list, and there is now a test holding that rather than a comment.

## Invariants

`Mail.composeBody` builds every starting body, so a signature and a quote are joined the same way instead of being special-cased against each other, and the three call sites that used to concatenate by hand agree by construction. With no signature set it returns exactly what the caller built before — that equivalence is what makes the field optional rather than a new shape of draft, and it is asserted directly.

A saved draft already carries whatever it was written with, so reopening one leaves it untouched. A `mailto:` is a new message and is signed like one.

The editor saves when focus leaves it rather than on every keystroke. The write rebuilds the rows the `Repeater` built, so a save per character would reassign the text under the cursor; by the time the rows come back the stored value is what is already on screen, and an identical save is not a write at all.

## Verification

`make validate` — node tests, source regressions, `qmllint`, `omarchy plugin validate` — all pass, on Omarchy with Qt 6 and the shell's `qs.Ui` on the import path.

Eight new cases. Five are QML, which is where the compose body and the settings field can be reached at all: an unsigned account composes exactly as it did before, a new message and a `mailto:` open signed, a reply signs above the quote, and a reopened draft is not signed twice. The settings case also instantiates `SettingsPage` itself, so the new field is built rather than only linted. Three are node, over the model and the body builder: trimming the ends but not the middle, clearing, an unknown id changing nothing, and a multi-line signature surviving `serialize` and `load`.

Not covered: I have one Gmail mailbox, so the multi-account behaviour is exercised by the model tests and the per-account field rather than by a second live account.

## Release Notes

- Each mailbox can now have its own signature, set on the settings page under Writing.
- A signature is placed under a new message and above the quoted text in a reply, and is sent exactly as typed — no separator line is added.
- Reopening a saved draft does not sign it a second time.


huacnlee (changes requested):
The design is right and the writing around it is the best in the repository — `Mail.composeBody` as the one place every starting body is built, the equivalence with no signature set asserted directly rather than described, and a reason given for the per-account decision rather than a preference stated. The `-- ` argument is correct and the comment earns it.

Three things stop it, and the first two are the same shape: a value that is placed rather than typed is being read as a value the user chose.

I merged `main` into the branch to clear the conflict with #59, so the head here is `59eb982` and `make validate` passes on it — 103 QML tests. Every finding below was reproduced with a probe test on that head, not read off the diff; the probe output is quoted with each.

## An untouched new message is uploaded as a draft

`components/ComposeView.qml:398` now writes the body unconditionally, so with a signature set a brand-new compose window opens holding `"\n\n" + signature`. `hasMeaningfulDraft()` at line 477 asks `String(bodyEdit.text).trim() !== ""`, and that predicate is the only gate in `App.qml:324`:

```qml
if (!service || !compose.hasMeaningfulDraft()) { compose.finish(); return }
```

Both Back (`App.qml:1215`) and Escape (`App.qml:517`) go through it. So: press the compose key, change your mind, press Escape — a draft containing nothing but your own sign-off is uploaded to the server and "Draft saved" appears. Once per abandoned compose, and the Drafts mailbox keeps them.

```
FAIL!  : Signature::test_probe_untouched_signed_new_message_is_not_a_draft()
FAIL!  : Signature::test_probe_untouched_signed_mailto_is_not_a_draft()
PASS   : Signature::test_probe_untouched_unsigned_new_message_is_not_a_draft()
```

`compose/Recovery.js:71` carries a second copy of the same predicate, reached from `onDraftChanged: root.scheduleComposeRecovery()`, so a crash-recovery file is written for the same untouched window and offered back at the next launch.

A reply was already "meaningful" because of the quote, so this is specifically new messages and `mailto:`. What separates the two cases is not a length or a trim — a user who typed and then cleared back down to the signature has written something — it is whether the body is still the one `begin` placed. Remembering that string and comparing against it answers both this and the next finding.

## The signature field is destroyed under the cursor by a routine poll

`components/SettingsPage.qml:234` builds its Repeater over `service.accountSummaries`, and `Service.qml:541` rebuilds that array from `unread`, `active`, `signedIn`, `busy` (`host.listLoading`) and `error`. `MailAccount`'s poll timer flips `listLoading` twice every cycle, so the array is replaced — and every delegate in the Repeater is destroyed and rebuilt — on a timer, with no account having changed.

The comment at lines 281-287 reasons about this correctly for the case it considered: the write causes the rebuild, so the field is filled once and saved on the way out rather than bound. But the rebuild is not only ever caused by the write. When it happens on a poll instead, `Component.onDestruction` force-saves half a sentence, and the replacement's `Component.onCompleted` re-reads `modelData.signature` from the array computed before that save — so the text on screen reverts to what was stored, and the keyboard is gone with it.

```
FAIL!  : Signature::test_probe_a_poll_does_not_disturb_a_signature_being_typed()
   Actual   (): Maarten
   Expected (): Half a sign-
```

That is a signature being typed at the moment a poll lands: the field reverted mid-word. The probe does nothing but replace `accountSummaries` with an identical list carrying one flipped `busy` flag — no save, no account change.

A model that only changes when a signature does would hold still through this. The account list is the right dependency; the summaries array is not, because it carries live mailbox state on purpose.

## From reaches every mailbox, and the sign-off does not follow it

`Service.sendIdentities` (`Service.qml:606`) is built from every account, so the From menu offers another mailbox's address and `chooseFrom` calls `switchTo` to make it active. `accountSignature` follows — but the body does not, and nothing rewrites it.

```
qml: BODY AFTER SWITCH: "\n\nMaarten"
FAIL!  : 'body should carry the sending mailbox's sign-off, got "\n\nMaarten"'
```

Started under `me@example.com`, sent from `work@example.com`, signed `Maarten`. That is the case the description gives as the reason the field is on the account: "one sign-off under both is wrong for whichever it was not written for". The Verification section is candid that multi-account is the untested part, and this is what is in there.

The comment at line 47 is right that a binding is the wrong fix — it would overwrite what the user had changed. The condition it is missing is the same one as the first finding: while the body is still exactly what was placed, replacing it is not overwriting anything.

## Smaller

`components/SettingsPage.qml:227` — the "Signature" heading and the paragraph under the Repeater are unconditional, but the only delegate is hidden by `visible: String(modelData.id || "") !== ""`. On first run, before any mailbox is signed in, WRITING shows a heading and an explanation with no field between them.

## What I checked and cleared

`Accounts.setSignature` trims the ends and not the middle, keeps every other field on the account, does not mutate the list it was given, and is a no-op for an id that is not there. After #59's rework of draft handling, `beginDraft`'s `mode === "draft"` guard and `restoreDraft`'s verbatim write still mean a saved draft is never signed twice — I re-ran the branch's own tests against the merged head and they hold. The equivalence that makes the field optional — no signature set produces exactly the previous body — is asserted in the tests and is true.

maartendra:
Thank you for the probes — all three reproduced, and the framing that they were one shape of bug was the useful part. `placedBody` is the body exactly as `placeBody` left it, and each finding is that same question asked somewhere else.

**An untouched new message is uploaded as a draft.** `hasMeaningfulDraft` now asks whether the body still equals what was placed rather than whether it trims to anything. `Recovery.hasMeaningfulDraft` gets the same predicate and `placedBody` is carried on the recovery row; a row written before the field existed has none, which reads as an empty string, so any body at all differs from it and an older draft is still offered back.

**The signature field is destroyed under the cursor by a routine poll.** You were right that the account list is the dependency and the summaries array is not. There is now a `Service.accountSignatures` built from `accountList` alone, and the Repeater and the heading both hang on it. My comment reasoned about the rebuild the write causes and simply did not consider that anything else caused one.

**From reaches every mailbox, and the sign-off does not follow it.** While the body is still the one placed there is nothing of the user's in it, so it is rebuilt for the new mailbox; the moment they type, it is theirs and left alone. Rebuilt rather than patched — with a quote under it the old signature is at no known offset, and searching for its text would find a sign-off quoted from somebody else.

**Smaller.** The heading and the note now hang on the same list as the fields, so first run shows neither.

`make validate` passes: 109 QML tests. Five new cases cover the findings directly — an untouched signed compose and an untouched signed `mailto:` are not drafts while a sentence above the sign-off is, a reply is a draft on its quote regardless, changing the sending mailbox re-signs an untouched body and leaves a typed one alone, a poll that replaces the summaries disturbs neither the text nor the keyboard in it, and the section is absent with no mailbox signed in. Two more in node over the recovery predicate, including the older-row case.

huacnlee:
This branch has conflicts that must be resolved

huacnlee (approved):
Rechecked the requested changes against the reproduced cases. The untouched-compose, polling rebuild, From/signature, and empty-settings cases are covered and pass. During review I found that body equality alone did not preserve an edit that was deleted back to the placed signature; the merge commit now records actual user edits and carries that state through recovery, with QML and node regressions. Resolved the SettingsPage conflict against current main. make validate passes (169 QML tests).

PR #69, opened 2026-09-01
Preserve invitation responses during message refresh

eliasfaltin:
## Why

Reopening an invitation must retain the response already sent. Google calendar attachments arrive separately, which previously erased the cached response during refresh.

## Fix

`MailAccount` preserves the response while live message details and calendar attachments arrive. Cache writes wait for the existing cached response. An attachment reference cannot replace the stored invitation with null.

This revision limits the change to response preservation. Automatic CalDAV import is removed from the diff. The reported recurrence, calendar selection, refusal, and write-timeout concerns therefore leave no new calendar writes in this PR.

## Blast radius

A response belongs to the selected message. The existing selection serial rejects stale callbacks after navigation. The parser preserves responses only for matching invitation UIDs.

## Testing

- Six arrival-order regressions passed against the actual `select` and `loadInvite` methods, including delayed and missing attachments.
- The regression fails against upstream main because ACCEPTED becomes unanswered.
- Full QML suite: 581 passed, plus 4 native HTTP and 64 sidebar checks.
- Calendar parser tests, source checks, `qmllint`, and plugin validation passed.
- `make validate` reaches an existing failure in the unchanged AI bridge suite. The same 10 failures reproduce on upstream main at `77db4b1`.

Latest upstream main is merged without rewriting branch history.


huacnlee (changes requested):
Both halves are worth having, and the diagnosis behind the first one is exactly right: the invitation inside a message never changes, so a live read was always going to put the organiser's original REQUEST back over an answered card. Carrying the response forward on a UID match is the right shape for that, and extracting `bodyRecord` so the three writers cannot disagree about what a record holds is a real improvement on its own.

`make validate` passes on a clean checkout of this branch (91 QML tests), and it merges cleanly with `main`.

Three things stop it. The first is the one most users will meet.

## An IMAP mailbox with no CalDAV calendar is told it failed

`importInvitation` finds its target with `Sources.invitationTarget`, which returns `null` when no CalDAV calendar is configured. `Calendar.writeRefusal(null, null)` then answers `"Choose a calendar"`, and `Service.qml:830` turns that into `host.fail(...)`:

```
S.invitationTarget({ sources: [{ id: "google:me", kind: "google", enabled: true }] }, "imap")  ->  null
C.writeRefusal(null, null)  ->  "Choose a calendar"
```

So an IMAP user who has never set up a CalDAV calendar — the default state, and the state of anybody using this on Fastmail or iCloud without also adding their CalDAV URL — accepts an invitation and sees two notices: `Answer sent to organiser@example.com`, and then `Choose a calendar`. The RSVP succeeded. Nothing failed. And the sentence names an action that does not exist in that flow: there is no picker to choose from, because the import is automatic.

Having no calendar to copy into is not a refusal, it is the absence of the feature. It should return quietly, the way the Gmail and HEY exclusion already does one line above it.

## Declining after accepting leaves the event in the calendar

`MailAccount.qml:1853` emits `invitationAccepted` only for `"accepted"`, and nothing removes the event for `"declined"` or `"tentative"`. Because the import uses the invitation UID as the resource identity, the event is already on the server by then — so a user who accepts and then changes their mind has the meeting in their calendar with no way to notice, and the card says Declined.

The description's idempotence argument covers accepting twice, which it does correctly. It is changing the answer that is not covered. `deleteEvent` already exists and is guarded the same way, so the missing case has a path to run down.

## The stored UID is not the invitation's UID

`importInvitation` builds the body with `veventLines(icsText(uid), ...)` but returns the unescaped `uid`, and the resource path is derived from the unescaped one. For a UID carrying any of `\ , ;` the two disagree:

```
returned uid : "meeting;1,2\\x"
ics UID line : "UID:meeting\;1\,2\\\\x"
href         : https://dav.example.com/cal/meeting%3B1%2C2%5C%5Cx.ics
```

Idempotence survives, because the href is computed from the same raw UID every time. What does not survive is the correspondence RFC 5546 is built on: another client, or the server's own scheduling, matches an event to a meeting by UID, and the one on disk is no longer the one in the mail. Note that `createEvent` does not escape its own UID — it does not need to, since it generates `omamail-<digits>` — so the two paths disagree about whether a UID is TEXT.

The escaping itself is right and should stay: I checked that an organiser-controlled UID carrying a CRLF cannot inject properties into the VEVENT because of it. Escape for the body, and keep the raw value for the identity — or escape both and derive the href from the escaped one.

## What I checked and cleared

The reschedule case I went looking for is not reachable: all three `preserveResponse` call sites compare a fresh parse of one message against the cached copy of that same message, and `MailAccount.qml:948` clears `selectedInvite` before a new message loads, so an answered invitation cannot be carried onto a later REQUEST with the same UID. The UID equality check is what would hold that line if the clearing ever moved, which is worth the comment it already has.

`bodyRecord` changes `html: rawHtml` to `html: sourceHtml` in the detail path. Those are the same string at that point in both branches — `renderSource` assigns `sourceHtml = rawHtml`, and the branch is skipped only when they were already equal — so the extraction is behaviour-preserving.

The `detailLive` branch writes a record from `selectedBody`, `selectedAttachments` and `sourceHtml`, which are all populated by the live fetch before `detailLive` is set, so it cannot persist a half-filled record.

An organiser-controlled UID cannot escape the CalDAV collection: `caldavEventUrl` percent-encodes it, which is the rule AGENTS.md states about credentials going only to the configured source's own origin, doing its job.

huacnlee:
A second pass turned up more, and two of them belong above the three I listed. Both are in the half this PR exists for.

## The RSVP is erased for exactly the invitations Google sends

`loadInvite` runs only when `Calendar.pendingPart` is non-null, and `pendingPart` returns a part only when it has no `body.data`. `fromPayload` on that same part decodes nothing, finds no `BEGIN:VCALENDAR`, and returns `null` — so by the time `loadInvite`'s callback runs, line 1044 has already assigned:

```qml
root.selectedInvite = Calendar.preserveResponse(null, cached, addr)   // -> null
```

`preserveResponse` returns its first argument when either is falsy, so `selectedInvite` is `null`, and the new call in the callback is `preserveResponse(fromAttachment(...), null, ...)` — which returns the fresh, unanswered invite. `record.invite = invite` and `bodyCache.put(messageId, record)` then write that over the answered copy on disk.

So for a message whose calendar part is delivered as a reference rather than as octets — which the comment at line 1052 says is what Google's own organiser calendar produces — this does not merely fail to preserve the response. It **destroys the one already stored**, permanently, on the next open. The race branch at line 976 is a no-op for the same reason: `selectedInvite` is `null` there too, so `restored !== root.selectedInvite` is false and `cached.invite` is dropped.

The cached invite has to be held somewhere that line 1044 does not clear, and read from there in both places.

## Accepting one moved occurrence destroys the series

`importInvitation` does not read `invite.recurrenceIdMs`, and the CalDAV write PUTs to `base + encodeURIComponent(uid) + ".ics"` unconditionally.

An organiser moves one occurrence of a weekly meeting. That arrives as a `METHOD:REQUEST` carrying `RECURRENCE-ID` and no `RRULE`, with the series' own UID. Accepting it builds a standalone one-off VEVENT and overwrites the resource holding the whole series — the recurring event is gone, replaced by the single moved instance. `writeRefusal(source, null)` is called with a null event here on purpose, so the existing "Recurring CalDAV events can only be changed in a full calendar client" guard does not fire.

## Four smaller ones

`Calendar.js:511` — `importInvitation` reads `event.start.ms` without asking `start.resolved`. `resolveTime` returns `resolved: false` with the wall clock read as UTC when a `TZID` has no usable inline `VTIMEZONE`, and `veventLines` writes that as `DTSTART:<naive>Z`. The card correctly warns the reader that the zone is unresolved; the calendar event is silently hours out. An unresolved time is a reason to refuse the import, not to write a guess.

`Sources.js:265` — `invitationTarget` takes the first enabled writable CalDAV source and never looks at `accountId`, and `forAccount` keeps every non-Google source regardless of owner, so the candidate list is not account-scoped either. Two IMAP mailboxes with a calendar each: a work invitation lands in whichever calendar is first in `calendars.json`. Prefer one owned by the accepting account, and fall back only when there is none.

`Calendar.js:520` — `recurrenceRule` is carried across but `excludedMs` is not, so a series with cancelled dates is imported showing those occurrences as still happening.

`Calendar.js:527` — the generated ICS is never folded. RFC 5545 wants content lines folded at 75 octets, and `message/Calendar.js` already has `foldIcsLine` for this on the reply path; `calendar/Calendar.js` has none. A Google invitation's `DESCRIPTION` — several hundred characters plus the Meet block — goes out as one long line, and a strict server can refuse the PUT. This one predates the PR, since `createEvent` shares `veventLines`; it becomes reachable here because the text is now the organiser's rather than something the user typed into a field.

## And two on the guard itself

The `creatingEvent || eventWriting` refusal reports through `invitationImported`, which becomes `host.fail(...)` on the mailbox, and the import is then dropped with no retry — save an event in the composer while accepting an invitation in the other pane and you get an error on the mail account for something you did not do, with the event silently never added.

In the other direction, `startEvent` sets `creatingEvent` for the background import, and the composer binds its busy spinner and its Save button to that flag. The CalDAV create path never arms `eventDeadline`, so a hung `calendar-write.sh` holds it indefinitely: the composer opens with Save disabled and a spinner running, and every create, update and delete is refused, with nothing on screen saying why.

huacnlee:
Thanks for the diagnosis, it was right. #174 closed this race structurally: the Rust reader now reads the body cache while it renders the live message and returns the stored invitation as cachedInvite in the same reply, MailAccount prefers it over the live REQUEST and only fetches a referenced calendar part when there is no cached copy, and Rsvp writes the answered invitation back to the cache. None of the files here exist any more, so closing as superseded. main has no regression test for reopening an answered invitation yet; a test against the new reader path would be a welcome follow-up.

PR #71, opened 2026-09-01
Show all accessible Google calendars, not just primary

### LINKED ISSUE #53: Support multiple Google calendars per Gmail account
## Problem

Omamail currently shows only the primary Google Calendar for a signed-in Gmail
account. Users who have secondary, shared, or subscribed calendars cannot view
or select them in Omamail.

## Steps to reproduce

1. Add and sign in to a Gmail account.
2. Add or subscribe to another calendar in Google Calendar, for example
   “Work” or “Family”.
3. Open the Calendar view in Omamail.
4. Open Settings → Calendars.

Only the primary calendar is available. The “Add a calendar” option only
accepts CalDAV details and does not provide a Google Calendar selector.

## Expected behavior

Omamail should retrieve the calendars available to the signed-in Google
account and provide per-calendar controls to:

- show or hide individual calendars
- display calendar names and colors
- keep read-only/shared calendars read-only
- choose an appropriate writable calendar when creating events

The primary calendar should remain enabled by default.

## Possible implementation

Use the Google Calendar `calendarList.list` endpoint to retrieve calendar IDs,
then fetch events from `/calendars/{calendarId}/events` for the selected
calendars. Persist the selected calendar IDs and visibility settings locally.

The OAuth flow may need the `calendar.calendarlist.readonly` scope in addition
to the existing event scope.

## Environment

- Omarchy:
- Omamail version: 0.5.0

Roddygithub:
Fixes #53

## Problem

Omamail only queries `/calendars/primary/events`, so secondary, shared, or subscribed Google calendars never appear. The Calendar API and OAuth scope were already enabled, but the code never asked for the calendar list.

## Solution

1. **Discover calendars**: call `calendarList.list` to get every calendar the account can access.
2. **Fetch events from each**: iterate the list and query `/calendars/{calendarId}/events` for each enabled calendar.
3. **New OAuth scope**: add `calendar.readonly` alongside the existing `calendar.events` scope.
4. **Keyring grant version bump**: force re-authentication so the new scope is actually granted.
5. **Safe URL encoding**: all `calendarId` values are percent-encoded in API URLs.

## Changes

- `providers/OAuth.js` — add `calendar.readonly` scope
- `calendar/Calendar.js` — add `googleCalendarsUrl()`, parameterize `googleEventsUrl()` and `googleEventUrl()` with `calendarId`
- `calendar/CalendarController.qml` — discover calendars, iterate and fetch events per calendar, preserve `sourceName`
- `providers/Credentials.js` — bump keyring grant to force re-auth
- Tests updated to match

## Verified

- `make test-js` passes
- `make validate` passes (qmllint warnings are pre-existing)
- Tested locally: secondary Google calendars now appear in the calendar view

## Notes

- Event creation still defaults to the primary calendar; edits/deletes use the event's discovered `calendarId`.
- Existing Google sessions will need to sign in again to grant the new scope.

### DIFF
--- a/calendar/Calendar.js
+++ b/calendar/Calendar.js
@@ -347,7 +347,7 @@ function googleMoment(value, dateOnly) {
   return { ms: ms, allDay: dateOnly, tzid: "", resolved: true }
 }
 
-function eventsFromGoogle(payload, sourceId) {
+function eventsFromGoogle(payload, sourceId, calendarId) {
   var items = payload && Array.isArray(payload.items) ? payload.items : []
   var out = []
   for (var i = 0; i < items.length; i++) {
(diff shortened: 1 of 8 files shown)


huacnlee (changes requested):
Four findings block this as it stands.

1. [P1] Preserve each calendar access role and refuse writes for read-only entries. calendarList.list returns reader and freeBusyReader calendars as well as writable ones, but startGoogleCalendar() carries only id and summary into each event. Every event still resolves to the one synthesized account-level Google source, whose readOnly is false. CalendarEventDetail.canWrite therefore draws Edit/Delete and Calendar.writeRefusal accepts them, only for Google PATCH/DELETE to fail after the user commits. Carry accessRole (or a derived per-event readOnly) through parsing and use it in both the button rule and controller refusal. Google documents only writer, writerWithoutPrivateAccess, and owner as writable roles.

2. [P1] Page each calendar events before advancing the queue. googleEventsUrl() asks for 2500 and startGoogleCalendar() immediately advances to the next calendar after one response, ignoring payload.nextPageToken. Google explicitly says a page may contain fewer than maxResults even when more matches exist, and a non-empty nextPageToken is the only completeness signal. Busy calendars therefore silently lose every event after the first page. Add a page token to the URL/state and keep accumulating the same calendar until the token is empty.

3. [P1] Fix the new install-test expectation. tests/test_install.sh:90 adds two literal spaces before store inside expected_calls, but the fake secret-tool writes no such indentation. bash tests/test_install.sh and consequently make validate fail with keyring-store.sh did not replace the unversioned grant before storing the versioned one. Remove those spaces.

4. [P2] Give aggregated events a calendar-qualified UI identity. eventsFromGoogle() keeps uid = iCalUID, while CalendarView selects, navigates, highlights, and opens events solely by that value. Once multiple calendars are aggregated, two copies/imports of the same iCalendar event can legitimately share an iCalUID. Both rows then highlight together and keyboard activation always opens the first. Preserve uid for calendar semantics, but add and use a stable UI key that includes calendarId (and occurrence identity where applicable).

The PR also currently conflicts with main. I ran the branch as submitted: make validate stops at finding 3.

huacnlee:
We have refactor entire of backend by use Rust, this PR has been outdated. Sorry I'm going to close this, if you want keep going, you can reopen a new PR.

PR #50, opened 2026-08-27
Keep message headers out of IMAP failures

alberthild:
## Summary

- inspect literal-aware IMAP response records before deciding that a request failed
- accept only tagged `NO` or `BAD` completions, not matching text inside fetched message bodies
- cover untagged statuses, continuation responses, and a fetched message carrying `X-Spam-Flag: NO`

The previous multiline regular expression searched the complete `FETCH` response. A normal message header such as `X-Spam-Flag: NO` therefore looked like an IMAP command failure, so the successfully fetched body was discarded and the reader showed “This message has no text to show”.

Fixes #42.

## Testing

- `make test` (all JavaScript and shell tests; 81 QML tests passed)
- verified against an affected multipart IMAP message containing `X-Spam-Flag: NO`

## Release Notes

- Fix IMAP messages with spam verdict headers appearing to have no readable body.

PR #75, opened 2026-09-02
Read the tagged completion, not any line the message happens to contain

### LINKED ISSUE #74: IMAP: a message's own X-Spam-Flag header is read as a failed FETCH
**What happens**

Opening a message over IMAP sometimes shows "This message has no text to show" for a message that plainly has text, and the status line carries a fragment of the message's own header as if it were a server error — for example `X-UI-Out-Filterresults: notjunk:1;M01:P0:...` or similar.

**Why**

`ImapProtocol.js`'s `isFailure` / `failureDetail` regex-scan the *entire* raw FETCH response text, including the literal payload — which is the message itself. Any header line that happens to match `/^\S+\s+(NO|BAD)\s/im` gets read as a tagged `NO`/`BAD` completion.

GMX stamps every delivered message with:

```
X-Spam-Flag: NO
```

which matches that pattern exactly. The client reports a successful FETCH as a failed command, discards the parsed body/html, and shows the offending header line as the error detail.

**Repro**

1. IMAP account against GMX (or any server whose antispam scanner adds an `X-Spam-Flag: NO`-shaped header — this is a common convention, not GMX-specific).
2. Open any message.
3. Reader shows "This message has no text to show"; status line shows a chunk of the message's own header instead of a real error.

**Confirmed via raw IMAP** — fetched a plain single-part text/plain test message via a separate IMAP client and found:

```
X-Spam-Flag: NO
```

sitting among its headers, matching `isFailure`'s regex directly.

**Fix**

`splitResponse` already folds each literal into the one line it belongs to, specifically so a regex never has to guess where a literal ends. `isFailure`/`failureDetail` weren't going through it — they ran directly against the raw text instead. Routing them through `splitResponse` and checking only the last line (the only line that can be a tagged completion) fixes it. PR incoming.

Andre1Becker:
Fixes #74

## Problem

`isFailure` and `failureDetail` in `ImapProtocol.js` regex-scanned the whole raw FETCH response text — literal payload included, which for a FETCH is the message itself — for a line matching `/^\S+\s+(NO|BAD)\s/im`. A message header that happens to read like a tagged `NO`/`BAD` completion trips it. GMX stamps every delivered message with `X-Spam-Flag: NO`, which matches directly: a successful fetch gets reported as a failed command, the parsed body/html is discarded, and a fragment of the message's own header shows up as the error text. The reader then shows "This message has no text to show" for a message that has text.

Confirmed against a real account: fetched a plain test message over IMAP directly and found `X-Spam-Flag: NO` sitting among its headers, matching the regex `isFailure` runs.

## Fix

`splitResponse` already exists to fold each literal into the one response line it belongs to — the file's own comment above it says why: "where a literal ends is the one thing this cannot be wrong about, and no regular expression can count octets." `isFailure`/`failureDetail` weren't going through it. They now do, and check only the last line `splitResponse` produces — the only line that can be a tagged completion, since an untagged line (including one holding a literal) can never be the final line of a single command's response.

## Verification

- `make test-js` and `make test-shell` both pass, unchanged.
- Added a manual regression against a synthetic FETCH response with `X-Spam-Flag: NO` embedded in the literal: `isFailure` now returns `false` and `failureDetail` returns `""`, where it previously misfired.
- Existing `isFailure`/`failureDetail` unit tests in `tests/test_imap.js` are unaffected — all inputs there are single-line, so the last-line lookup returns the same value the old regex did.

## Release Notes

- Fixed messages sometimes showing as empty ("This message has no text to show") with a garbled error line, caused by a false-positive match on a header inside the message itself (e.g. GMX's `X-Spam-Flag: NO`).

https://claude.ai/code/session_01AbQ5kH613UMeECpu9vchoo

### DIFF
--- a/providers/ImapProtocol.js
+++ b/providers/ImapProtocol.js
@@ -898,12 +898,25 @@ function redact(text) {
 
 // Whether the tagged completion said OK. curl hides the tag from us for a
 // custom request, so this reads what it does surface.
+//
+// Only the *last* response line can be that completion — mid-response lines
+// are untagged data, and with a FETCH that data is the message itself. A
+// server-side spam scanner header like "X-Spam-Flag: NO" reads exactly like
+// a tagged NO to a regex that isn't told where the literal ends, so this
+// goes through `splitResponse` first: it already folds each literal into the
+// one line it belongs to, which is the only way to reach the true last line
+// without a regex trying to count octets.
+function lastResponseLine(text) {
+  var lines = splitResponse(text)
+  return lines.length ? lines[lines.length - 1] : ""
+}
+
 function isFailure(text) {
-  return /^\S+\s+(NO|BAD)\s/im.test(String(text || ""))
+  return /^\S+\s+(NO|BAD)\b/i.test(lastResponseLine(text))
 }
 
 function failureDetail(text) {
-  var match = String(text || "").match(/^\S+\s+(?:NO|BAD)\s+([\s\S]*)$/im)
+  var match = lastResponseLine(text).match(/^\S+\s+(?:NO|BAD)\s+([\s\S]*)$/i)
   return match ? trimmed(match[1].split(/[\r\n]/)[0]) : ""
 }
 


huacnlee:
The diagnosis here is right, and `main` already carries this fix: #50 (merged 2026-08-28, before this PR was opened) replaced the whole-text regex with `failureCompletion`, which goes through `splitResponse` and tests only the first protocol line of each folded response. `tests/test_imap.js` there pins the exact `X-Spam-Flag: NO` FETCH literal from this report, plus a real tagged `NO` arriving after such a literal.

The conflict is only in `providers/ImapProtocol.js`, and resolving it means keeping `main`'s version, which leaves this PR with no diff at all.

I tried the other resolution too, applying this PR's `lastResponseLine` hunk onto current `main`. It fails `tests/test_imap.js:737`: `/^\S+\s+(NO|BAD)\b/` also matches an untagged `* NO [ALERT] ...` or a continuation `+ NO ...` when one happens to be the last line, and `main`'s tag pattern excludes those on purpose.

So closing this as superseded by #50. Thanks for the report and the reproduction against a real account — that is what #50 was verified against. If a build from `main` (0.7.0 or later) still shows an empty message with a garbled error line, please say so on #74 with the response shape and I will look again.

PR #81, opened 2026-09-02
Add an omamail command and the agents skill that uses it

keejkrej:
Agents and scripts can talk to the same mailboxes the Omamail window is signed into, through `omamail` on the command line.

The plugin installs `~/.local/bin/omamail` when it loads, and `~/.agents/skills/omamail` next to it. The command is another client of the existing accounts file, GNOME Keyring entries, and provider libraries — Gmail, HEY and IMAP — not a second protocol stack and not a socket into the running shell. The skill is what an agent reads first: `--json`, IMAP search is Inbox only, quote ids whose folder names contain spaces.

The service is on-demand, so an agent should not depend on the window being up. `GmailApi.js`, `ImapProtocol.js` and `HeyCli.js` already own every string that reaches a server; `cli/Cli.js` parses flags and names commands, and `cli/run.js` is the process.

## Release Notes

- `omamail` lists, reads, sends, replies, and triages mail from the terminal, using the mailboxes already signed into the window.
- Agents pass `--json` for a stable object on stdout. Message ids are the provider's own, never row numbers.
- A verb the mailbox cannot honour (star on HEY, archive on HEY, junk on IMAP) is refused before anything is sent.
- IMAP senders and message ids are validated before SMTP or mailbox actions; malformed ids never produce a partial action.
- IMAP mailbox names retain their exact server case; actual INBOX and SPECIAL-USE folders use the built-in keys, while ordinary case-colliding folders remain selectable.
- Human-readable output strips terminal control sequences while JSON preserves the original mailbox data.
- Installing the plugin also drops an agents skill in `~/.agents/skills/omamail`, so an agent uses that command instead of talking to Gmail, IMAP, or `hey` itself.

## Interface

Noun-verb, GNU flags, JSON opt-in, the shape `gh` and Himalaya already taught agents:

```
omamail account list
omamail mailbox list
omamail list --json
omamail read <id>
omamail send --to jane@example.com --subject "Hello" --body "Hi Jane"
omamail reply <id> --body "Thanks" --all
omamail archive <id>
omamail star <id>…
omamail search 'from:jane newer_than:7d'
```

`omamail --help` is the live map. `docs/CLI.md` is the same contract written down.

Parsing and output shape live in `cli/Cli.js` and `cli/View.js` (pragma libraries, node-tested). `cli/run.js` is the process: curl config on stdin so tokens never reach the process table, `mail-transport.sh` for IMAP/SMTP, `hey` for HEY.

`agents/SKILL.md` is the agent-facing counterpart. `register-cli.sh` symlinks that directory to `~/.agents/skills/omamail`. It does not copy the command table out of `--help`; it names the surprises `--help` does not.

## Demo

Signed an IMAP mailbox in through the window, then used only the CLI for a triage task: find every message associated with one collaborator's address, and star them.

Surfaces used, in order:

1. `omamail --json account list` — confirmed the IMAP mailbox was signed in and active. Sign-in itself stays in the window; the CLI only reads the accounts file and the keyring.
2. `omamail --json mailbox list` — listed the rail keys this server actually has (`inbox`, `sent`, `starred`, plus the server's own folders). That is how the next two calls knew which `--mailbox` values were real.
3. `omamail --json search ADDRESS --limit 100` — IMAP typed search. The provider turns that into `folder:INBOX TEXT "…"`, so it searches Inbox headers and body, not every folder. It returned the full Inbox set in one page (estimate matched the row count, empty `nextPageToken`).
4. `omamail --json message list --mailbox sent --limit 100` and `--page-token` — Sent is a different IMAP folder, so `search` cannot see it. Paged the whole Sent listing and kept rows whose From/To/Cc carried the same address.
5. `omamail --json message star <id>…` — starred the Inbox hits and the Sent hits in two calls. Ids are `<uid>:<folder>`; a folder name with a space is one argv token, so the ids have to be passed quoted. An attempt that splits on the space is refused locally as invalid input (exit 2), before the CLI opens a server connection.
6. The same `search` and `message list --mailbox sent` / `--mailbox starred` again — every Inbox hit now had `"starred": true`. The starred rail is `folder:INBOX FLAGGED`, so Sent flags do not appear there; they stay on the Sent rows.

The CLI did not invent a cross-folder search. Inbox used `search`; Sent used `message list --mailbox sent`; starring used the ids those two surfaces returned.

## Invariants

- Sign-in stays in the window. The CLI does not start OAuth, ask for an IMAP password, or run `hey auth login`.
- Secrets still cross on stdin. Refresh tokens and IMAP passwords are read with `secret-tool lookup`; Gmail HTTP uses curl `--config -`.
- Provider differences still end at `Registry.capabilities`. An undeclared verb exits 5.
- `cli/run.js` is the one `.js` file that is not a pragma library, because it is a Node process. `tests/test_source.sh` exempts it by name.
- HEY's thread read returns the conversation body, not the subject or sender — that is how `hey threads` answers. List first, then read.

## Verification

- `make test` and `make validate` pass, including 172 QML tests and CLI regressions for provider reads, mailbox selection, send authorization, terminal output, and documented exit statuses.
- `tests/test_link_plugin.sh` and `tests/test_cli.sh` assert `register-cli.sh` writes the `omamail` symlink and the `~/.agents/skills/omamail` skill, and refuses to replace a regular file.
- Live IMAP: installed this checkout with `make install` (validate, symlink into `~/.config/omarchy/plugins/omamail`, enable the plugin, put `omamail` on PATH). After signing the mailbox in through the window, the demo above listed accounts and mailboxes, searched Inbox, paged Sent, starred both sets, and re-read the flags. Gmail and HEY were not exercised live in this pass.


huacnlee:
I’m closing this because the CLI expands OmaMail beyond its intended boundary as a [Quickshell plugin hosted by `omarchy-shell`](https://github.com/omacom/omarchy/blob/quattro/docs/omarchy-shell.md).

Programmatic access to mail is a useful direction, but bundling a standalone command, a second application entry point, and an agent skill into the plugin is not the right architecture for it. OmaMail should remain focused on the service and UI hosted by the shell.

Omarchy defines [IPC as the standard boundary for commands that communicate with the running shell](https://github.com/omacom/omarchy/blob/quattro/docs/omarchy-shell.md#ipc), while its plugin installer deliberately [clones, validates, and enables plugins without running installation hooks](https://github.com/omacom/omarchy/blob/quattro/manual/32-shell-plugins.md#adding-a-plugin-from-git). A broader automation interface therefore needs a separately designed integration boundary rather than becoming part of the plugin’s runtime and distribution.

PR #89, opened 2026-09-03
Scroll as far as the wheel was turned

patrykbart:
A `Flickable` answers each wheel event with a flick it then decelerates, so the distance depends on how the turn was chopped up rather than on how far the wheel went.

Measured against a real `Flickable`, one notch arriving as a single `angleDelta` of 120 moves about 72 pixels. The same notch from a high-resolution mouse, which reports it as eight deltas of 15, moves about 9 — each one starting and damping its own small flick. Eight times less for the same gesture, which is what "slower than every other application" turns out to be. Logitech's MX line reports this way, and so does a lot of recent hardware.

Rotation is the part that does not change: `angleDelta` is eighths of a degree, a notch is 15 degrees, and eight fractions of a notch still add up to 15. So the distance comes from rotation and nothing else — 120 pixels a notch, which is three lines of text and what a GTK application moves for the same turn.

Nothing is capped. A ceiling on one *event* would put the chunking dependence straight back at the coarse end: a free-spinning wheel delivers ten notches as one event, and a cap moves that a notch and a half while the same ten notches arriving as ten events move ten. A bound worth having would be per unit time.

## Which scrollers

Every vertical one: the message list, the reader, the settings page, the account setup page, the mailbox sidebar, the shortcut sheet, all three of compose's, the week grid, the contacts picker, the recipient suggestions, both calendar event sheets, and the label picker.

`MailboxTabs` is deliberately out — it is a horizontal strip, and a vertical wheel there is the wrong axis.

## What a scroller can reach

`contentY` does not run from 0 to `contentHeight - height`, which is what three separate clamps in this repository assumed.

Margins extend both ends, so a view resting in its top margin answered a scroll *up* by moving *down* to 0, after which the margin could never be seen again. And `originY` moves the start, so a `ListView` with a 200-tall header turned its first notch into a 200-pixel jump and could never wheel the header back into view.

`Model.contentYBounds` is the single answer now, and `contentYToReveal` and `ShortcutHelp.scrollBy` — the other two copies — ask it too. The `originY` half is measured: a `ListView` with a header settles at exactly `originY` and `originY + contentHeight - height`. The margin half is from Qt's documented semantics, since `StopAtBounds` does not correct a programmatic assignment and the same probe cannot reach it.

## Two choices worth stating

**A mouse only.** `acceptedDevices: PointerDevice.Mouse`, said out loud rather than left to the default. A touchpad reports `pixelDelta` and its own momentum and a `Flickable` tracks fingers with it properly, so where Qt can tell one apart this stays out of the way.

**An instant jump rather than an animated slide.** A wheel is a hand movement and should track it. The 160 ms animation elsewhere is for a jump to a section, which is a discontinuity the eye needs to follow; animating a wheel would also fight a repeating one, each event restarting an animation the next interrupts.

## Why the component is a handler

`WheelScroller` is a `WheelHandler` and not an `Item` holding one. A `Flickable` reparents its visual children into its content, so an `Item` dropped inside would be a zero-sized thing scrolling along with the list and its handler would never see a wheel event. A handler is not reparented — it attaches to the `Flickable`, which is what has to be got in front of.

## Verification

`make validate` green, 240 QML assertions.

`tests/test_model.js` covers the distance, that ten notches move the same however they arrive, and every bound — margins at both ends, a `ListView` header, and content shorter than its own view. `tests/qml/tst_wheel_scroll.qml` drives three real scrollers, each paired against a control that has to move differently. `tests/qml/tst_settings_sidebar.qml` covers a real call site in the real window hierarchy.

Mutation-checked two ways, because an earlier round of these tests passed for the wrong reason: with the component neutered 7 of 11 fail, and with the component intact but the bounds reverted to the old 0-floor clamp the 3 margin and header cases fail. Every assertion is caught by one or the other.

## Release Notes

- Scrolling moves as far as the wheel was turned. A notch now moves three lines of text, the same as other applications on the desktop, in every scrolling surface rather than some of them.
- High-resolution scroll wheels are no longer roughly eight times slower than the rest of the desktop, which is most recent hardware including Logitech's MX line.
- A view with margins, and a list with a header, can now be scrolled to their real ends rather than losing the first screenful.

### DIFF
--- a/App.qml
+++ b/App.qml
@@ -1368,8 +1368,6 @@ Item {
           Flickable {
             id: listFlick
 
-            // The Flickable's own wheel handling is a flick per event, which a
-            // high-resolution wheel turns into a seventh of the distance.
             WheelScroller { view: listFlick }
             anchors.fill: parent
             contentWidth: width
@@ -1603,8 +1601,6 @@ Item {
         Flickable {
           id: setupFlick
 
-          // The Flickable's own wheel handling is a flick per event, which a
-          // high-resolution wheel turns into a seventh of the distance.
           WheelScroller { view: setupFlick }
           anchors.fill: parent
           anchors.margins: Style.space(18)
@@ -1708,8 +1704,6 @@ Item {
         Flickable {
           id: settingsFlick
 
-          // The Flickable's own wheel handling is a flick per event, which a
-          // high-resolution wheel turns into a seventh of the distance.
           WheelScroller { view: settingsFlick }
           // The whole width, rail included: the wheel scrolls the page from
           // anywhere in the block, and the scrollbar keeps the window's edge.
--- a/App.qml
+++ b/App.qml
@@ -1367,6 +1367,10 @@ Item {
           // viewport, which would push the bar inward with it.
           Flickable {
             id: listFlick
+
+            // The Flickable's own wheel handling is a flick per event, which a
+            // high-resolution wheel turns into a seventh of the distance.
+            WheelScroller { view: listFlick }
             anchors.fill: parent
             contentWidth: width
             contentHeight: list.implicitHeight + Style.space(16)
@@ -1598,6 +1602,10 @@ Item {
         // the mailbox is connected.
         Flickable {
           id: setupFlick
+
+          // The Flickable's own wheel handling is a flick per event, which a
+          // high-resolution wheel turns into a seventh of the distance.
+          WheelScroller { view: setupFlick }
           anchors.fill: parent
           anchors.margins: Style.space(18)
           anchors.topMargin: parent.pageTop
@@ -1699,6 +1707,10 @@ Item {
 
         Flickable {
           id: settingsFlick
+
+          // The Flickable's own wheel handling is a flick per event, which a
+          // high-resolution wheel turns into a seventh of the distance.
+          WheelScroller { view: settingsFlick }
           // The whole width, rail included: the wheel scrolls the page from
           // anywhere in the block, and the scrollbar keeps the window's edge.
           anchors.left: parent.left
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ QML_FILES := Service.qml BarWidget.qml App.qml \
 	components/MessageMenu.qml \
 	components/MenuActionRow.qml components/MenuSeparatorLine.qml \
 	components/KeyRouter.qml \
+	components/WheelScroller.qml \
 	components/ActionIcon.qml \
 	components/IconButton.qml \
 	components/IconTextButton.qml \
--- a/account/Model.js
+++ b/account/Model.js
@@ -842,6 +842,50 @@ function settingsContentHeight(sections, pageHeight, viewportHeight) {
   return Math.max(page, known[known.length - 1].y + viewport)
 }
 
+// ------------------------------------------------------------------ the wheel
+//
+// How far a wheel turn moves a Flickable, which the Flickable itself gets
+// wrong on a mouse that reports finely.
+//
+// A Flickable answers a wheel event with a *flick* — a velocity it then
+// decelerates — so the distance depends on how the turn was chopped up rather
+// than on how far the wheel went. One notch arrives as `angleDelta` 120 and
+// moves about 43 pixels; a high-resolution wheel reports the same physical
+// notch as eight deltas of 15, each starting and damping its own little
+// flick, and the same turn of the same wheel moves about 6. Seven times less
+// for the same gesture, which is what "slower than every other app" is.
+//
+// Rotation is the thing that does not change: `angleDelta` is eighths of a
+// degree, a notch is 15 degrees, and eight fractions of a notch still add up
+// to 15. So the distance is computed from rotation and nothing else.
+//
+// 8 pixels per degree puts a notch at 120, which is what three lines of text
+// comes to and what a GTK application moves for the same turn.
+var WHEEL_PIXELS_PER_DEGREE = 8
+
+// A notch is 15 degrees, so this is the ceiling on one *event* rather than on
+// one gesture: a touchpad's flung two-finger scroll arrives as a stream of
+// large deltas, and without a bound a single one of them can cross a long list
+// in a jump nobody asked for.
+var WHEEL_MAX_DEGREES_PER_EVENT = 60
+
+function wheelDistance(angleDelta) {
+  var degrees = (Number(angleDelta) || 0) / 8
+  var limit = WHEEL_MAX_DEGREES_PER_EVENT
+  if (degrees > limit) degrees = limit
+  if (degrees < -limit) degrees = -limit
+  return degrees * WHEEL_PIXELS_PER_DEGREE
+}
+
+// Where the view lands, clamped to what it can reach. A list shorter than its
+// own viewport cannot scroll at all, and answering with a negative limit would
+// let it drift above its first row.
+function wheelScrollTarget(contentY, angleDelta, contentHeight, viewportHeight) {
+  var limit = Math.max(0, (Number(contentHeight) || 0) - (Number(viewportHeight) || 0))
+  var next = (Number(contentY) || 0) - wheelDistance(angleDelta)
+  return Math.max(0, Math.min(limit, next))
+}
+
 // Where a click on a section name scrolls to: its heading, clamped into the
 // range the page can actually reach, so the last section lands at the end of
 // the page rather than asking for a gap under it. -1 for a name the page does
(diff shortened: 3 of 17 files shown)


huacnlee (changes requested):
The premise is real and I reproduced it. On a bare `Flickable` in this repo's own harness, one `angleDelta` of 120 moves 72 px while the same notch reported as eight deltas of 15 moves 9 px. The fix works: driving the real `App.qml` offscreen, `settingsFlick` moves exactly 120 px per notch, including with the pointer over `SettingsPage` content, and the `Flickable` does not additionally flick because `WheelHandler.blocking` defaults true. Four things to settle before merge.

**1. The reader did not get the fix, and the description says it did.** The description lists "the list, the reader, the settings page and both calendar views". The five actual sites are `listFlick`, `setupFlick` (account setup, not the reader), `settingsFlick`, `CalendarEventComposer` and `CalendarEventDetail`. `components/MessageReader.qml:436` is untouched, and the reader is the most-scrolled surface in a mail client. Scroll the list at 120 px per notch, move the pointer to the reader beside it, and the same gesture gives about 9 px. Two panes visible at once behaving differently is worse than both being uniformly slow. Also still on the old path: `MailboxSidebar.qml:54`, `ShortcutHelp.qml:63`, `ComposeView.qml:1449` and `:1535` and `:1595`, `WeekCalendarView.qml:168` (the actual week grid, as distinct from the two event sheets that were patched), `ContactsPicker.qml:139`, `RecipientSuggestions.qml:50`.

**2. The per-event cap reintroduces the chunking-dependence the PR set out to remove.** `WHEEL_MAX_DEGREES_PER_EVENT = 60` at `account/Model.js:807` makes distance depend on how the turn was chopped up again, at the coarse end. The PR's own test documents it: `wheelDistance(-4800) === -480`. An MX Master in free-spin, the exact hardware cited, delivering ten notches as one event of 1200 gives 480 px; the same ten notches as ten events gives 1200 px. If a bound is wanted it should be per unit time, not per event.

**3 and 4. Two latent bugs in the extracted component**, both confirmed with probes. `components/WheelScroller.qml:29` clamps to `[0, contentHeight - height]`, but a `Flickable` with margins reaches `[-topMargin, contentHeight + bottomMargin - height]`. A 300-tall view with `contentHeight` 5000 and 50 px margins, resting at the top with `contentY == -50`, responds to one wheel-**up** notch by moving to 0. It scrolled down in response to a scroll up, and the top margin is now unreachable. The same clamp ignores `originY`: `required property Flickable view` accepts a `ListView`, and one with a 200-tall header reports `originY == -200`, so one notch down jumps 200 px instead of 120 and the header can never be wheeled back into view. Neither bites the five current call sites, but reuse is the whole reason to extract a component.

**5. One new test passes for the wrong reason.** `WheelHandler.orientation` defaults to `Qt.Vertical` and Qt filters horizontal-only wheel events before the signal fires. A bare handler with a counter saw zero deliveries for `mouseWheel(item, x, y, -120, 0)`. So the `event.angleDelta.y === 0` guard at `components/WheelScroller.qml:28` is dead code, and `test_a_sideways_wheel_is_not_a_scroll` passes with the component removed. Same for `test_it_stops_at_the_top`, `test_it_stops_at_the_bottom` and `test_content_shorter_than_the_view_does_not_move`: they read `contentY` synchronously after `mouseWheel`, before a bare `Flickable` has advanced a frame, so they assert nothing beyond what `tests/test_model.js` already covers. The PR is honest that four of eight fail without the component; these are the other four.

Two questions rather than findings. `pixelDelta` and `event.phase` are ignored entirely, so a touchpad would be driven by a derived `angleDelta` at 8 px per degree rather than tracking fingers, and would lose momentum since `cancelFlick()` runs on every event. I could not measure this, and it cuts both ways: `acceptedDevices` defaults to `PointerDevice.Mouse`, so where Qt classifies a distinct touchpad the handler is bypassed and native behavior survives, which also makes the touchpad rationale given for the cap inapplicable there. Worth measuring on real hardware. Separately, wheel scrolling is now an instant jump rather than an animated slide, which the description does not mention and which sits oddly beside `App.qml:1661`, where the repo deliberately chose a 160 ms animation for rail clicks "so the eye can follow where the page went".

Cleanup, non-blocking: this is the third copy of the same clamp, alongside `Model.contentYToReveal` and `ShortcutHelp.scrollBy`, and a single margin-aware and `originY`-aware helper would serve all three. `wheelDistance` is the identity function below the cap, since `(d / 8) * 8` cancels exactly, so the three tests asserting `wheelDistance(-120) === -120` are tautologies. `cancelFlick()` is near-dead given `blocking` is true. The four-line rationale comment is duplicated verbatim at all five call sites when the component already carries it. And no test covers any real call site, though `tst_settings_sidebar.qml` already has a fixture that finds `settingsFlick`, which is what I used to confirm the fix reaches the real hierarchy.

Suite is green: `make test`, `make qml-check`, `omarchy plugin validate .` and `git diff --check` all pass, 192 QML assertions, no qmllint warning mentions `WheelScroller`. The revert experiment matches the claim exactly, 4 failures.

patrykbart:
Thank you — the two latent ones were real and I could reproduce both from your description. All five addressed, and the description is corrected.

## 1. Every vertical scroller now

You were right that this was the worst of them: the reader beside a list moving eight times further for the same turn is worse than both being slow. Added to the reader, the sidebar, the shortcut sheet, all three of compose's, the week grid, the contacts picker, the recipient suggestions — and #83's `LabelPicker`, which landed while this was open and needed it too.

`MailboxTabs` is the one I left out on purpose: `flickableDirection: Flickable.HorizontalFlick`, so a vertical handler there is the wrong axis.

The description said "the reader" and was wrong; it now lists what is actually touched.

## 2. Cap removed

Agreed, and the argument is the PR's own: a per-event ceiling reintroduces exactly the dependence it set out to remove, at the other end. Ten notches as one event gave 480 px and as ten events gave 1200. There is no bound now. A per-unit-time bound would be the defensible version and I would rather not add one speculatively.

## 3 and 4. One range, margin- and `originY`-aware

Both confirmed. `contentYBounds(originY, contentHeight, height, topMargin, bottomMargin)` is the single answer, and `contentYToReveal` and `ShortcutHelp.scrollBy` — your third copy — now ask it as well.

I measured the `originY` half rather than inferring it: a `ListView` with a 200-tall header settles at exactly `originY` and `originY + contentHeight - height`. The margin half is from Qt's documented semantics, because `StopAtBounds` does not correct a programmatic assignment, so the same probe cannot reach it. Worth saying so you know which is which.

## 5. Those four tests

Confirmed, including the reason: `orientation` defaults to `Qt.Vertical` and Qt filters a horizontal-only wheel before the signal fires, so the guard was unreachable and its test passed with the component gone. The guard is out and `orientation: Qt.Vertical` says it out loud.

The other three read `contentY` synchronously before a bare Flickable had advanced a frame, so they asserted nothing. Rewritten so each is paired against a control that has to move differently, and mutation-checked two ways:

- component neutered: **7 of 11 fail**
- component intact, `contentYBounds` reverted to the old 0-floor clamp: **3 fail** (the margin and header cases), plus one node assertion

Every test is caught by one mutation or the other. And `tst_settings_sidebar` now covers a real call site — `settingsFlick` moves exactly 120 px per notch in the real hierarchy — using the fixture you pointed at.

## Cleanup

Done: one clamp instead of three; the four-line rationale removed from all five call sites since the component carries it; `wheelDistance` rewritten per notch rather than per degree, which makes the identity visible instead of hidden — the constant a reader changes is "a notch moves 120 pixels", and the tests now pin that rather than restating the arithmetic.

`cancelFlick` I kept, with a comment saying why: `blocking` stops a flick from *this* event, but one already in flight from a drag goes on decelerating and fights every turn after it. Happy to drop it if you have measured otherwise.

## Your two questions

**Touchpad.** You are right that `acceptedDevices` bypasses the handler for a distinct touchpad, and that this removes the cap's stated justification — which is part of why the cap is gone. I have made it explicit rather than relying on the default: `acceptedDevices: PointerDevice.Mouse`, with a comment that a touchpad reports `pixelDelta` and its own momentum and a Flickable tracks fingers with it properly. Where Qt cannot tell one apart it will be driven by `angleDelta`, which I cannot fix without hardware to measure on. If you can measure it and it is wrong, `pixelDelta` when present is a small change on top.

**The instant jump.** Deliberate, and I should have said so. A wheel is a hand movement and wants to track it; a jump-to-section is a discontinuity and wants the 160 ms so the eye can follow. Animating the wheel also fights a repeating one — each event would restart an animation the next interrupts. It is in the description now.

`make validate` green: 240 QML assertions, no qmllint warning about the component.

PR #90, opened 2026-09-03
Find Exchange's Junk folder by the name Exchange gives it

patrykbart:
Exchange Online's IMAP advertises no SPECIAL-USE, so every special folder there is found by name or not at all. Four of the five already were: `Sent Items`, `Deleted Items`, `Drafts` and `Archive` all match a pattern written for other servers that happens to cover Exchange too.

`Junk Email` did not. The pattern asked for the bare word, so the Junk mailbox resolved to `Junk` — a folder Exchange does not have — and selecting it fails on the one row in the rail whose whole job is to be somewhere else.

Still only a gap-filler behind the flags, and still narrow enough that a folder which merely starts with the word is left alone: somebody's own `Junk Drawer` is not the junk folder.

## Verification

`make validate` green. `tests/test_imap.js` runs Exchange's own LIST reply — no SPECIAL-USE flags, `Sent Items` / `Deleted Items` / `Drafts` / `Junk Email` / `Archive` — through `parseList`, `specialFolders` and `resolveFolder`, and covers `Junk E-mail`, `Junk E-Mail` and `Spam` as well as refusing `Junk Drawer`. The Exchange case fails without the change.

Confirmed against the live server: `outlook.office365.com:993` answers `CAPABILITY IMAP4 IMAP4rev1 AUTH=PLAIN AUTH=XOAUTH2 SASL-IR UIDPLUS MOVE ID UNSELECT CHILDREN IDLE NAMESPACE LITERAL+` — no SPECIAL-USE and no LIST-EXTENDED.

## Release Notes

- The Junk mailbox opens on Exchange and Microsoft 365 accounts, which call the folder `Junk Email` rather than `Junk`.

### DIFF
--- a/providers/ImapProtocol.js
+++ b/providers/ImapProtocol.js
@@ -1055,7 +1055,10 @@ function specialFolders(folders) {
     if (!map["\\sent"] && /^sent( mail| items| messages)?$/.test(leaf)) map["\\sent"] = name
     if (!map["\\trash"] && /^(trash|deleted( items| messages)?)$/.test(leaf)) map["\\trash"] = name
     if (!map["\\drafts"] && /^drafts?$/.test(leaf)) map["\\drafts"] = name
-    if (!map["\\junk"] && /^(junk|spam|bulk mail)$/.test(leaf)) map["\\junk"] = name
+    // Exchange calls it "Junk Email", which is the one name in this list that
+    // the bare word cannot stand in for: every other fallback below happens to
+    // be what Exchange calls the folder anyway.
+    if (!map["\\junk"] && /^(junk([ -]?e-?mail)?|spam|bulk mail)$/.test(leaf)) map["\\junk"] = name
     if (!map["\\archive"] && /^(archive|all mail)$/.test(leaf)) map["\\archive"] = name
   }
   return map
--- a/tests/test_imap.js
+++ b/tests/test_imap.js
@@ -504,6 +504,37 @@ assert.strictEqual(plainSpecial["\\sent"], "Sent")
 assert.strictEqual(plainSpecial["\\junk"], "Junk")
 assert.strictEqual(plainSpecial["\\archive"], "Archive")
 
+// Exchange Online's IMAP advertises no SPECIAL-USE either, so its folders are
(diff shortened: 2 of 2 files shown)


huacnlee (approved):
Looks right, and I verified it rather than only reading it.

The layer is the correct one. Exchange Online advertises no SPECIAL-USE, so pass one of `specialFolders()` is empty and the name fallback is the only thing that can classify anything. The flag pass still runs to completion first and each name assignment stays gap-guarded, so a server that advertises both cannot be downgraded — checked with the matching folder both before and after the flagged one in the LIST. The matcher also runs against the raw wire name, which is what the following SELECT needs, not the decoded display name.

Anchoring holds. I ran the matcher over 30 folder names: `Junk Email`, `Junk E-mail`, `Junk E-Mail` and the German `Junk-E-Mail` all match, while `Junk Email Archive`, `Junk Drawer`, `Junk Emails` and `Junk_Email` are correctly left alone. `INBOX/Junk Email` matches and `INBOX/Junk Email/old` does not, which is the same leaf rule `[Gmail]/Spam` already depends on.

One thing the PR undersells: this also repairs **mark as spam** on Exchange. `flagPlanForLabels` reads `map["\\junk"]`, which was undefined there, so the move target came out as `""` and the operation was a silent no-op that additionally clobbered the archive target set on the line above. With the fix, `flagPlanForLabels(["SPAM"], ["INBOX"], map)` yields `move: "Junk Email"`. Worth a second release note.

Non-blocking:

- `Junk Mail` (no `e`) is still unmatched, and some GroupWise and older Courier setups use it. `junk([ -]?e-?)?mail` would cover it with no realistic false positive.
- The comment says "every other fallback below", but three of the four sibling matchers are above this line; only `\archive` is below.
- The new test block declares `exchangeList`, `exchange` and `junkNamed` at file scope, while the `isSpecialFolder` test just below wraps its locals in a block. Wrapping it the same way keeps `junkNamed` from leaking into the rest of the file.

`make test-js`, `make test-shell` and `make qml-check` all pass. Reverting the production hunk fails the new test as it should, on `Junk` versus the expected `Junk Email`.

PR #91, opened 2026-09-03
Keep an attachment, not only open it once

patrykbart:
Opening was the only thing an attachment row did: the file went to a private runtime directory, `xdg-open` got it, and it was gone with the session. There was no way to keep one, which the README said plainly under "What it does not do" and which is most of what an attachment is for.

The filename still opens — that is what a filename does everywhere else — and the arrow beside it saves. Two verbs, two targets: one click meaning both, or guessing which was wanted, is the thing that would go wrong.

## Where it lands

The desktop's own answer, read from `XDG_DOWNLOAD_DIR` and then from `user-dirs.dirs` rather than by running `xdg-user-dir`, whose package is not everywhere. The folder is named rather than tested for, and created if it is missing: falling back to the home directory when `~/Downloads` did not exist put saved mail loose among the dotfiles, which a test caught.

No file chooser. `omarchy-file-select` does OpenFile and not SaveFile, so a picker would mean either a zenity dependency or a portal client of its own — and a saved file that names its folder in the notice is what a browser does. A "Save as" is a follow-up rather than a thing missing here.

## Nothing is overwritten

Two messages carrying `invoice.pdf` are two invoices, and the second replacing the first is the kind of loss nobody notices until they need the file. So it is numbered the way a browser numbers it, with the count on the stem so the extension stays last and the file still opens in what opens a PDF. `O_EXCL` rather than a plain write, because asking whether a name is free and then writing to it are two moments.

The name is a stranger's, so it is a filename and not a path: both separators go, control characters become underscores, and a name that is only dots is not a name. Saved `0600`, like everything else this plugin writes.

## Verification

`make validate` green. `tests/test_attachment_save.sh` covers the bytes surviving intact, the mode, the numbering and that the original survives it, the extension staying last, `../../etc/passwd` and a Windows path both becoming filenames, a newline in a name, a folder that does not exist yet, `user-dirs.dirs` naming the folder, nothing configured at all, and that a refused save leaves no half-written file. `tests/qml/tst_attachment_save.qml` asserts the row offers both verbs, that saving does not also open, and that the button fits inside the row.

That last one is there because it did not. The row measured itself from its text, so a 24px control beside 15px of caption hung four pixels out of both ends and the reader's scroller clipped it — leaving the filename as the only thing that could be clicked. The row reports a height that holds it now, and the test asserts both axes rather than only the one it was right about.

## Release Notes

- Attachments can be saved, not only opened. The filename opens one in whatever handles its type; the arrow beside it keeps the file in your download folder and the notice says where it went.
- Two messages carrying one filename are two files: the second is numbered rather than landing on top of the first.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,6 +107,7 @@ test-shell:
 	bash tests/test_unsubscribe_transport.sh
 	bash tests/test_image_fetch.sh
 	bash tests/test_attachment_open.sh
+	bash tests/test_attachment_save.sh
 	bash tests/test_attachment.sh
 	bash tests/test_calendar_transport.sh
 	bash tests/test_calendar_write.sh
--- a/README.md
+++ b/README.md
@@ -56,6 +56,10 @@ other IMAP server — including one you run yourself.
   offers an address gets a message; one that only offers a page says so before
   it opens your browser. Nothing is ever fetched from a sender's address until
   you ask.
+- **Attachments open or keep.** The filename opens one in whatever handles its
+  type; the arrow beside it saves the file to your download folder and says
+  where it went. Two messages carrying one name are two files: the second is
+  numbered rather than landing on top of the first.
 - **Images stay blocked.** Loading a sender's pictures tells them the mail was
   read, from which address and when. They load when you ask, for that one
   message.
@@ -256,7 +260,6 @@ A signature is set per mailbox on the settings page, under Writing. It is placed
   written in. A browser engine cannot be embedded in a plugin at all:
   `QtWebEngineQuick::initialize()` has to run before the host process builds
   its `QGuiApplication`, and a plugin loads long after that.
-- **No attachment downloads.** Not yet.
 
 Remote images in a message body are blocked until you ask for them, and asking
 covers that one message. Qt really does fetch an `<img src="https://…">`, so
--- a/Service.qml
+++ b/Service.qml
@@ -738,6 +738,7 @@ Item {
   readonly property bool canOpenOnWeb: !current || current.canOpenOnWeb
   readonly property bool canOpenWebInbox: !!current && current.canOpenWebInbox
   readonly property var unavailableActions: current ? current.unavailableActions : []
+  readonly property var savingAttachmentIds: current ? current.savingAttachmentIds : ({})
   readonly property bool canSend: !current || current.canSend
   readonly property string mailboxKey: current ? current.mailboxKey : "inbox"
   readonly property string searchQuery: current ? current.searchQuery : ""
--- a/Service.qml
+++ b/Service.qml
@@ -844,6 +844,10 @@ Item {
   function openAttachment(messageId, attachment) {
     if (current) current.openAttachment(messageId, attachment)
   }
+
+  function saveAttachment(messageId, attachment) {
+    if (current) current.saveAttachment(messageId, attachment)
+  }
   function preferredSendAs(recipients) {
     return current ? current.preferredSendAs(recipients) : null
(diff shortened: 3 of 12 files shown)


huacnlee:
I pushed three commits to this branch for the items above. The security core I did not touch — it was already right.

**Shorten a long attachment name by its stem, not its extension.** `safe_filename` trimmed the tail until the name fit 240 bytes, which took the extension with it. A 250-character name ending in `.pdf` landed as 240 bytes of stem and opened in nothing. The stem gives up the characters now and the suffix goes back on the end, measured in bytes so an accented name cuts in the same place a plain one does.

This one came with a refactor, so flag it if you would rather it did not. `decode` and `safe_filename` were byte-identical copies in `open-attachment.py` and `save-attachment.py`, and fixing the bug in one of them is precisely the drift that duplicating a security function sets up. Both now import `scripts/attachment_common.py`. The bug was harmless in `open-attachment.py`, where the file is a throwaway in the runtime directory, and permanent here.

**Say what the attachment was actually saved as.** The notice kept only the folder, and the comment justified that with "the name is already on the row that was clicked" — false in the one case this feature exists for. Downloads holding an `invoice.pdf` means the new one is written as `invoice (2).pdf`, the row still says `invoice.pdf`, and the reader opens last month's. The name now comes before the folder, because the notice elides from the right and the folder is the half that is always the same.

**Refuse a second save of an attachment already being saved.** A double-click started two fetches, and since the script will not overwrite, the second was numbered rather than refused. Two identical files, the same notice both times, nothing to say it happened. `MailAccount` now keeps the in-flight attachment ids and the button turns rather than dims, using the `busy` property `IconButton` already carries.

That guard needed one supporting change. `QProcess` emits no `finished()` when the program could not be started, so `onExited` never arrived and the caller waited forever — survivable while nothing was held open, but with an in-flight guard it would have left the button turning for as long as the window stayed open. The save `Process` now reports a failure to start too, deferred a turn so a real exit always reports first.

Verification: `make validate` passes, 191 QML assertions. Each of the three commits was checked out on its own and run through `make test`; all green, so the history bisects. New coverage is four shell assertions for the long-name cases, both plain and multi-byte, and two QML cases for the refusal and the busy state. I reverted each production hunk to confirm its test fails without it — the truncation revert fails both extension assertions, and the row revert fails both new QML cases. qmllint goes up by 2, both `Unqualified access` on `root.service` in the reader's delegate, which is the same category as the 62 already in that file.

Left for you, from the review above and not fixed here:

- `tests/qml/tst_attachment_save.qml` still duplicates `tst_attachment_row.qml` — same fixture rebuilt, the recursive lookup helper reimplemented as `find` beside the existing `named`, and one case restated. The genuinely new cases belong in the existing file.
- `test_the_button_does_not_overlap_the_size` never references the size label; it only asserts the button is horizontally inside the row.
- A dangling symlink at a target name permanently blocks it: `exists()` is false for one, so `unique_path` returns that path and `O_EXCL` then fails with a raw errno and no retry. Falling through to the numbering loop on `FileExistsError` fixes it. The security behaviour is right, only the recovery is not.
- `safe_filename` strips C0 and DEL but not U+202E, so a right-to-left override survives into the saved name. Low severity — the file is written `0600` and non-executable — and now a one-line change in one place rather than two.
- `attachmentSaveComponent` and `attachmentOpenComponent` differ only in whether stdout is collected.
- RFC 2231 `filename*` parameters are still unhandled in `message/Message.js`, so such an attachment saves as the generic `attachment`. Pre-existing and upstream of this PR.

PR #92, opened 2026-09-03
Let the bar icon be turned off without turning the mailbox off

patrykbart:
There was no way to keep Omamail and not have an envelope in the bar.

The Omarchy-level answer — deleting the widget from the bar layout in `shell.json` — hides it and costs more than it looks. `BarWidget.pushSettings` is the only thing that hands plugin settings to the service, as `AGENTS.md` says under Entry points, so a mailbox without a widget in the bar runs on the manifest's defaults: the refresh interval, the notification setting and every other stored answer quietly replaced.

So the widget stays and stops drawing. Off, it gives up its width as well as its visibility, because a zero-width widget lets the bar close over the gap rather than leaving a hole where the envelope was. Settings still cross, which is the whole reason this is a setting rather than a line removed from `shell.json`.

Shown unless something stored says otherwise, rather than on a strict `true`, so a settings file written before this existed keeps the icon it already had.

## The trap it would otherwise be

Turning the icon off with no keybinding leaves nothing to open the window with. The switch says what turning it off does not do — mail is still checked and still notifies — and prints the binding, because a setting that can strand the application should say so where it is turned off.

## Test harness

Testing the widget at all needed the bar types the shell owns. Without a `BarWidget` stub the root type of `BarWidget.qml` resolves to the plugin's own file of the same name and QML reports it as instantiated recursively, so `tests/qml/imports/qs/Ui` gains the three it uses.

The settings sidebar's assertions moved off section positions and onto section keys, because adding a section renumbered assertions that had nothing to do with it.

## Verification

`make validate` green. `tests/qml/tst_bar_widget_icon.qml` asserts the icon is drawn by default, that turning it off takes the width with it, that a hidden widget still hands settings to the service, and that no service yet still draws — so the widget does not flicker out of the bar while the service starts. `tests/qml/tst_service_settings.qml` covers the setting defaulting on, persisting, and that only a stored `false` hides it.

## Release Notes

- The bar icon can be turned off in Settings → Bar, without turning the mailbox off. Mail is still checked and still notifies; only the envelope goes.
- Bind a key to the window before turning it off, or there is no way left to open it. The setting names the binding.

### DIFF
--- a/BarWidget.qml
+++ b/BarWidget.qml
@@ -72,11 +72,23 @@ BarWidget {
     else if (typeof bar.shell.toggle === "function") bar.shell.toggle("omamail", payload)
   }
 
-  implicitWidth: button.implicitWidth
-  implicitHeight: button.implicitHeight
+  // Whether this draws anything. The widget itself stays: it is the only thing
+  // that hands plugin settings to the service, so a widget removed from the
+  // bar layout instead of hidden here would leave the service running on
+  // defaults — the refresh interval, the notification and every other stored
+  // answer quietly replaced by the one in the manifest.
+  //
+  // Off, it is not merely invisible but takes no width either, so the bar
+  // closes over the gap rather than leaving a hole where the envelope was.
+  readonly property bool drawsIcon: !gmail || gmail.showBarIcon
+
+  visible: drawsIcon
+  implicitWidth: drawsIcon ? button.implicitWidth : 0
+  implicitHeight: drawsIcon ? button.implicitHeight : 0
 
   BarIconButton {
     id: button
+    visible: root.drawsIcon
     anchors.fill: parent
     bar: root.bar
     tooltipText: root.gmail ? root.gmail.barTooltip : "Omamail"
--- a/Service.qml
+++ b/Service.qml
@@ -73,9 +73,13 @@ Item {
 
   // Whether the bar draws an envelope for this.
   //
-  // Read as "anything but a stored false" rather than as `=== true`, so a
-  // settings file written before this existed keeps the icon it already had
-  // instead of losing it to a field nobody chose.
+  // A settings file written before this existed keeps its icon because
+  // `applySettings` lays every default down first, so a missing key is
+  // already the manifest's `true` — the same way `unifiedCalendarView` gets
+  // its `false`. What "anything but a stored false" buys instead is the
+  // hand-edited `shell.json`: a `"false"` or a `0` in there is somebody's
+  // typo rather than an answer given in the interface, and a typo should not
+  // be what takes the icon away.
   readonly property bool showBarIcon: !settings || settings.showBarIcon !== false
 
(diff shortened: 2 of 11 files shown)


huacnlee:
Reviewed, and I pushed one commit to the branch rather than sending you back for it: `9c4161f`.

Both claims that matter check out against the running shell, so the design is right. `Bar.qml`'s `ModuleSlot` keys `registryLoader.active` on registry membership rather than visibility, so a hidden widget goes on calling `pushSettings` — and the slot's `implicitWidth` is already `activeItem && activeItem.visible ? … : 0`, so `visible: false` alone gives up the width. The extra zero-width belt is harmless.

What I changed:

- **`test_no_service_yet_still_draws` never reached the case it names.** `fakeShell.serviceFor` answers with a service on every call, so `drawsIcon` was true because that service said `showBarIcon: true`, not because there was no service. Replacing `!gmail || gmail.showBarIcon` with `!!gmail && gmail.showBarIcon` — deleting the branch the test is about — left it passing. It now swaps in a shell holding no service and asserts the widget draws *and* keeps its width; the same mutation now fails it.
- **The `qs.Ui` `BarWidget` stub declared a `barForeground`.** The real base has none — which is exactly what the comment at the top of `BarWidget.qml` says, and why the plugin reads it off the bar. A stub that supplies the property would let that regression through.
- **The comment on `showBarIcon` explained `!== false` as a migration, and it is not one.** `applySettings` lays `defaultSettingValues` down before the stored answers, so a settings file with no `showBarIcon` already reads as `true`, the same way a missing `unifiedCalendarView` reads as `false`. What the reading actually buys is tolerance of a hand-edited `shell.json`, and the comment says that now. Behaviour unchanged.
- **The warning contradicted itself**, telling the reader there is no way left to open the window and then printing a command that opens it. Both copies now say the window opens from a keybinding or a terminal and nowhere else. Worth getting right: I checked, and nothing in the installed shell renders a plugin settings form — `schema` is stored and read by no QML in `/usr/share/omarchy/shell` — so that manifest entry is not a recovery path for anyone today. Omamail's own settings page is the only UI, which makes this warning load-bearing.

`make validate` green on the merged result (215 QML assertions with current `main`), and no textual or semantic conflict with what has landed since.

Two things left for you, neither blocking. `objectName: "showBarIconSwitch"` is added and used by nothing, and the settings-page-to-service wiring has no test — the stub `ToggleSwitch` has no `MouseArea`, so it would be a direct `toggled()` call plus `showBarIcon`/`setShowBarIcon` on the fake service in `tst_settings_sidebar.qml`. And `checked: !root.service || root.service.showBarIcon !== false` is the only row on that page not written as `!!root.service && root.service.X`; on a `bool` the `!== false` is a no-op, and the only real difference is that this switch reads on with no service where every other reads off. Deliberate is fine, but it reads as drift.

PR #93, opened 2026-09-03
Add an All Mailboxes view

patrykbart:
## Summary

Add an **All mailboxes** view that combines the inbox, spam, sent, and trash mailboxes from every configured account into one readable list.

## Behaviour

- Rows are sorted across mailboxes and carry their owning mailbox identity.
- Selecting or opening a row routes actions to the mailbox that owns it.
- Replies and new drafts keep the correct sending mailbox.
- Search runs across all mailboxes and keeps its loading state until every source settles.
- Capabilities are limited to actions supported by every mailbox in the combined view.
- Conversation counts are shown only on rows whose own source provides conversation members.
- Paging stops at the shallowest source page so the merged list does not contain gaps.
- Selecting an individual account leaves the existing single-mailbox behaviour unchanged.

## Validation

- Added unit coverage for merge ordering, composed IDs, paging, capability intersection, search state, and row-level conversation handling.
- Added QML coverage for account switching, routing, compose identity, attachments, and unified service behaviour.
- `node tests/test_unified.js`
- `bash tests/test_source.sh`

The full `make validate` suite is currently blocked on macOS because `tests/test_config_store.sh` uses the GNU `stat -c` option.

### DIFF
--- a/Service.qml
+++ b/Service.qml
@@ -236,14 +236,38 @@ Item {
     // a half-added mailbox could never be the one on screen, and setup would
     // have nothing to run in.
     if (!next && accountHosts.count > 0) next = accountHosts.objectAt(0)
-    if (next === current) return
+    if (next !== current) current = next
+    applyLiveHosts()
+  }
+
+  // Which mailboxes are live, which is what earns a list.
+  //
+  // `active` gates every list load there is: `refresh` will not reload an
+  // already-loaded list without it, `onActiveChanged` is what asks for the
+  // first one, and the poll timer checks it before extending one. Leaving it
+  // on the visible account meant a merged list refreshed one mailbox's rows
+  // and left the others as they were — counts moved, mail did not. A merged
+  // view has every mailbox on screen at once, so every mailbox gets one.
+  //
+  // `windowOpen` travels with it for the same reason: `refresh` reloads a
+  // loaded list only for a mailbox the window is showing.
+  function applyLiveHosts() {
     for (var i = 0; i < accountHosts.count; i++) {
       var host = accountHosts.objectAt(i)
-      if (host) host.active = host === next
+      if (!host) continue
+      // A mailbox part-way through being added has no id and no rows to
+      // contribute, so it is live only when it is the one on screen — which
+      // is what setup runs in.
+      var live = host === current
+        || (unified && String(host.accountId || "") !== "")
+      host.active = live
+      if (live) host.windowOpen = windowOpen
     }
-    current = next
-    if (current) current.windowOpen = windowOpen
   }
+
+  // Turning the merged view on or off changes which mailboxes are live, and
+  // `refreshCurrent` does not run for it: the account on screen has not moved.
+  onUnifiedChanged: applyLiveHosts()
 
   // The whole point of switching is that it is instant, which it is because
   // each account keeps its own cache on disk. A queued send belongs to its
@@ -788,10 +812,32 @@ Item {
 
   // Which services are involved, which is what decides the rail and every
   // button on it. Two Gmail mailboxes ask one provider's questions.
-  readonly property var unifiedProviders: {
+  // What every mailbox in the merge can actually do, and which rail rows it
+  // has. Read off the hosts rather than their provider ids: each one has
+  // already narrowed its provider by the refusals its server reported and the
+  // mailboxes that turned out not to be there, and those arrive while the app
+  // is running. Reading them here is what makes the intersection follow them.
+  readonly property var unifiedAbilities: {
+    var _epoch = listEpoch
+    var out = []
     var accounts = accountList ? accountList.accounts : []
-    return Unified.providersOf(accounts)
+    for (var i = 0; i < accounts.length; i++) {
+      if (!accounts[i].id) continue
+      var host = accountHosts.objectAt(i)
+      out.push({
+        archive: !!(host && host.canArchive),
+        spam: !!(host && host.canReportSpam),
+        star: !!(host && host.canStar),
+        labels: !!(host && host.hasLabels),
+        web: !!(host && host.canOpenOnWeb),
+        move: !!(host && host.canMove),
+        conversations: !!(host && host.showsConversations),
+        mailboxes: host ? host.mailboxes : []
+      })
+    }
+    return out
   }
+
 
   readonly property var unifiedMessages: Unified.mergeMessages(unifiedSources)
 
@@ -858,7 +904,7 @@ Item {
 
   property bool windowOpen: false
   onWindowOpenChanged: {
-    if (current) current.windowOpen = windowOpen
+    applyLiveHosts()
     if (windowOpen) restoreWindow = false
     saveWindowPrefs()
   }
@@ -911,7 +957,7 @@ Item {
   // view cannot offer it: the picker would open on an empty list, and a chosen
   // id would belong to whichever mailbox happened to be active.
   readonly property bool canMoveToLabel: !unified
-    && Provider.can(providerId, "move")
+    && !!current && current.canMove
 
   // Which service the mailbox on screen is, what mailboxes it has, and what it
   // can be asked to do. Forwarded like everything else so a view never has to
@@ -920,28 +966,28 @@ Item {
   // `providerId` stays the *visible* mailbox's even in a merged view: it
   // decides what a query string means and what a setup page offers, both of
   // which belong to one account. What a merged list may *do* comes from
-  // `Unified.sharedCapability` rather than from here.
+  // `Unified.everyMailboxCan` rather than from here.
   readonly property string providerId: current ? current.providerId : Provider.DEFAULT_ID
   readonly property var mailboxes: unified
-    ? Unified.sharedMailboxes(unifiedProviders)
+    ? Unified.sharedMailboxRows(unifiedAbilities)
     : (current ? current.mailboxes : Provider.mailboxes(Provider.DEFAULT_ID))
   readonly property bool canArchive: unified
-    ? Unified.sharedCapability(unifiedProviders, "archive")
+    ? Unified.everyMailboxCan(unifiedAbilities, "archive")
     : (!current || current.canArchive)
   readonly property bool canReportSpam: unified
-    ? Unified.sharedCapability(unifiedProviders, "spam")
+    ? Unified.everyMailboxCan(unifiedAbilities, "spam")
     : (!current || current.canReportSpam)
   readonly property bool canStar: unified
-    ? Unified.sharedCapability(unifiedProviders, "star")
+    ? Unified.everyMailboxCan(unifiedAbilities, "star")
     : (!current || current.canStar)
   readonly property bool hasLabels: unified
-    ? Unified.sharedCapability(unifiedProviders, "labels")
+    ? Unified.everyMailboxCan(unifiedAbilities, "labels")
     : (!current || current.hasLabels)
   // Intersected like every other capability: a merged list holds rows from
   // mailboxes whose provider has no web UI at all, and "Open in browser" on
   // one of those is a button that cannot be honoured.
   readonly property bool canOpenOnWeb: unified
-    ? Unified.sharedCapability(unifiedProviders, "web")
+    ? Unified.everyMailboxCan(unifiedAbilities, "web")
     : (!current || current.canOpenOnWeb)
   readonly property bool canOpenWebInbox: !unified && !!current && current.canOpenWebInbox
   // A key is not a button: `e` and `s` are bound whatever mailbox is open, so
@@ -986,14 +1032,25 @@ Item {
   // account's threads beside a mailbox that has none would draw a member count
   // on the rows that happened to carry one and nothing on the rest.
   readonly property bool showsConversations: unified
-    ? Unified.sharedCapability(unifiedProviders, "conversations")
+    ? Unified.everyMailboxCan(unifiedAbilities, "conversations")
     : (!!current && current.showsConversations)
   // The rail and its members are about the message being read, so they come
   // from the mailbox holding the selection — the same place `selectedMessage`
   // and `detailPainted` come from.
   readonly property bool showsRail: !!reading && reading.showsRail
-  readonly property var selectedThread: reading ? reading.selectedThread : null
-  readonly property var memberSummaries: reading ? reading.memberSummaries : ({})
+  // Composed on the way out, like `selectedId` and for the same reason: the
+  // rail hands a member id back to `select` and to `act`, and compares the
+  // open one against `selectedId`. A raw one reached no mailbox at all.
+  readonly property var selectedThread: {
+    if (!reading) return null
+    if (!unified) return reading.selectedThread
+    return Unified.composeThread(reading.accountId, reading.selectedThread)
+  }
+  readonly property var memberSummaries: {
+    if (!reading) return ({})
+    if (!unified) return reading.memberSummaries
+    return Unified.composeMembers(reading.accountId, reading.memberSummaries)
+  }
   readonly property string mailboxKey: unified
     ? unifiedMailboxKey : (current ? current.mailboxKey : "inbox")
   // What `mailboxKey` means for a reader that is looking at a thread, which is
@@ -1237,11 +1294,50 @@ Item {
     // account and carries the id, so compose can already name one; this is the
     // routing it was missing.
     var values = fields || ({})
-    var target = String(values.accountId || "")
-    var host = target !== "" ? findAccount(target) : null
-    if (!host) host = senderFor(String(values.from || ""))
-    if (!host) host = current
-    return host ? host.send(values) : false
+    var host = sendHostFor(values)
+    return host ? host.send(withSourceDraftId(values)) : false
+  }
+
+  // The mailbox a submission is sent from.
+  //
+  // A named one is the answer, and a named one that is not here is a refusal
+  // rather than permission to guess: falling through to matching the address
+  // sent the message from whichever mailbox matched first, which for two that
+  // share a send-as alias is not the one the composer chose. The address is
+  // only consulted when nothing named a mailbox at all.
+  //
+  // Resolved in one place so the choice can be asserted, rather than inferred
+  // from what happened after it.
+  function sendHostFor(fields) {
+    var values = fields || ({})
+    var target = draftOwner(values)
+    if (target !== "") return findAccount(target)
+    var host = senderFor(String(values.from || ""))
+    return host ? host : current
+  }
+
+  // The mailbox a submission belongs to: the one it names, or — for a draft
+  // raised from a merged list — the one its id was composed from.
+  function draftOwner(values) {
+    var named = String(values.accountId || "")
+    if (named !== "") return named
+    return Unified.accountOf(String(values.draftId || ""))
+  }
+
+  // The same submission with its draft id as the owning provider issued it.
+  //
+  // Asked of the id rather than of `unified`, because the composer can be
+  // opened from a merged list and saved after the reader has left it — and a
+  // provider handed a composed id answers that the draft is no longer there
+  // and writes nothing. A bare id cannot hold the separator, so this is safe
+  // to ask of any of them.
+  function withSourceDraftId(values) {
+    var id = String(values.draftId || "")
+    if (Unified.accountOf(id) === "") return values
+    var out = ({})
+    for (var key in values) out[key] = values[key]
+    out.draftId = Unified.sourceIdOf(id)
+    return out
   }
 
   // Which mailbox owns an address, for a From that named no account. Asked of
@@ -1269,13 +1365,13 @@ Item {
   }
   function saveDraft(fields, callback) {
     var values = fields || ({})
-    var target = String(values.accountId || "")
+    var target = draftOwner(values)
     var host = target !== "" ? findAccount(target) : current
     if (!host) {
       if (typeof callback === "function") callback(null, "The mailbox for this draft is unavailable")
       return null
     }
-    return host.saveDraft(values, callback)
+    return host.saveDraft(withSourceDraftId(values), callback)
   }
   function fail(text) { if (current) current.fail(text) }
   function note(text) { if (current) current.note(text) }
--- a/Service.qml
+++ b/Service.qml
@@ -741,7 +741,11 @@ Item {
       out.push({
         id: accounts[i].id,
         label: Accounts.label(accounts[i]),
-        messages: host ? host.messages : []
+        messages: host ? host.messages : [],
+        // Whether this mailbox has more to come, which is what decides where
+        // the merge has to stop: a source still holding mail below its last
+        // row would leave a hole in the middle of the list.
+        hasMore: !!(host && host.hasMore)
       })
     }
     return out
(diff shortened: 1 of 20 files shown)


huacnlee (changes requested):
The unified view is the right shape — answering everything from the `Service` façade instead of building a second panel is where this belongs, and `Unified.js` is well argued and well tested. What is not finished is the audit of *which* façade properties you actually merged. Several were left reading `current`, and the ones that matter are not cosmetic.

I pushed `b654d9c` to the branch with the three that were self-evident. The rest need your decisions.

## What I pushed

**The id contract, which is the serious one.** A merged row is addressed by mailbox and id, but `MessageReader` called back into the service with `selectedMessage.id` — the id the owning account issued, with no mailbox on it. `hostForId` finds nothing, so in a unified view the reader's star button and the attachment rows silently do nothing at all, and `openInBrowser` falls through to whichever account is active and builds its Gmail URL out of another mailbox's id. `MessageMenu` happens to pass the composed id because it reads from `service.messages`, so one function was getting two id shapes from two callers. Everything now goes through `service.selectedId`, which `ComposeView` and `MessageList` already use, and `openInBrowser` is routed through `hostForId`/`sourceIdFor` like every other action on a row. The invariant is worth stating somewhere: nothing outside `Service` ever handles an account's own id.

**Two capabilities left out of the intersection.** `canOpenOnWeb` came from the active account, so an IMAP row in a merged list drew a browser button its provider has no address for — the exact promise `Registry.capabilities` exists to stop. And `unavailableActions` came from `current`, which is the AGENTS.md rule about `e` and `s`: they are bound in every mail context whatever mailbox is open, so with HEY in the merge the status row went on advertising archive and star for messages `MailAccount.act` refuses. Both now answer from the same intersection the buttons use.

## What needs you

**Compose sends from the wrong mailbox.** `ComposeView.begin` takes `accountId` from `service.activeAccountId` and `selectPreferredFrom` goes through `current.preferredSendAs`, so replying to a message that arrived in mailbox B, from a merged list, composes and sends from mailbox A with nothing on screen saying so. That is the failure a unified inbox most has to avoid, and it belongs in the "could not be merged" list — or better, be merged: route `preferredSendAs` through `reading` and seed the draft's account from the row's owner.

**`accountOf` cannot tell a composed id from a bare IMAP one.** `"42:Sent Items"` parses as account `"42:Sent"`. It fails safe today only because no account is named that, which is luck rather than structure. Either a test for that case or a separator a message id provably cannot contain.

**Three smaller ones.** `actionStatus` still reads `current`, so archiving a row from a non-active mailbox produces no confirmation and no undo affordance. `allLoaded` requires every mailbox to have loaded and a signed-out account never does, so the merged list shows no skeleton, no "Nothing here" and no error — a blank panel, indefinitely. And `chooseUnified` returns early while the calendar is visible without calling `selectMailbox`, so the accounts stay on whatever mailboxes they were on while the rail claims one.

**Paging.** Your `loadMore` comment says the merge cannot extend past the oldest row of the shallowest source without leaving a gap, and I agree — but the code loads from every source and then draws all of it, so the gap is on screen rather than avoided. With two accounts on different page sizes the tail of the list is one account's mail with the other's missing from the middle. Either truncate at the shallowest watermark, or say in the description that it is accepted.

**`MessageRow`'s source label has no width ceiling.** `sender.width` subtracts `source.width`, and `source` only has an `implicitWidth`, so its elide never fires; nothing limits the length of a name a user can set. A long one squeezes the sender to zero and pushes the row out of its width.

## Rebase before the next round

This branch is eight commits behind and two of them collide semantically, not just textually.

- **#83 (`v` moves a message to a label or folder)**: the picker is built from `service.labels`, which you hard-code to `[]` in a unified view, so `v` opens empty — and `refuseUnavailableAction` asks `current`, while a chosen label id would come from a different account's label list. `Registry` also gained a `move` capability that `sharedCapability` does not cover.
- **#91 (attachment saving)**: `saveAttachment` and `savingAttachmentIds` need exactly the routing you built for `openAttachment`, and the reader call site has the same `summary.id` bug I just fixed on the others.
- **#88 (`forwardReplyFailure`)** calls `switchToIndex` directly, bypassing the "choosing one mailbox turns the combined view off" rule in `App.switchAccount`.

The textual conflicts are all in `Service.qml` and two of them matter: `main` replaced `sending` with a `sendingHost` scan — which happens to fix a hole here, since `sending: !!current && current.sending` means the send button never goes busy for a non-active mailbox and can be pressed twice — and `selectLabel` now takes `(name, labelId)`, whose second argument your override drops.

## On splitting

Yes, please — not for size, but because the two halves have different review shapes. Naming a mailbox is local, fully tested and mergeable today; it is only waiting on the unified work. Four of the nine commits are rework of the name field, which makes the combined history harder to read than either half would be alone.

## Untested claims

The description says no view knows more than one mailbox is involved; three do (`MessageRow.sourceLabel`, `AccountSwitcher.unifiedActive`, `App` reading `service.unified`), all deliberately and with comments, so that is a wording point. But `searchQuery`, `rawQuery`, `resultSummary`, `serverSearchLoading`, `canSend`, `providerId` and `syncedLabel` also still read `current` with nothing saying why they were left. And the `Service`-level routing — `select`, `hostForId`, `sourceIdFor` on real messages, the `listEpoch` recompute, `chooseUnified`, the optimistic update and rollback in a merged list — has no test at all. Every defect above landed in exactly those gaps.

`make validate` is green on my pushed commit (209 QML assertions), and the new reader assertion fails against the previous code, so it measures the fault.

patrykbart:
Thank you — you were right that the audit was the unfinished part, and `b654d9c` is a better fix than the one I would have written: routing everything through `service.selectedId` makes the invariant structural rather than a rule to remember. I have stated it in the description.

Rebased onto `main` (#89, #95 and #101 are in), and every item is addressed. Description and title rewritten; the naming half is out of this branch as #102.

**Compose.** Merged, not documented. `composeAccountId` seeds the draft from the mailbox the message arrived in, `preferredSendAs` is asked of that mailbox, and `fromIdentities` is filtered against it — so a reply to a message that came to B is written, addressed and sent from B. A new message with no selection is unchanged.

**The separator.** U+001F. Gmail issues hex, HEY issues `<posting>:<topic>`, and IMAP forbids a control character in a mailbox name while `quote` drops the two that would end a command — so a composed id is recognisable and a bare one provably is not. `test_an_id_for_a_mailbox_that_is_not_here_reaches_nothing` asserts that `"42:Sent Items"` reaches no host and that `act` on it returns false, and `test_an_id_names_the_mailbox_that_owns_it` round-trips it.

**`actionStatus`.** The most recent of them wins, which is the one the press just produced. `actionStatusAt` is what orders them.

**`allLoaded`.** Fixed, and worse than you described: no skeleton *and* no "Nothing here", so the placeholder was blank rather than wrong. A mailbox that has settled on an error has finished as far as the list is concerned — it will never set `listLoaded` — and what went wrong is `firstError`'s to say. A mailbox that is retrying is still on its way even while holding the last attempt's error, which is the case the test pins.

**`chooseUnified`.** No longer returns before `selectMailbox`.

**Paging.** Truncated, not accepted. `pageWatermark` is the newest of the sources' own last rows; above it every source has reported, so the list is complete. A source with `hasMore !== true` sets no watermark, since it cannot leave a hole. You were right that the comment described the fix and the code did the opposite.

**`MessageRow`'s source label** has a width ceiling of a third of the row, so a long name cannot squeeze the sender to nothing.

**The rebase.** #83's move is refused in a merged view with that reason, and only that reason — a single mailbox whose provider has no move verb still gets the provider guard's answer, which I had conflated at first. `move` is in `sharedCapability`'s reach through `canMoveToLabel`. #91's `savingAttachmentIds` is joined and re-keyed like every other row address. #88's `forwardReplyFailure` and `undoSend` no longer switch accounts in a merged view: the switch existed so a retry could not be addressed to whatever was visible, and the draft now names its own mailbox, so switching would only drop the reader out of the combined view behind `App.switchAccount`'s back. `main`'s `sendingHost` scan and `selectLabel(name, labelId)` are kept as they are.

**The untested claims.** The description said no view knows more than one mailbox is involved and three do — corrected. The seven properties that stay with the visible mailbox each say why now. And the Service-level routing has tests: `tst_unified_routing.qml` drives a real `Service` with real account hosts and covers the merge, the composed ids, the routing of a select and the clearing of the others, the compose owner, the rail and a search reaching every mailbox, and an id for a mailbox that is not there reaching none. Every fix above was mutation-checked against it.

**On splitting** — done, #102. It touches `AccountSwitcher.qml` for the same two testability lines this branch adds (the `menuRows` alias and `objectName: "account-row"`), so whichever lands second will want a one-hunk resolution.

`make validate` green: 287 assertions.

patrykbart:
One correction to the above, found while integrating this with #94 locally: I cited the wrong test for the separator.

`tst_unified_routing.qml`'s assertion that `"42:Sent Items"` reaches no host passes under a colon separator *and* under the space separator that was the original bug — no test account is named `"42"` or `"42:Sent"`, so it fails safe by luck, which is the thing you objected to in the first place.

The property is measured in `tests/test_unified.js`, against the id shapes the three providers issue: `splitUnifiedId("42:Sent Items")` and `splitUnifiedId("1a0682969660774e")` must both come back empty. Those assertions do kill both separator mutants. I have said so where the weaker assertion lives, and dropped the part of it that could not tell the difference.

Also worth flagging from that integration: #94's `selectionIsPreview` has to be asked of the mailbox holding the selection rather than `current`, and `markPreviewRead` has to route by id — otherwise a previewed row from a non-visible mailbox reads as an open message and the two decisions that property exists for both go wrong. Whichever of the two lands second will want that hunk; it is three lines and I am happy to carry it on either branch, whichever you prefer.

patrykbart:
Rebased onto `main` at 0.8.1, and the conflict I flagged did land: #102 and this branch both add properties to the switcher's account row, so `named`/`primaryText`/`secondaryText` and this branch's `inUse` met in one hunk. Both kept — they answer different questions about the same row, and the `menuRows` alias and `objectName` came in with #102 so there is nothing left to duplicate.

#104 auto-merged but is worth a second pair of eyes since it rewrote the line this branch writes into: `readingStatusLine` supplies the text and #104's inset arithmetic sizes the box around its `implicitWidth`, so "All mailboxes" gets a snug box rather than an elided one. Nothing to reconcile that I can find.

302 assertions green.

huacnlee (changes requested):
Thanks for addressing the earlier review and rebasing. I re-reviewed the current head and ran the existing validation plus additional offscreen regressions. The baseline make validate passes (302 QML tests, Qt 6.11.2), but this is not ready to merge yet.

1. [P1] Attachment saving still crosses account ownership. MessageReader.qml:639 passes summary.id without the unified account qualifier, and Service.qml:1254–1255 dispatches directly to current. With A active and B's message selected, a real-Service regression records the save on A and none on B. Pass the qualified selectedId and resolve hostForId/sourceIdFor as openAttachment already does. A separate reader regression also shows the save busy indicator remains false: Service prefixes saving keys with account ID, while MessageReader.qml:627–628 looks up only attachmentId. Please cover both the reader signal and the real Service dispatch. Matching IMAP UID/part identifiers could return A's attachment under B's filename; that consequence is source-supported, not a live-mailbox exploit reproduction.

2. [P1] A reply owned by B still contains A's signature. ComposeView.begin now correctly sets accountId from composeAccountId, but accountSignature at ComposeView.qml:52–53 still uses service.activeSignature, which Service derives from active account A. An offscreen ComposeView regression confirms A's synthetic private signature appears in B's draft. Resolve the signature using the draft's owning account so another identity's contact information cannot be inserted into the outgoing body.

3. [P2] Reply-all still excludes the wrong identity. ComposeView.otherRecipients at line 350 uses service.accountEmail (A). For a reply owned by B to original recipients [B, A], the regression produces Cc [B], omitting A and copying the sender. Resolve the address from the draft owner and test this separately from preferredSendAs/accountId routing.

Security verdict: BLOCK. The wrong-account attachment dispatch and signature insertion are demonstrated account-isolation boundary violations. No real credentials or live mailboxes were used, and the possible wrong-file download was not claimed as a live exploit. Passing the existing suite does not cover these cases: five focused new probes fail across service routing, attachment reader behavior, and compose identity.

The other reviewed fixes show progress. Please add the account-ownership regressions above and complete the remaining routing before we merge. When this and #94 are combined, selectionIsPreview must derive from reading, and markPreviewRead must route and decode the composite ID. That integration should be verified on the combined tree.

patrykbart:
All three confirmed and fixed. You were right that the audit was still not finished — I had merged the routing at the message level and left three places asking the active account, and all three cross from one identity into another's.

**1. Attachment saving.** Both halves, as you asked.

The reader handed back `summary.id` while every other action on that row had moved to `selectedId`, and `Service.saveAttachment` dispatched to `current` without routing at all. It goes through `hostForId`/`sourceIdFor` like `openAttachment` now, from a reader that answers with the id the service gave it. Your note about matching IMAP part identifiers is the part that made me stop treating this as cosmetic: a part id is a number in a message tree, so two accounts collide routinely rather than exceptionally.

The spinner was the same fault from the other side. The merged map is keyed by mailbox and attachment; the reader looked up the attachment alone, found nothing, and the row stayed idle for the whole save. `attachmentIsSaving(messageId, attachmentId)` composes the key, because composing ids is the service's job — the invariant is that nothing outside `Service` handles an account's own id, and a view holding an attachment and its message should not be the exception to it.

Covered in both places: `tst_unified_routing.qml` drives a real `Service` with real hosts and asserts the save refusal lands on the mailbox that owns the message and nowhere else, and that the spinner belongs to that mailbox's row; `tst_message_attachment.qml` covers the reader signal, beside the assertion that already covered opening.

**2. The signature.** Fixed. `begin` set `accountId` correctly and then `accountSignature` read `service.activeSignature` two lines away — B's reply signed with A's name and contact details.

**3. Reply-all.** Fixed, and your reading of it is exactly right: for a message addressed to both, answered from B, it removed A — a real recipient — and kept B, copying the sender to their own reply.

Both resolve through `signatureFor(accountId)` / `accountEmailFor(accountId)`, kept on the service for the reason `activeSignature` was: the window never reaches into the account list. `activeSignature` is now `signatureFor(activeAccountId)`.

Tested separately from the routing, as you asked. `tst_compose_identity.qml` is new because `tst_signature.qml` has one account and cannot tell owner from active — with one mailbox the two answers coincide, which is why its nineteen assertions all passed against the fault.

Each of the six fixes was mutation-checked on its own: reverting any one of them fails its test and nothing else. 312 assertions green.

**On the combined tree.** Verified rather than assumed — and it turned up something neither suite could see. `selectionIsPreview` is `reading`'s, which is what `detailPainted` and `detailLoading` already answered from; #94's dwell now waits on those, so reading any of them from the visible mailbox would arm a dwell against a body belonging to another account. `markPreviewRead` routes and decodes the composite id. Three regressions in `tst_unified_routing.qml` cover that seam, each mutation-checked, and the combined tree is 342 green. It is on my fork's `dev` branch if you want to look at it; the three tests will land with whichever of the two merges second.

patrykbart:
Rebased onto `main` at c5d87ef. Three conflicts, and #111 raised two questions the merge had to answer rather than just re-apply:

**`showsConversations` is a capability**, `Provider.can(..., "conversations")`, so a merged list intersects it like archive, spam, star and labels. A list holding a JMAP account's threads beside a mailbox that has none would draw a member count on the rows that happened to carry one and nothing on the rest.

**`showsRail`, `selectedThread` and `memberSummaries` are about the message being read**, so they come from `reading` — the same place `selectedMessage` and `detailPainted` come from. `viewedMailboxKey` is asked of the service's own `mailboxKey` rather than of whichever account is underneath it, since it is the same question of a merged list as of one mailbox.

`act` keeps its new `memberOnly` argument through the routing. `MessageRow` now carries sender, source label and thread count in one row: the sender takes what the other two leave, both are optional, and the source label keeps its width ceiling so a long mailbox name cannot squeeze the sender out.

One process note, since it is the kind of thing that bites later: resolving that `MessageRow` conflict I added the width ceiling early, and the commit that introduces it properly then added it again — a duplicate property that made the file uncompilable from that commit onward, while the final tree looked fine. Fixed across the affected commits with a scripted rebase rather than patched at the tip, so every commit builds.

`make validate` green: 400 assertions.

Service.qml
```diff
@@ -893,7 +1226,39 @@ Item {
...
-    return current ? current.send(fields) : false
+    // The mailbox the From address belongs to. `sendIdentities` spans every
+    // account and carries the id, so compose can already name one; this is the
+    // routing it was missing.
+    var values = fields || ({})
+    var target = String(values.accountId || "")
+    var host = target !== "" ? findAccount(target) : null
+    if (!host) host = senderFor(String(values.from || ""))
+    if (!host) host = current
+    return host ? host.send(values) : false
```
huacnlee: **P1: Missing sender identity can route mail through the wrong account**

This first checks `fields.accountId`, but `ComposeView.submit()` does not pass the `accountId` it owns, so the code falls back to the first address match in `senderFor(from)`. When two accounts share a send-as alias, selecting account B can call account A, whichever host appears first; ordinary single-account views are affected by the same regression.

Executing the actual Compose submit and Service routing functions with synthetic accounts produced: `EXPECTED B; ACTUAL A; fields.accountId undefined`. Always carry and validate the selected account ID; do not replace the selected identity with first-address matching.

Service.qml
```diff
@@ -848,39 +1087,133 @@ Item {
...
-  function refresh() { if (current) current.refresh() }
-  function loadMore() { if (current) current.loadMore() }
-  function select(id) { if (current) current.select(id) }
-  function clearSelection() { if (current) current.clearSelection() }
+  function refresh() {
+    if (!unified) {
+      if (current) current.refresh()
+      return
+    }
+    eachHost(function(host) { host.refresh() })
```
huacnlee, line 1191: **P1: Refreshing the unified list still updates only the active account**

Iterating over all hosts and calling `refresh()` is insufficient because `MailAccount.refresh()` still gates list loading on `active`, while `refreshCurrent()` marks only the current account active. In unified mode the other accounts can refresh counts but their existing message lists remain stale.

In an offline QML test with a ready B account and `listLoaded=true`, calling `service.refresh()` left B's `listSerial` unchanged. All participating accounts need list loading and refresh eligibility, including initialization, manual refresh, and later polling.

Service.qml
```diff
@@ -762,74 +892,183 @@ Item {
...
+  // on the rows that happened to carry one and nothing on the rest.
+  readonly property bool showsConversations: unified
+    ? Unified.sharedCapability(unifiedProviders, "conversations")
+    : (!!current && current.showsConversations)
+  // The rail and its members are about the message being read, so they come
+  // from the mailbox holding the selection — the same place `selectedMessage`
+  // and `detailPainted` come from.
+  readonly property bool showsRail: !!reading && reading.showsRail
+  readonly property var selectedThread: reading ? reading.selectedThread : null
+  readonly property var memberSummaries: reading ? reading.memberSummaries : ({})
```
huacnlee: **P2: Conversation member IDs are not encoded consistently**

`selectedId` is now `account ID + separator + message ID`, but the `selectedThread.memberIds` and `memberSummaries` exposed here still contain raw message IDs. `Conversation.memberStep()` and the conversation rail therefore return a bare ID; passing it to `Service.select()` makes `hostForId()` return null, and the open-member comparison also fails.

The QML reproduction logged `selected="jmap:b@example.net\u001f2"`, next member `"2"`, and a null host. Convert conversation member IDs, summary keys, and membership checks at the Service boundary.

huacnlee (commented):
Review conclusion: do not merge yet; five issues require fixes.

Reviewed commit: `ca22a0f173ed4ff79dcf535bda7d0ba4192dcbdb`.

Verification: `make validate` passed in an isolated temporary copy; the existing QML suite reported 400 passed and 0 failed. Additional offline QML tests using real Service/MailAccount objects, plus Node probes executing the Compose/Service/Gmail functions, reproduced the omissions below. Runtime: Qt 6.11.2. No real mailbox was accessed and no real mail was sent.

Security verdict: **BLOCK**. The sender-identity issue proves that a submission selected for account B can be dispatched through account A, so the account-ownership boundary is not preserved. The evidence establishes incorrect dispatch, not external delivery; this is not a complete security audit.

The current GitHub account is the PR author, so GitHub does not allow a Request changes review; this was submitted as a comment review with the requested fixes.

Service.qml
```diff
@@ -762,74 +892,183 @@ Item {
...
+  // the attachment row never learn which of the two they are in.
+  readonly property var reading: unified ? selectionHost : current
+
+  // Composed on the way back out, because the panel compares this against the
+  // ids in the list it was given — and in a unified view those carry the
+  // mailbox. The account only ever knows its own.
+  readonly property string selectedId: {
+    if (!reading) return ""
+    if (!unified) return reading.selectedId
+    return Unified.unifiedId(reading.accountId, reading.selectedId)
```
huacnlee, line 1105: **P2: A composite message ID is passed to the provider as a raw draft ID**

Editing an existing draft from the unified Drafts list sends this composite `selectedId` through `App.resumeHeldDraft()` into `ComposeView.sourceDraftId`; saving and sending then pass it unchanged as `draftId`. Providers do not recognize that composite value.

Executing the actual Service.saveDraft and Gmail update logic with a synthetic draft list showed that raw `message123` resolves to `draft123`, while `b@example.net\u001fmessage123` does not; the callback returned `That draft is no longer in the mailbox` and no PUT was made. Preserve the owning account and original draft message ID separately; conversion must not depend on whether unified mode is still active when the save occurs.

Service.qml
```diff
@@ -762,74 +892,183 @@ Item {
...
-  readonly property bool canOpenOnWeb: !current || current.canOpenOnWeb
-  readonly property bool canOpenWebInbox: !!current && current.canOpenWebInbox
-  readonly property var unavailableActions: current ? current.unavailableActions : []
-  readonly property var savingAttachmentIds: current ? current.savingAttachmentIds : ({})
+  readonly property var mailboxes: unified
+    ? Unified.sharedMailboxes(unifiedProviders)
+    : (current ? current.mailboxes : Provider.mailboxes(Provider.DEFAULT_ID))
+  readonly property bool canArchive: unified
+    ? Unified.sharedCapability(unifiedProviders, "archive")
+    : (!current || current.canArchive)
```
huacnlee, line 977: **P2: The capability and mailbox intersection drops account-level restrictions**

This computes an intersection over provider IDs only; it does not include each host's `capabilityRefusals` or `absentMailboxes`. For example, the JMAP provider may support Archive while one account has no Archive mailbox; the single-account view hides it correctly, but unified mode reintroduces it.

The QML reproduction showed a JMAP account with `canArchive=false` while `service.canArchive=true`. The same issue affects missing mailbox rows and related hints. Intersect the actual capabilities and mailbox directories of every participating account, and recompute when those dynamic restrictions change.

patrykbart:
All five fixed, each with a regression and each mutation-checked. Two of them I would not have found, and one of them is mine rather than the merge's.

**The sending identity.** You are right that this is not a unified-only fault, and it is worse than that: the address-matching fallback in `send()` is something *I* added on this branch, and `send()` is not gated on `unified`, so it applied with the merged view off as well. On `main` the question could not arise — `send` went to `current` and `begin` seeded `accountId` from `activeAccountId`, so owner and current were the same account.

`ComposeView.submit` now names its mailbox, and the choice is resolved in one place, `sendHostFor`: a named mailbox that is not here is a refusal rather than permission to guess, and the address is consulted only when nothing named one. The regression puts both mailboxes on one `shared@example.com` and asserts that naming B is what settles it — with nothing named, the first match is genuinely all there is to go on, so that stays as it was.

**Refresh.** Confirmed, and the mechanism is `active` rather than `refresh` itself: `refresh` will not reload a loaded list without it, `onActiveChanged` is what asks for the first one, and the poll timer checks it before extending one. So the merged list moved counts and not mail. `applyLiveHosts` gives it to every mailbox in the merge, along with `windowOpen`, which `refresh` also asks about — and it runs when the setting is toggled, since `refreshCurrent` returns early there: the account on screen has not moved.

**The draft id.** Fixed, and decoded from the id rather than from `unified`, which is the part of your note that mattered — the composer can be opened from a merged list and saved after the reader has left it, and a conversion gated on the mode would break exactly then. A bare id cannot hold the separator, so asking is safe unconditionally. There is a regression for the after-the-fact case specifically.

**Conversation members.** Confirmed, including the null host. Composed at the boundary like `selectedId`: `composeThread` and `composeMembers`, plus a merged row's own `thread.memberIds`, which is what `rowHoldsMember` compares when an archive asks whether the reader is showing a member. The thread's own id I left raw and said why — it is a provider's handle for a conversation and is never routed to a mailbox. Tell me if you would rather it were composed for symmetry.

**The intersection.** Confirmed. `unifiedAbilities` reads the hosts, so the intersection follows the refusals and absent mailboxes as they arrive rather than answering from a generic provider. I deleted `sharedCapability`, `sharedMailboxes` and `hasSharedMailbox` rather than leaving them beside the new pair: they were only reachable from their own tests, and a function that gives the wrong answer is worse kept than removed.

One honest limit there. The refusals arrive through `api`, a read-only loader item, so I could not inject the divergent case into a real host in a QML test — `bob().api = {...}` does not assign. So the divergence is held in node (`everyMailboxCan` with one mailbox that refuses archive, `sharedMailboxRows` with one that lacks a row), the Service test asserts every row of `unifiedAbilities` equals its own host's answer, and `test_source.sh` guards that the call sites read the hosts and that the deleted helpers do not come back. Both guards were checked against their faults. If you know a way to reach `api` in an offscreen test I will replace all of that with the direct assertion.

`make validate` green: 410 assertions.

**The combined tree** is rebuilt on `main` and green at 437 — `selectionIsPreview` from `reading`, `markPreviewRead` routing and decoding the composite id, and the three regressions for that seam. It is on my fork's `dev`.

Note for whichever of the two lands second: #94 cannot add to `account/MailAccount.qml` without crossing `test_source.sh`'s 128 KiB ceiling, which `main` is already just under. That is asked on #94 and is not a question about this PR.

PR #94, opened 2026-09-03
Preview messages as the cursor moves

patrykbart:
## Summary

Allow users to preview a message as the list cursor moves, without opening it or marking it read immediately.

## Behaviour

- Preview is opt-in and disabled by default.
- `j`, `k`, and arrow keys show the message after a short settle period.
- A preview does not add reader history, mark the message read on arrival, or fetch remote images.
- The message is marked read only after the cursor remains on it for the configured dwell time.
- Enter still opens the message immediately and marks it read as before.
- Moving away cancels the pending dwell; returning to a preview restarts it.
- Preview is disabled in compact layouts and while another page, dialog, or composer is active.
- Existing single-mailbox and unified-mailbox routing remain account-owned.

## Validation

- Added regression coverage for delayed/failed bodies, zero and invalid dwell values, away-and-back previews, image privacy, read marking, compact layouts, and action routing.
- `node tests/test_model.js`
- `node tests/test_message.js`
- `node tests/test_unified.js`
- `bash tests/test_source.sh`

The local QML runner requires the Qt ICU library `libicui18n.74.dylib`, which is unavailable in the current macOS environment.

### DIFF
--- a/App.qml
+++ b/App.qml
@@ -431,6 +431,9 @@ Item {
 
   function close() {
     closingFromHost = true
+    // A window that is gone shows nothing, so a dwell counting down in it has
+    // nothing left to be counting for.
+    markReadDwell.stop()
     if (compose.opened || compose.parkedForSend) saveComposeRecovery()
     opened = false
     if (service) service.windowOpen = false
@@ -544,6 +547,21 @@ Item {
     previewCursor()
   }
 
+  // Whether there is a preview on screen to be talking about.
+  //
+  // Never in a narrow window: there the reader takes the list's place, so
+  // every press would navigate away from the list being moved through and
+  // there would be nothing left to move. Never over a page, a draft or the
+  // calendar either, for the same reason — the list is not what is on screen.
+  //
+  // Asked twice, when the cursor moves and again when the dwell fires, because
+  // every part of it can change in between: a page opens, a draft is started,
+  // the window is dragged across the breakpoint. A message that is no longer
+  // being shown is not a message being read, however long the cursor has sat
+  // on its row.
+  readonly property bool canPreview: !!service && service.previewOnCursor
+    && !compact && !showPage && !composing && !calendarVisible
+
   // Moving is not opening, and this is the difference.
   //
   // It used to open whatever it landed on, which made stepping through a list
@@ -551,17 +569,16 @@ Item {
   // preview is off unless it was asked for, it pushes no history — Escape and
   // Back mean what they meant — and it does not mark anything read. A dwell
   // does that, which is what stops a held arrow key from reading a mailbox.
-  //
-  // Never in a narrow window: there the reader takes the list's place, so
-  // every press would navigate away from the list being moved through and
-  // there would be nothing left to move.
   function previewCursor() {
     markReadDwell.stop()
-    if (!service || !service.previewOnCursor || compact) return
-    if (cursorId === "" || showPage || composing || calendarVisible) return
+    if (!canPreview || cursorId === "") return
     // A message already open stays open on its own terms: it was opened, and
     // re-selecting it as a preview would take back the read mark it earned.
     if (currentView === "reader" && service.selectedId === cursorId) return
+    // Per message, the same as opening one. Insisting on a document the bounds
+    // refused is an answer about the message it was given for, and the row the
+    // cursor moved to is a different message.
+    reader.forceRichAnyway = false
     service.select(cursorId, true)
     if (service.markReadDelaySec <= 0) {
       service.markPreviewRead(cursorId)
@@ -582,7 +599,11 @@ Item {
     repeat: false
     onTriggered: {
       if (!root.service || dwelledOn === "") return
+      if (!root.canPreview) return
       if (root.cursorId !== dwelledOn) return
+      // And it has to still be the message on screen: a search and a mailbox
+      // switch both drop the selection without moving the cursor off the row.
+      if (root.service.selectedId !== dwelledOn) return
       root.service.markPreviewRead(dwelledOn)
     }
   }
--- a/App.qml
+++ b/App.qml
@@ -580,11 +580,18 @@ Item {
   // does that, which is what stops a held arrow key from reading a mailbox.
   function previewCursor() {
     markReadDwell.stop()
+    markReadDwell.dwelledOn = ""
     previewSettle.stop()
     if (!canPreview || cursorId === "") return
     // A message already open stays open on its own terms: it was opened, and
     // re-selecting it as a preview would take back the read mark it earned.
-    if (currentView === "reader" && service.selectedId === cursorId) return
+    //
+    // A *preview* of the same message is not that. Moving away and back
+    // inside the settle stops the dwell above and used to return here, so the
+    // message the cursor was sitting on never became read at all — the id
+    // matched, which is exactly what a preview does while not being open.
+    if (currentView === "reader" && service.selectedId === cursorId
+      && !service.selectionIsPreview) return
     // Per message, the same as opening one. Insisting on a document the bounds
     // refused is an answer about the message it was given for, and the row the
     // cursor moved to is a different message.
@@ -615,13 +622,40 @@ Item {
     if (id === "" || id !== cursorId || !canPreview) return
     if (currentView === "reader" && service.selectedId === id
       && !service.selectionIsPreview) return
-    service.select(id, true)
+    // Coming back to the message that is still the preview restarts its
+    // dwell rather than asking for it again.
+    if (service.selectedId !== id || !service.selectionIsPreview)
+      service.select(id, true)
+    markReadDwell.dwelledOn = id
+    armReadDwell()
+  }
+
+  // Whether the previewed message is actually on screen.
+  //
+  // The dwell measures time spent looking at something, so it cannot start
+  // before there is anything to look at. It began when the request went out,
+  // so a slow fetch spent the whole dwell loading and marked read a message
+  // whose body never appeared — worst at a zero dwell, which marked it read
+  // before the request had even been made. A fetch that fails never paints,
+  // so it never arms this at all.
+  readonly property bool previewShowing: !!service
+    && service.detailPainted && !service.detailLoading
+
+  onPreviewShowingChanged: if (previewShowing) armReadDwell()
+
+  function armReadDwell() {
+    if (!service || markReadDwell.dwelledOn === "") return
+    if (!canPreview || !previewShowing) return
+    // Still the message on screen: a search and a mailbox switch both drop the
+    // selection without moving the cursor off the row.
+    if (service.selectedId !== markReadDwell.dwelledOn) return
     if (service.markReadDelaySec <= 0) {
-      service.markPreviewRead(id)
+      var now = markReadDwell.dwelledOn
+      markReadDwell.dwelledOn = ""
+      service.markPreviewRead(now)
       return
     }
     markReadDwell.interval = service.markReadDelaySec * 1000
-    markReadDwell.dwelledOn = cursorId
     markReadDwell.restart()
   }
 
--- a/App.qml
+++ b/App.qml
@@ -446,6 +446,9 @@ Item {
   // override is a per-message decision about one specific message and does not.
   function openMessage(id) {
     if (!service) return
+    // Opening marks it read itself, so a dwell still counting down for the
+    // same message has nothing left to do.
+    markReadDwell.stop()
     pendingComposeMode = ""
     pendingDraftId = ""
     reader.forceRichAnyway = false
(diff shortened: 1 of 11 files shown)


huacnlee (changes requested):
The shape of this is right — the dwell holding its own id, the flag through `select`, the compact guard, and refusing to push history are all the correct calls, and I confirmed by mutation that the compact guard is genuinely measured: dropping it fails `test_a_narrow_window_does_not_preview`, and making the preview a real `select` fails three others.

I pushed two commits to the branch (`094d46d`, `776c12b`) with the faults that had one obvious fix. Four things still need you.

## What I pushed

**A preview fetched the sender's remote images.** `select()` sets `remoteImagesAllowed = alwaysShowImages` without asking whether this is a preview, so with the standing "always show images" answer on, a body arriving for a row the cursor passed over runs `renderSource`, which queues every remote source through `image-fetch.sh`. That tells each sender's host that this address opened this mail at this moment — the exact sentence the reader's own notice puts on screen, and the exact class of thing AGENTS.md says is never bought without an explicit act. It is also worse than the read mark you were careful about: the read mark waits two seconds, the fetch goes out as soon as the body lands, which for a cached body is immediately. Hold `j` down an inbox of newsletters with images on and every one of them is told. Now gated on `!selectionIsPreview`, with a line in `test_source.sh` beside the rest of the image policy. If you would rather the dwell *promote* a preview into a real read, images and all, say so — but then the promotion has to re-render, and blocking until `Enter` seems the honest reading of "this was not opened".

**The dwell's guards were decided once and never rechecked.** `compact`, `showPage`, `composing` and `calendarVisible` are read when the cursor moves, so opening Settings, starting a draft, going to the calendar, dragging the window under the breakpoint, or shutting the window entirely all leave the timer running — and two seconds later a message nobody can see is marked read on the server. `close()` never stopped it, and Escape at the list root calls `requestClose()`. A search or a mailbox switch calls `clearSelection()` without moving the cursor, which slips past the `cursorId` check for the same reason. The predicate is now `canPreview`, asked at the move and again in `onTriggered`, plus `service.selectedId === dwelledOn`, plus a stop in `close()`. Five new assertions; all five fail against the previous `App.qml`.

**`forceRichAnyway` leaked between previewed messages.** `openMessage` clears it because it is a per-message decision; `previewCursor` did not. Show a heavy message anyway, then move the cursor, and every row afterwards is laid out in the form the bounds refused, on the shell's GUI thread.

**`markReadDelaySec` read a junk value as 0**, which is your "read it at once" setting. `Number` turns `null`, `false` and `""` into 0, so a settings file that lost the key or holds a word opted the user into the least safe answer silently. Non-numbers now read as the default, on the way in as well as out.

## What needs you

**`actOnCursor`'s `wasOpen` cannot tell a preview from an open.** `Enter` on m1, `j` to m2 (previewed, `selectedId` is now m2, view still `reader`), then `e`: `wasOpen` is true, so it calls `openMessage(next)` and marks m3 read. An archive that reads a message is the same class of fault this feature exists to avoid.

**`composeFromCursor` skips the open for a previewed message.** It guards on `service.selectedId !== cursorId`, which a preview satisfies, so `r`/`a`/`f` push no `reader` entry: the message is never marked read, and closing the draft lands on the list instead of the message being answered.

Both want `selectionIsPreview` visible above `MailAccount`, which is a shape decision I did not want to make for you.

**The fetch has no settle.** Cancellation is right — `detailSerial` and `abortRequest` are both correct, and I could not construct a stale response landing in the reader — but a request still *starts* per keystroke, and on IMAP that is a `mail-transport.sh` curl process, a TLS handshake and a LOGIN per row, spawned and killed as the key repeats. Thirty rows held under `j` is thirty connections. A 150–250 ms settle before `select` costs nothing perceptible and would have contained the image problem on its own.

**`test_a_preview_marks_nothing_read_on_arrival` cannot fail.** The fake service's `select` marks nothing read whatever flag it is given, so it passes even with the preview mutated into a real open — it asserts that `App.qml` passed a flag, not that the fault is absent. The behaviour it names lives at `MailAccount.qml:1071` (`if (summary.unread && !root.selectionIsPreview)`) and has no test at all.

## Smaller

`docs/KEYS.md:40` still says `j`/`k` "move the cursor without opening", which is conditionally false now. Zoom is bound to the `reader` context only, so a previewed message can be read but not zoomed while the same message opened with `Enter` can — worth deciding whether the zoom keys should follow the message on screen rather than the nav entry. And Escape at the list root closes the window without clearing the selection, so it reopens showing a stale previewed message beside the list.

## On rebasing

Eight commits behind, no textual conflicts, and the full suite on a trial merge is 219 passed / 0 failed. One thing to add when you do: #83's `LabelPicker` is an `anchors.fill` overlay that is not on the nav stack and so is not in `canPreview`'s guard list — pressing `v` during a dwell still lets the timer mark read a message that is about to be moved. `!labelPicker.opened`, or put the picker on the stack.

patrykbart:
Thank you — and thank you for the two commits. The remote-image one is the finding I would least have liked to ship: the read mark waits and the fetch does not, so a held `j` down a list of newsletters tells every sender before the dwell has even fired once. That is a worse leak than the one I was careful about.

To your question there: **blocking until `Enter` is the honest reading**, and I have not made the dwell promote a preview. A message the cursor passed over was not opened, and the sender should not be told it was; promoting would also mean re-rendering two seconds in, which is a visible change nobody asked for.

## What needed me

**`actOnCursor` and `composeFromCursor`.** Both fixed, and you were right that they wanted `selectionIsPreview` above `MailAccount` — neither can be answered from `selectedId`, because a preview satisfies the id test while not being open. It is a `Service` property now.

The archive case is exactly as you described: `wasOpen` true, `openMessage(next)`, m3 read. The reply case is the mirror — no `reader` entry pushed, so nothing marked it read and the draft closed onto the list.

**The settle.** Added, 180 ms. You are right that cancellation was already correct and that the cost is in *asking*: on IMAP each preview is a curl process, a TLS handshake and a LOGIN, and thirty rows under a held `j` was thirty of them. Short enough that a single deliberate `j` still feels immediate, long enough that a key repeat never gets through.

It would indeed have contained the image problem on its own, but only for a held key — a single deliberate move would still have fetched. Your gate is the actual fix and the settle is the cost control; I would not want either alone.

**`test_a_preview_marks_nothing_read_on_arrival`.** You are right that it cannot fail, and that the behaviour it names lives at `MailAccount.qml` where no unit test can reach it. Rather than delete it I have guarded the real thing in `test_source.sh`, beside your image line: removing `!root.selectionIsPreview` from the read mark now fails the suite. Confirmed by mutation.

## Smaller

`docs/KEYS.md` fixed. The label picker is in `canPreview` now — `!labelPicker.opened`, since putting an `anchors.fill` overlay on the nav stack is a bigger change than this PR should make. And `close()` clears a preview, so Escape at the list root no longer leaves a stale message beside the list on reopen.

**Zoom I have left alone, and I would rather you decided.** I did make it follow the body on screen, then found `test_a_reader_only_key_is_dead_in_the_list` asserting "nothing to zoom from the list" — so reader-only is a tested decision, not an oversight, and overriding it uninvited seemed wrong. My view is that zoom is about the body on screen rather than the nav entry, and that the size being a stored preference makes it harmless from a list with nothing showing. But it is your keymap and the test says the current answer was deliberate. Say the word and it is three lines plus that test.

## Verification

`make validate` green, 251 QML assertions.

Mutation-checked, each one separately: dropping `&& !service.selectionIsPreview` from `wasOpen` fails the archive test; dropping `|| service.selectionIsPreview` from `composeFromCursor` fails the reply test; dropping `!labelPicker.opened` fails the picker test; and removing the read-mark gate fails `test_source.sh`.

The tests that asserted the preview arrives synchronously now wait for the settle, which is the behaviour rather than an inconvenience — and one of them asserts the point directly: three rows crossed, one request.

patrykbart:
Rebased onto `main` at 0.8.1. One conflict, in `tests/test_source.sh`: #106 moved the image fetch to `image-fetch.py`, so this branch's `max-redirs = 0` assertion against `image-fetch.sh` was checking a file that no longer exists. Dropped it — the redirect and DNS policy are the Python transport's now, and #106's own assertions cover the swap.

The assertion this branch adds is unchanged and still the point: `remoteImagesAllowed = alwaysShowImages && !selectionIsPreview`, so a standing "always show images" answer does not fetch the sender's pictures for a message the cursor merely passed over. That is the same argument #106 is making about who gets told what, one layer up: a preview would tell the sender's host that this address opened this mail at this moment, which is exactly what the read mark waits for a dwell to avoid.

292 assertions green.

huacnlee (changes requested):
Thanks for the fixes and rebase. I re-reviewed this alongside #93 and reproduced two remaining preview defects with offscreen Qt 6.11.2 tests. Please fix these before merging:

1. [P2] Start the read dwell only once the preview body is ready. App.qml:581–588 starts markReadDwell immediately after select(), and the timeout at lines 599–606 only checks cursor/selection identity. A regression keeping detailLoading=true and detailPainted=false still receives markPreviewRead: a slow request can therefore mark a message read while its body has never been displayed. Gate the dwell on the selected preview becoming ready, and test delayed/failed detail loads as well as the successful case.

2. [P2] Returning to a preview must restart its dwell. Open m1, move to m2 and let its preview settle, then move to m3 and back to m2 before the 180 ms settle. previewCursor() stops the dwell and then returns at App.qml:550 because selectedId already equals cursorId. That guard mistakes a preview for an explicitly opened message; unlike previewSettled(), it does not check selectionIsPreview. The regression shows m2 remains unread indefinitely after returning. Preserve the early return for a genuinely opened message, but let an existing preview resume its dwell, with a rapid-away-and-back regression.

Security verdict: PASS for the reviewed preview-arrival/image gating. A real MailAccount.select/detail-callback test with a synthetic Gmail resource confirms that preview arrival does not call markRead and that alwaysShowImages does not queue remote images during preview; an explicit open does call markRead. Provider delivery was stubbed, so this does not claim live Gmail delivery or a full transport audit. The two functional failures above still block merging.

The zoom preference can remain as it is for now; it is not a blocker. If #93 is merged as well, the combined tree must read selectionIsPreview from the reading account and route markPreviewRead through the composite account/message ID. Please verify that integration rather than assuming separate green suites are sufficient.

patrykbart:
Both confirmed and fixed.

**1. The dwell started at the request, not at the body.** Fixed, and it was worse at the other end of the setting than in the middle: with `markReadDelaySec` at 0 — "read it as soon as it is shown" — the message was marked read *before* the request had been made, since that branch called `markPreviewRead` inline and never looked at the reader state at all.

The dwell is armed when the body lands now, through a `previewShowing` that asks for `detailPainted && !detailLoading`. A fetch that fails never paints, so it never arms at all, which is the right answer: a message that could not be shown was not read. `previewCursor` also drops what the dwell was waiting for when the cursor moves off, so a body landing late cannot arm a dwell for a row nobody is on any more.

Three regressions, as you asked — a slow body, one that never arrives, and a zero dwell that still waits.

**2. Returning to a preview.** Fixed exactly as you framed it: the early return was written for a message that was *opened*, where re-selecting it as a preview would take back the read mark it earned, and a preview satisfies the id test while not being open. It asks `!service.selectionIsPreview` now, and a preview returned to restarts its dwell without asking for the body again.

Worth recording how that test went, since it is the same mistake in miniature: I first wrote the away-and-back regression starting from the list, and it passed against the unfixed code. The early return needs `currentView === "reader"`, which a preview from the list never reaches — so the test was measuring nothing until I started it from an opened message, as your repro does. Mutation is what caught it, not the suite.

Both fixes mutation-checked separately: dropping the readiness gate fails three of the new tests, and restoring the old early return fails the away-and-back one.

296 assertions green.

**On the combined tree.** Verified, not assumed, and it was worth doing. `selectionIsPreview` is `reading`'s — the mailbox holding the selection — which is also where `detailPainted` and `detailLoading` come from, and the dwell now waits on those: reading any of the three from the visible mailbox would arm a dwell against a body belonging to a different account. `markPreviewRead` routes and decodes the composite id, since the dwell fires with the id it started on. Three regressions in `tst_unified_routing.qml` cover that seam, each mutation-checked; neither branch's own suite can see any of it, because the preview tests have one mailbox and the merge tests have no preview. Combined tree is 342 green, on my fork's `dev` if you want to look, and those tests will land with whichever of #93 and #94 merges second.

Noted on the zoom preference — leaving it as it is.

patrykbart:
Rebased onto `main` at c5d87ef. Two conflicts, both semantic:

`MailAccount.markPreviewRead` and #111's rail block land in the same place — kept both. And `actOnCursor`'s `wasOpen` is now refined by both sides, which compose rather than compete: `currentView === "reader" && !service.selectionIsPreview && (service.selectedId === acted || Model.rowHoldsMember(row, service.selectedId))`. The rail's question is whether the reader is showing a member of the acted row; mine is whether it was opened at all. Both have to hold.

**One thing needs you, and it is not really about this PR.** `account/MailAccount.qml` is 130702 bytes on `main` against `test_source.sh`'s 131072 ceiling — 370 bytes of headroom, which #111 consumed. Anything added to that file now fails the gate.

I have cut this branch's footprint as far as I honestly can: the three preview decisions moved into `Model.js`, which is where AGENTS.md wants a decision and makes them answerable from node rather than only through a running reader — `marksReadOnArrival`, `showsRemoteImages`, `previewReadable`, twelve assertions, each mutation-checked. The reasoning moved with them rather than being cut; two of those comment blocks are yours from `e7a0dee` and I was not going to delete them to buy bytes.

That leaves 1035 bytes of property, changed signature and one five-line function, so the file is 665 bytes over. I can get under it only by stripping comments the house style asks for, which seemed the wrong trade to make silently.

Your call, and I will do whichever you prefer:

1. Raise the ceiling for tracked source, keeping the tight one for assets — the check reads "keep large assets out of the clone", and a 128 KiB QML file is not the thing it was written to catch.
2. Split `MailAccount.qml`. Probably right eventually, but far too large a refactor to smuggle into this PR.
3. Tell me which comments to cut.

Everything else is green with the ceiling lifted: `make validate` clean, 384 assertions. The two `test_source.sh` guards follow the call sites and still bite — I verified both against their faults with the ceiling raised, since the size check exits before they run.

PR #95, opened 2026-09-03
Move a message back to the inbox, and out of the label it was filed under

patrykbart:
`unarchive` has been in the action vocabulary all along — in `labelChangesFor`, in `actionCapability`, and `flagPlanForLabels` already turns it into a move into `INBOX`. Nothing ever offered it, so there was no way to put a message back in the inbox from inside the application.

It is a row in the message menu now, in place of Archive rather than beside it.

## The label is the point

From a label's list, moving a message to the inbox also takes it out of that label, and the row says so. A label somebody files things under — TODO, Waiting, Later — is a queue, and one that keeps everything ever put in it only grows.

This writes the same pair a `label:` move already writes, using #83's `rawLabelId` and `sourceLabelId`, so a folder name never reaches a Gmail remove list and a system label is never removed. `isSystemLabelId` is one fact asked in two places — by `labelChangesFor` before it removes anything, and by the menu before it promises a removal — so the wording cannot offer one that will not happen.

## Where the row appears

Only where it is the truth: out of the inbox, and not somewhere that has its own verb.

`inInbox === false` alone would put it in Spam, Trash, Drafts and Sent, where adding `INBOX` is not what "move to the inbox" means. `labelChangesFor` removes neither `SPAM` nor `TRASH`, so a spam message would carry `INBOX` *and* `SPAM` and stay in Spam while the row claimed otherwise; on IMAP the same press physically relocates a draft out of `\Drafts`. Spam has its own verb and trash has its own endpoint for exactly that reason.

`summarize` gained `inSpam` and `isSent` to say this. IMAP synthesises all four from the folder in `labelIdsFor`, so one rule serves both kinds of provider.

The row the menu reads is the one `applyLabelChange` wrote, and that maintained three of these flags and left the rest as they were — so reporting spam from the inbox produced a row carrying `SPAM` whose `inSpam` still said false, and the menu went on offering the press this rule exists to prevent. Every flag that mirrors a label follows the labels now. The cache version goes with it, because the page a menu reads is the page the cache paints first and rows written before `inSpam` existed answer "not spam" to a menu that asks; bodies are separate files, so the bump costs one cold list load.

## Whether the row survives

`survivesAction` keeps its `rawQuery` string third argument and gains two more: whether the provider files by label or by folder, and which label the list is. The first is the one thing it cannot infer, and `unarchive` is the action that turns on it.

The second is there so the two answers cannot disagree. `survivesAction` asks `labelChangesFor` whether the label actually comes off rather than reading the system-label rule a second time — in a system label's list the label stays on the message, and so the row stays in the list. A label view with nothing naming its label still evicts: a row that leaves and should not have comes back on the reload `invalidatesPage` asks for, and one that stays and should not have is stale until something else loads the list.

A folder provider answers with a `UID MOVE`. The message is given a new UID in `INBOX`, nothing parses `COPYUID`, and the row it left would point at a message that is no longer there — opening it says so, a star succeeds against nothing, and `invalidatesPage` would not reload to correct it. So on a folder provider the row never survives.

Scoped to `unarchive` on purpose: a `label:` move from All mail on IMAP has the same shape, and extending it there uninvited seemed worse than naming it.

## Verification

`make validate` green, 254 QML assertions.

`tests/test_model.js` covers the label arithmetic, the system-label refusal, and both providers' answers to `survivesAction`. `tests/qml/tst_message_menu_inbox.qml` covers the QML layer, which had none: the visibility rule against all four of spam, trash, sent and draft, both wordings, the system-label and folder-provider cases, and the routing.

Mutation-checked five ways: reverting the gating to `inInbox` alone fails four tests; removing the row from `menuRows` fires the `test_source.sh` guard and fails the keyboard test, which walks the cursor onto the row rather than looking it up; removing the folder-provider branch from `survivesAction` fails a node assertion; reverting `applyLabelChange` to the three flags it used to maintain fails the spam case; and reverting `survivesAction` to reading the system-label rule itself fails a pair asserted against `labelChangesFor` rather than against a constant.

## Release Notes

- A message can be moved back to the inbox from the message menu, which nothing offered before. The row appears in place of Archive, on messages that are out of the inbox.
- Done from a label's list it also takes the message out of that label, so a label used as a queue empties as things are dealt with instead of only growing.

### DIFF
--- a/App.qml
+++ b/App.qml
@@ -688,8 +688,12 @@ Item {
     var wasOpen = currentView === "reader" && service.selectedId === acted
     // Worked out before the action, while the row still has neighbours.
     var next = Model.cursorAfterRemoval(service.messages, acted)
+    // The same four facts `MailAccount.act` decides with. Asking with three of
+    // them made the cursor repair disagree with the list it repairs: moving a
+    // message back to the inbox removes the row on a provider that moves, and
+    // this read it as staying.
     var leaves = !Model.survivesAction(service.mailboxKey, action,
-      service.rawQuery)
+      service.rawQuery, service.hasLabels)
     if (!service.act(acted, action)) return false
     if (!leaves) return true
     // The row is going and the cursor must not go with it: a cursor on a
--- a/App.qml
+++ b/App.qml
@@ -688,12 +688,12 @@ Item {
     var wasOpen = currentView === "reader" && service.selectedId === acted
     // Worked out before the action, while the row still has neighbours.
     var next = Model.cursorAfterRemoval(service.messages, acted)
-    // The same four facts `MailAccount.act` decides with. Asking with three of
+    // The same five facts `MailAccount.act` decides with. Asking with three of
     // them made the cursor repair disagree with the list it repairs: moving a
     // message back to the inbox removes the row on a provider that moves, and
     // this read it as staying.
     var leaves = !Model.survivesAction(service.mailboxKey, action,
-      service.rawQuery, service.hasLabels)
+      service.rawQuery, service.hasLabels, service.rawLabelId)
     if (!service.act(acted, action)) return false
     if (!leaves) return true
     // The row is going and the cursor must not go with it: a cursor on a
--- a/account/MailAccount.qml
+++ b/account/MailAccount.qml
@@ -1370,7 +1370,8 @@ Item {
     var before = index >= 0 ? messages[index] : previewMessages[previewIndex]
     var sourceLabelId = hasLabels ? rawLabelId : ""
     var updated = Model.applyLabelChange(before, action, sourceLabelId)
-    var survives = Model.survivesAction(mailboxKey, action, rawQuery, hasLabels)
+    var survives = Model.survivesAction(mailboxKey, action, rawQuery, hasLabels,
+      sourceLabelId)
 
     if (action === "markRead" && before.unread) inboxUnread = Math.max(0, inboxUnread - 1)
     if (action === "markUnread" && !before.unread) inboxUnread = inboxUnread + 1
--- a/account/MailAccount.qml
+++ b/account/MailAccount.qml
@@ -1370,7 +1370,7 @@ Item {
     var before = index >= 0 ? messages[index] : previewMessages[previewIndex]
     var sourceLabelId = hasLabels ? rawLabelId : ""
     var updated = Model.applyLabelChange(before, action, sourceLabelId)
-    var survives = Model.survivesAction(mailboxKey, action, rawQuery)
+    var survives = Model.survivesAction(mailboxKey, action, rawQuery, hasLabels)
 
     if (action === "markRead" && before.unread) inboxUnread = Math.max(0, inboxUnread - 1)
     if (action === "markUnread" && !before.unread) inboxUnread = inboxUnread + 1
@@ -1518,6 +1518,7 @@ Item {
     if (action === "unstar") return "Unstarred"
     if (action === "markRead") return "Marked read"
     if (action === "markUnread") return "Marked unread"
+    if (action === "unarchive") return "Moved to Inbox"
     if (action === "spam") return "Reported as spam"
     // Named, not "Moved": the destination was chosen a keystroke ago from a
     // list of thirty, and a note that does not say which one leaves the only
--- a/account/Model.js
+++ b/account/Model.js
@@ -142,17 +142,44 @@ function setupActionLabel(state, provider, authKind) {
 // have had to be carried, and correctly put back, by all four.
 var MOVE_PREFIX = "label:"
 
+// Gmail's own labels are upper case with no `Label_` prefix; a user's carry one
+// or are a folder name on IMAP. Only the second kind is a place a message is
+// filed under and can be taken out of, and `survivesAction` asks the same
+// question of the same fact so the two cannot disagree about a row.
+var SYSTEM_LABEL_IDS = ["INBOX", "UNREAD", "STARRED", "IMPORTANT", "SENT",
+  "DRAFT", "TRASH", "SPAM", "CHAT"]
+
+function isSystemLabelId(id) {
+  var value = String(id || "")
+  if (SYSTEM_LABEL_IDS.indexOf(value) >= 0) return true
+  return value.indexOf("CATEGORY_") === 0
+}
+
 function labelTarget(action) {
   var verb = String(action || "")
   if (verb.indexOf(MOVE_PREFIX) !== 0) return ""
   return verb.slice(MOVE_PREFIX.length)
 }
 
-function survivesAction(mailboxKey, action, rawQuery) {
+// `labels` is whether the provider files by label rather than by folder, which
+// is the one thing this cannot infer and the one thing `unarchive` turns on: a
+// folder provider answers it with a UID MOVE, so the message leaves the list
+// with a new id and a surviving row would point at nothing.
+function survivesAction(mailboxKey, action, rawQuery, labels) {
   var key = String(mailboxKey || "inbox")
   var verb = String(action || "")
   if (verb === "trash") return key === "trash"
   if (verb === "untrash") return key !== "trash"
+  if (verb === "unarchive") {
+    // Moving relocates on a folder provider: the message is given a new UID in
+    // INBOX, nothing parses COPYUID, and the row it left would point at a
+    // message that is no longer there — opening it says so, a star succeeds
+    // against nothing, and no reload corrects it.
+    if (labels !== true) return false
+    // On a label provider the row leaves a label's list, because the label
+    // came off, and stays in a mailbox or a search, which still contain it.
+    return String(rawQuery || "") === ""
+  }
   // A move takes INBOX away exactly as archive does, so it leaves exactly the
   // lists archive leaves. Said once, because two branches with the same answer
   // are two places for it to drift.
@@ -170,7 +197,16 @@ function labelChangesFor(action, sourceLabelId) {
   if (action === "star") return { add: ["STARRED"], remove: [] }
   if (action === "unstar") return { add: [], remove: ["STARRED"] }
   if (action === "archive") return { add: [], remove: ["INBOX"] }
-  if (action === "unarchive") return { add: ["INBOX"], remove: [] }
+  // Moving back to the inbox takes the message out of the label whose list it
+  // was found in, the same way a move does and for the same reason: a label
+  // somebody files things under is a queue, and one that keeps everything ever
+  // put in it only grows. `sourceLabelId` is already "" on a provider whose
+  // labels are folders, so a folder name never reaches a Gmail remove list.
+  if (action === "unarchive") {
+    var filed = String(sourceLabelId || "")
+    if (filed === "" || isSystemLabelId(filed)) return { add: ["INBOX"], remove: [] }
+    return { add: ["INBOX"], remove: [filed] }
+  }
   if (action === "spam") return { add: ["SPAM"], remove: ["INBOX"] }
   // A move is archive with somewhere to go. Where a folder is a label, putting
   // a message in one is adding that label and taking INBOX away -- the same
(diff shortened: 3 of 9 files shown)


huacnlee (changes requested):
The direction is right and the Gmail label arithmetic is right, but this needs a rebase before anything else: #83 landed while it was open and already added the exact state you are adding, under a different name.

I pushed `cb09d27` with the three faults that had one obvious fix. The rest need you.

## Rebase first — #83 got there

`MailAccount.rawLabelId` and `Service.rawLabelId` are your `viewedLabelId`, set in `selectLabel` and cleared in `selectMailbox`. `Model.labelChangesFor(action, sourceLabelId)` and `applyLabelChange(summary, action, sourceLabelId)` already take a third argument meaning "the label this list is showing", used by `label:<id>` moves. Please drop `viewedLabelId` and use `rawLabelId`.

The one that will bite silently: **`Model.survivesAction`'s third parameter is already taken on main.** It is `rawQuery`, a string; this PR makes it `inLabelView`, a boolean. Same position, different type, and git will merge that without a word — a truthy query string then reads as "in a label view" at every existing call site.

Two things #83 also got right that this does not:

- It clears `rawLabelId` in `search()`. Without that, visiting a label and then typing a search leaves the menu saying "Move to Inbox and remove label" over a hit that may never have been filed under it, and sends that label id in `removeLabelIds` while `survivesAction` evicts the row. That is the destructive half firing on the wrong message; I fixed it on this branch, but it disappears anyway when you move to `rawLabelId`.
- It computes `sourceLabelId = hasLabels ? rawLabelId : ""`, so an IMAP folder name never reaches a Gmail-shaped `remove` list. Right now it does, and it is harmless only because `flagPlanForLabels` drops ids it does not recognise and because `has(added, "INBOX")` happens to be written after `has(removed, "INBOX")`.

## What I pushed

**The new row is mouse-only.** `unarchiveRow` was added to the `Column` and left out of `menuRows` — the array `selectableRows`, `moveCursor` and `runCursor` all index, and that `MenuActionRow.selected` compares `collection.indexOf(root)` against. Right-click an archived message and `j`/`k` steps over "Move to Inbox" into the hole the now-invisible `archiveRow` left behind; Enter can never fire it and it never highlights. Nothing in the suite touches either menu, so I also added a `test_source.sh` guard that fails when a menu draws a row it does not list; removing `unarchiveRow` from the array now fails the suite.

**`App.actOnCursor` asked `survivesAction` with two arguments while `MailAccount.act` asked with three.** In exactly the case this feature exists for — unarchive from a label's list — the account removes the row and the window computed `leaves === false`, so the cursor stayed on a row that had gone and the reader stayed open on a message no longer listed. That is the failure the comment above it describes.

Also: `actionLabel` had no `unarchive` case, so a successful move said "Done".

## What needs you

**IMAP: the row survives while its id dies.** `flagPlanForLabels` turns `add: ["INBOX"]` into a `UID MOVE`, the message gets a new UID in the destination, and nothing parses `COPYUID` — `ImapClient` always calls back with `null`. Every existing IMAP action survives that either by deleting the row or by not moving at all (archive from the Archive folder is a no-op through the `plan.move !== group.folder` guard). `unarchive` breaks the pattern: `survivesAction(key, "unarchive", false)` is true for the Archive, Junk, Sent and Drafts rows, so the row stays pointing at `uid:Archive` while the message is now `newuid:INBOX`. Opening it says "That message is no longer in the mailbox", a star silently succeeds against nothing, and `invalidatesPage` is false so nothing reloads to correct it. `survivesAction` cannot answer this without knowing whether the provider labels or moves.

**Gmail: the row is offered where it lies.** `archived` is only `summary.inInbox === false` and the menu is mounted for every list row with no mailbox condition, so "Move to Inbox" appears in Spam, Trash, Drafts and Sent. `labelChangesFor` never removes `SPAM` or `TRASH`, so a spam message ends up carrying `INBOX` *and* `SPAM` and stays in Spam, while `applyLabelChange` optimistically sets `inInbox: true`. Compare `spam`, which removes `INBOX` as it adds `SPAM`, and untrash, which has its own endpoint precisely because a label add does not undo a trash. The summary already carries `inTrash` and `isDraft`; gating on those, and removing `SPAM` when unarchiving from Spam, would make the label honest. On IMAP the same visibility rule physically relocates drafts and sent mail out of `\Drafts` and `\Sent`.

**`survivesAction` and `labelChangesFor` disagree about system labels.** `labelChangesFor` refuses to remove one, but `survivesAction`'s label argument is only "is a label showing", so a system-looking id evicts the row while the message stays in that label. Latent today because the rail filters `label.system`, but the two answers should come from one fact.

## Two claims that are not true as written

**"A queued quiet action records the label with the request rather than reading it back when it runs."** The only quiet action in the codebase is `markRead` on open, which ignores the label; and in the branch that does run, `runQueuedQuietAction` calls `act(request.id, request.action, true)`, which re-reads the live `viewedLabelId` and discards `request.viewedLabelId`. The recorded field is only consulted on the detached-row fallback. Either drop it or make that path use it.

**"`ImapProtocol` already maps it to a move into INBOX."** `ImapProtocol.actionPlan` — the function holding that case — is dead code: its only callers are `tests/test_imap.js`. The live mapping is `flagPlanForLabels`, which is where the UID problem above lives.

## Tests

`test_model.js` is the only coverage, and two of its new assertions (`survivesAction("all", "unarchive", false)`, and the two-argument call) pass identically with the feature reverted — they are guards rather than measurements. The rest do fail on revert; I checked by mutating `Model.js`. Nothing covers the QML layer: the row's visibility rule, the two wordings, or the `Service` → `MailAccount` routing — which is where every defect above landed.

`make validate` is green on my pushed commit, but the merge with `main` conflicts in `Service.qml`, `account/MailAccount.qml` and `account/Model.js`, and the conflicts are the semantic ones above rather than noise.

patrykbart:
You were right that this needed the rebase before anything else, and #83 got there in a way that deletes most of the change rather than merging with it. Rebuilt on `main` from scratch.

## Rebase — `rawLabelId`, and the signature

`viewedLabelId` is gone. Everything uses `rawLabelId` and the `sourceLabelId` argument that was already there, so `unarchive` now writes the same pair a `label:` move writes rather than a second arrangement of it — which is a much smaller change than what I had.

Both of the things #83 got right that I did not are now inherited rather than reimplemented: the clearing in `search()`, and `sourceLabelId = hasLabels ? rawLabelId : ""` keeping a folder name out of a Gmail remove list. Thank you for fixing the `search()` case on the branch in the meantime.

**`survivesAction`'s third parameter stays your `rawQuery` string.** That was the most valuable thing in the review — a boolean in the same position would have merged without a word and read every non-empty query as "in a label view" at every existing call site. It gains a fourth argument instead.

## IMAP: the row survives while its id dies

Confirmed, and this is what the fourth argument is for. `survivesAction` cannot answer it without knowing whether the provider labels or moves, exactly as you said, so it is told: `survivesAction(mailboxKey, action, rawQuery, labels)`.

On a folder provider `unarchive` never leaves the row. The message is given a new UID in INBOX, nothing parses `COPYUID`, and a surviving row points at nothing.

I deliberately scoped that to `unarchive` rather than to relocation in general. A `label:` move on IMAP from All mail has the same shape — `survivesAction` returns false only when a query is showing — and changing #83's behaviour uninvited seemed worse than naming it. Say the word and I will extend it.

## Gmail: offered where it lies

Fixed as you suggested, by gating rather than by removing SPAM. `archived` is now "out of the inbox, and not somewhere that has its own verb": not trash, not spam, not sent, not a draft. Spam has its own verb and trash has its own endpoint precisely because a label add does not undo either, so the honest answer is not to offer the row there at all rather than to make it do two things.

`summarize` gained `inSpam` and `isSent` for it. IMAP synthesises all four of those from the folder in `labelIdsFor`, so one rule covers both providers — which also handles the case you raised about drafts and sent mail being physically relocated.

## System labels: one fact

`isSystemLabelId` is now the single fact. `labelChangesFor` asks it before removing anything, and the menu's wording asks it before promising a removal — so "Move to Inbox and remove label" can no longer appear over a label that will not be removed. The menu also asks `hasLabels`, so a folder provider's `rawLabelId` is never promised as a Gmail label id.

## Your two corrections

Both accepted, and both claims are gone rather than reworded.

**The queued quiet action.** You are right: `runQueuedQuietAction` calls `act(request.id, request.action, true)`, which re-reads the live label, and the recorded field is only consulted on the detached-row fallback. Rather than make that path use it I have dropped the field — `markRead` is the only quiet action and it ignores the label, so the field was carrying a claim and no weight.

**`ImapProtocol.actionPlan` is dead code.** Confirmed — its only callers are `tests/test_imap.js`. The description names `flagPlanForLabels` now, which is where the UID problem actually lives.

## Tests

The QML layer has coverage now, which is where you correctly noted every defect landed. `tests/qml/tst_message_menu_inbox.qml` covers the visibility rule with the four cases as a data-driven row, both wordings, the system-label and folder-provider cases, and the routing.

One correction to myself there: my first version asserted against a fake `service.act`, which is a contract this component does not have — `run()` emits `actionRequested` and `App` wires it. It spies on the signal now.

Mutation-checked, three ways:

- gating reverted to `inInbox` alone → **4 fail** (spam, trash, sent, draft)
- `unarchiveRow` removed from `menuRows` → **your `test_source.sh` guard fires**, which is a nice thing to have inherited
- the folder-provider branch removed from `survivesAction` → the node assertion fails

I also dropped the two `survivesAction` assertions you identified as guards rather than measurements.

`make validate` green, 241 QML assertions.

huacnlee:
Seven of the nine are fixed and I reproduced each one against the code rather than the description: `viewedLabelId` is gone, `survivesAction`'s third argument is still `rawQuery`, `unarchiveRow` is in `menuRows` with a `test_source.sh` guard that fires when I remove it, `actOnCursor` passes `hasLabels`, `actionLabel` has its `unarchive` case, and the IMAP `labels !== true` branch is right the whole way through to `flagPlanForLabels`. All three of your mutation claims reproduce.

Two did not, and I have pushed a commit for both rather than send this round-trip again. Rebased onto `main`, which now has #89 — the only conflict was two test blocks appended to the same end of `tests/test_model.js`.

**The system-label disagreement is still there, and two comments now assert it is not.** `isSystemLabelId` is called from `labelChangesFor` and from the menu, but not from `survivesAction`. In a system label's list `labelChangesFor("unarchive", "IMPORTANT")` keeps the label while `survivesAction(key, "unarchive", "label:important", true)` returns false and evicts the row. The rail filters system labels so it stays latent, but `account/Model.js:147-149` and the note in `tests/test_model.js` say the two cannot disagree, and as written they could. `survivesAction` now asks `labelChangesFor` whether the label actually comes off, which takes the label id as a fifth argument at both call sites. A label view with nothing naming its label still evicts, since a row that wrongly leaves comes back on the reload `invalidatesPage` asks for and a row that wrongly stays does not.

**The new Spam gate has a live bypass through the optimistic update.** `applyLabelChange` recomputes `unread`, `starred` and `inInbox` and leaves the rest of the summary as it found it, so a row that has just been reported as spam carries `SPAM` in `labelIds` while its `inSpam` still says false — and `survivesAction` has no `spam` branch, so the row stays in the list. Report spam from the inbox, right-click the row that is still sitting there, and "Move to Inbox" is offered: `add: ["INBOX"]`, `remove: []`, the message keeps `SPAM` and stays in Spam. Two presses, and exactly the thing the gate was written to prevent. Every flag that mirrors a label follows the labels now, and the cache version goes with it — rows written before `inSpam` existed answer "not spam" to a menu that asks, and bodies are separate files so the bump costs one cold list load.

Also in that commit: `test_the_keyboard_can_reach_it` searched `menuRows` for `menuRows[4]`, which is true of whatever that array holds — it passed with the row unlisted. It walks the cursor onto the row and activates it now, and fails under that mutation. The fixture's `act`, `ranAction`, `ranId`, `toggleStar` and `unavailableActions` are gone; the menu asks for none of them.

Both fixes are mutation-checked and asserted against `labelChangesFor` rather than against constants, so the pair cannot drift apart again. `make validate` green, 254 QML assertions. Merging.

One thing left alone, worth a separate look if you want it: `archiveRow` gained `&& !root.archived`, but the reader's Archive button and the `e` key still offer Archive on an already-archived message. Harmless — a no-op `INBOX` removal — but the two surfaces now disagree.

PR #97, opened 2026-09-04
calendar: Unwrap XML CDATA in CalDAV responses

denialbb:
Apple iCloud (and some other CalDAV servers) wrap `<c:calendar-data>` XML responses in `<![CDATA[ ... ]]>`.

Currently, `tagText()` in `calendar/Calendar.js` passes the inner text directly to `decodeXml()` without stripping CDATA markers. As a result, the parsed iCalendar payload starts with `<![CDATA[BEGIN:VCALENDAR...`, causing `Ics.eventsFrom()` to fail to detect the `VCALENDAR` root and return zero events.

This change:
- Unwraps `<![CDATA[ ... ]]>` in `tagText()` before decoding XML entities.
- Adds regression test coverage in `tests/test_calendar_feed.js`.
- Verified against live Apple iCloud CalDAV endpoints.

### DIFF
--- a/calendar/Calendar.js
+++ b/calendar/Calendar.js
@@ -78,7 +78,11 @@ function tagText(block, localName) {
   var pattern = new RegExp("<(?:[A-Za-z0-9_-]+:)?" + name
     + "(?:\\s[^>]*)?>([\\s\\S]*?)</(?:[A-Za-z0-9_-]+:)?" + name + ">", "i")
   var match = pattern.exec(String(block || ""))
-  return match ? decodeXml(match[1]) : ""
+  if (!match) return ""
+  var text = match[1]
+  if (/<!\[CDATA\[/i.test(text))
+    return text.replace(/<!\[CDATA\[([\s\S]*?)\]\]>/gi, "$1")
+  return decodeXml(text)
 }
 
 function caldavResponses(xml) {
--- a/calendar/Calendar.js
+++ b/calendar/Calendar.js
@@ -72,17 +72,41 @@ function decodeXml(value) {
     .replace(/&amp;/g, "&")
 }
 
+// An element's text, with both of the encodings a server is allowed to use.
+// RFC 4791 section 9.6 lets a calendar object arrive entity-escaped or inside a
+// CDATA section, and servers send the second: DAViCal wraps every object, and
+// this reached the project from an iCloud account whose calendars read empty.
+// Unwrapped, the payload begins `<![CDATA[BEGIN:VCALENDAR`, which is not a
+// property line, so the parse finds no VCALENDAR root and there are no events.
+//
+// The two encodings cannot be undone in one pass. A CDATA section's content is
+// already literal — `&amp;` inside one is those five characters, and a summary
+// of "Q&amp;A" would come out "Q&A" if the entity pass ran over it. So the
+// markers are cut and the entity pass runs only on the text between sections,
+// which is where the entities an escaping server wrote actually are.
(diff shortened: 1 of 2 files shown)


huacnlee:
The fault is real and the diagnosis is exact — I reproduced it through the real path rather than taking the description's word for it. Same multistatus response in both encodings, on your base commit: entity-escaped gives one event, CDATA-wrapped gives zero. Worth spelling out in the body, because it is the part that makes the change obviously correct: with the markers left on, the first line reads `<![CDATA[BEGIN:VCALENDAR`, `parseProperty` cuts at the first colon and returns the property name `<![CDATA[BEGIN`, so `parse` never opens a VCALENDAR root and `eventsFrom` returns `[]`. The whole collection goes silently empty rather than losing one event.

And it is conformance rather than a workaround for one server: RFC 4791 §9.6 says the calendar data must use standard XML character-data encoding "or the use of a CDATA construct", warning only that a section cannot contain `]]>`. DAViCal has wrapped every object this way for years.

I pushed `16389d3` to the branch with one fix.

**The early return skipped entity decoding for the whole element.** Because you return as soon as a section is present, `decodeXml` never runs, so entity-escaped text *outside* the section is passed on raw: `<c:calendar-data>A &amp; B&#13;<![CDATA[C &amp; D]]> &lt;end&gt;</c:calendar-data>` came back as `A &amp; B&#13;C &amp; D &lt;end&gt;`. Mixed content is legal and is exactly what a serialiser that CDATA-wraps only the values needing it produces. The instinct behind the early return was right — a section's content is already literal, and decoding it turns a summary of `Q&amp;A` into `Q&A` — so the two passes have to be *separated* rather than reordered: cut each section and copy its content verbatim, run the entity pass only on the text between sections. That is `decodeXmlText`. I also dropped the `i` flag, since `<![cdata[` is ordinary character data in XML and a case-insensitive match would delete it from a payload.

Things I checked and did not change, in case they come up: multiple sections in one element, a `]]>` in adjacent text, whitespace around the markers, an unterminated `<CDATA[`, and backtracking on a 2 MB body (4 ms) are all fine. `tagText` has two call sites — `calendar-data` and `href` — and CDATA is legal in both, so unwrapping there rather than at the call site is right.

Your test does measure the fault: reverting just the unwrap fails it with `0 !== 1`. I extended rather than replaced it — the fixture now carries a `getetag`, a `propstat/status` and the XML declaration a real response has, and its SUMMARY carries both a bare `&` and a literal `&amp;` so nothing can quietly reintroduce decoding inside the section, plus a direct `tagText` assertion for the mixed case. Please read that commit and say if you disagree with any of it.

Two conventions before this merges, both in [CONTRIBUTING.md — sorry, it landed after you opened this.

- Merges here are squashes, so the PR title becomes the line in `git log`: an imperative sentence in sentence case, no `fix(...)` prefix. Something like **Read a calendar object a server sent inside a CDATA section**.
- The body needs two headings that are read by a machine. `## Verification` names the commands you ran and what they said, and that you watched the new test fail against the unfixed code — and if you did see this against a live iCloud account, say exactly what you observed, because that is the one claim in this PR nobody else can reproduce. `## Release Notes` is lifted verbatim by `scripts/release-notes.sh` into the GitHub release, so without it this ships unmentioned. One bullet in the user's terms: *Calendars on servers that send events in a CDATA section — iCloud, DAViCal — now show their events instead of appearing empty.*

`make validate` green on the pushed commit, and it merges cleanly with `main` (nothing that has landed since touches `calendar/`).

PR #102, opened 2026-09-04
Let a mailbox be given a name

patrykbart:
Split out of #93 at your request. This half is local, fully tested and does not depend on the unified work — it is only what that branch was waiting on.

It stands on its own. Two of my three mailboxes differ only in their domain and elide to the same `patryk.bartkow…` in the switcher, which is unreadable whether or not the lists are ever combined.

## The field was already there

`label` has been part of an account entry and preferred by `Accounts.label` since accounts became a list. Nothing ever offered a way to set it. `setLabel` is that, and the switcher, the mailbox list in settings and the account file all read the same answer.

## A row leads with the name

That reverses a deliberate choice, and the reason it was right before is exactly why it changes now: the only name available was the local part, and two mailboxes easily share one across different domains. A name somebody typed is the opposite case — it exists precisely to tell them apart.

The address moves to the second line rather than being lost, and a reason the mailbox cannot be used still beats both, so a name can never hide "Signed out" or a server's refusal. The avatar's initial follows the name, because one taken from the address would not match the word above it.

## Where it is set

Settings → Mailboxes, as a picker and one field rather than a field on each row. The rows carry live state and are rebuilt whenever a poll moves an unread count, which would take the field apart while it was being typed into. The signature editor beside it is shaped this way for the same reason, and says so.

The field is empty when a mailbox has no name. Two prompts were tried there and both read as a value rather than an invitation — a sentence looked like a status line and an example looked like the name already chosen — so the line under the field carries "leave it empty to use the address" instead.

## One guard worth noting

The row's two lines are worked out on the row rather than inside the bindings that draw them: it is one choice with two halves, and splitting it across two `Text` elements hid that.

That also moves the address out of sight of `test_qml_text_format.py`, which reads `Text` bodies for untrusted references. A body saying `row.primaryText` no longer shows the `.email` the value is made of, so a `Text` that stopped declaring plain text would have passed. `primaryText` and `secondaryText` are in the check's untrusted list now, and I confirmed it fails when the declaration is removed.

## Verification

`make validate` green, 241 QML assertions.

`tests/test_accounts.js` covers trimming, that clearing puts the address back, that naming an absent mailbox changes nothing, and that it does not forget a signature. `tests/qml/tst_account_names.qml` covers the switcher: a named row leads with its name and keeps the address below, an unnamed one is unchanged, a signed-out or failing mailbox still says so instead, and the initial follows the name. `tests/qml/tst_settings_name.qml` types into the field the way a person does — a click, then keys — rather than assigning `text`, because the field draws its box whether or not anything can reach it.

Mutation-checked: ignoring the name fails two switcher tests.

## Release Notes

- A mailbox can be given a name in Settings → Mailboxes. The switcher and the mailbox list use it instead of the address, which matters most when several addresses differ only in their domain.
- The address stays on the row's second line, and a mailbox that is signed out or failing still says so rather than showing its name.


huacnlee:
Thanks

PR #103, opened 2026-09-04
File a copy of sent mail in the server's Sent folder

jquiros2:
An IMAP message handed to SMTP submission lands nowhere on its own, so the Sent folder stayed empty whatever the mailbox held. The send now follows with an APPEND of the same RFC 822 message to the folder the server named with SPECIAL-USE, arriving \Seen — a draft's own flag word travels with it now too, decided in Imap.js rather than hard-coded in the transport. A server that named no Sent folder holds no copy, and the status row says so rather than failing a send that has already answered.


jquiros2:
Reopening after testing against a live Stalwart IMAP server.

The initial "does not work" report was local, not the change: the shell had hot-reloaded the plugin while the already-running mail service kept the previous send path. After a full shell restart the round trip behaves as designed:

- the send goes out over SMTP, then the same message is APPENDed to the server's \`\\Sent\` folder (resolved via SPECIAL-USE, \`Sent Items\` here), flagged \`\\Seen\`
- journal: \`omamail: filing the sent copy in: Sent Items\`
- server-side check: exactly one new message in Sent Items, \`\\Seen\`, INTERNALDATE at append time, no duplicate copies — the server does not file SMTP submissions itself
- a server that names no Sent folder is skipped with the status note, and an append failure is a warning carried on the send's success, never a send failure

Also verified on this server: LIST responses reach the client through the transport in both the single-command and CAPABILITY+LIST forms, so folder resolution works as the tests assert.

PR #107, opened 2026-09-05
Queue a row action taken while another is still finishing

mfilm77:
Deleting several messages in a row only works as fast as the server answers.

## Reproduction

1. Open a mailbox on an IMAP account with a few unread messages.
2. Press `d` on the first row.
3. Before the request completes, move to the next message and press `d` again.

The second press does nothing. The status note reads "Another action is still
finishing", the message stays where it was, and the keyboard cursor does not
move on — `App.qml` holds it deliberately when `act` answers false, so a
refused action strands the cursor on a row the reader has already left.

Reproduced against a self-hosted IMAP mailbox where a round trip takes a few
hundred milliseconds, which is long enough that triaging a morning's mail hits
it on nearly every message. It is not a caching problem: the optimistic edit
and the body cache both behave. The refusal is in `act` itself.

## Why it happens

`act` allows one mutation at a time, which is right — `restore` puts a row back
at the index it held when the action was taken, so overlapping optimistic edits
would restore into a list that had moved underneath them.

But a pending slot is a reason to make the next action **wait**, not a reason to
drop it. `queueQuietAction` and `runQueuedQuietAction` already implement exactly
that waiting, and automatic mark-read already uses them. Only the explicit path
was refused.

## The change

The queue takes every action rather than only the quiet ones, and carries the
quietness it was taken with. That last part matters: `keepOpen` is
`quiet === true && selectedId === messageId`, so draining an explicit trash as
if it were quiet would leave the deleted message on screen under the reader
that deleted it. `runQueuedAction` now forwards `request.quiet` instead of a
hardcoded `true`.

`markAllRead` still refuses on purpose, and the test says why: it reads the
unread set at the moment it runs, so one queued behind a mutation would send a
list the mailbox had already moved past.

The names lost their `Quiet` now that the queue holds both kinds.

## Verification

`make validate` — `test-js`, `test-shell-portable`, `test-shell-libcurl`,
`test-qml` and `qml-check` all pass. QML: 269 passed, 0 failed. `qmllint`
reports 2146 warnings both before and after the change, so this adds none.
`git diff --check` is clean.

Four assertions were added to `tests/test_source.sh`, the file that already
guards this area:

- an action taken while one is pending must queue rather than be refused
- a queued action must not report a failure the user did not cause (scoped to
  `act`, so `markAllRead` keeps its deliberate refusal)
- a queued action must run with the quietness it was taken with
- the existing mark-read and detached-action assertions, updated for the names

Each was checked to fail without the fix. Reverting only the guard, leaving the
renames and the call text identical, fails on
"a queued action must not report a failure the user did not cause" — so the new
assertions catch the behaviour rather than the rename.

Manually: pressing `d` down a list now clears every message pressed, in order,
with the cursor moving on at each press. No note appears. Held `d` dedupes on
`(id, action)` as before.

## Release Notes

- Deleting, archiving or starring several messages quickly no longer drops the
  presses that land while the previous one is still finishing. They queue and
  run in order, and the cursor moves on with each press instead of stopping on
  a row you have already left.

### DIFF
--- a/account/MailAccount.qml
+++ b/account/MailAccount.qml
@@ -320,15 +320,15 @@ Item {
   property string pendingAction: ""
   property string pendingActionQuery: ""
   property var deferredListLoad: null
-  property var queuedQuietActions: []
+  property var queuedActions: []
   property bool sending: false
   property var pendingSend: null
   property int sendSecondsRemaining: 0
   readonly property bool sendPending: pendingSend !== null
 
   onPendingActionChanged: {
-    if (pendingAction === "" && queuedQuietActions.length > 0)
-      Qt.callLater(root.runQueuedQuietAction)
+    if (pendingAction === "" && queuedActions.length > 0)
+      Qt.callLater(root.runQueuedAction)
   }
 
   // Notifications only start once the first successful load has established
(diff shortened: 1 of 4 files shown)


huacnlee (approved):
Thanks for the contribution. The queue is useful for rapid triage. I found and fixed the remaining queue semantics directly in 029c902:

- Queued moves retain the original source label after navigation, including switching from a label view to a search with the same cache key.
- Deduplication no longer drops the last action in read/unread/read. Consecutive duplicate intent for a message still coalesces.
- An explicit action takes precedence over a duplicate automatic mark-read, so the explicit operation can evict the row as intended.

Added runtime QML regressions using the real MailAccount and its existing provider Loader with synthetic callbacks. They cover original-label moves, same-query navigation, final intent, repeated trash, quiet/explicit precedence, and failure rollback. The three original regressions failed before the fix; the additional same-query regression also failed before its guard was added. All 8 queue test results now pass (including setup/cleanup).

make validate passes on the final changes: 277 QML tests, zero failures, plus the JavaScript, shell, lint and plugin validation checks. A fresh independent review confirmed no remaining actionable blocker.

Security verdict: PASS for this scoped queue change. Captured message IDs and source labels still flow through the existing provider interfaces; no new credential, network-origin, shell or HTML boundary is introduced. Tests use synthetic provider callbacks, not live mailbox mutations. This is not a complete transport audit.

Approving the updated implementation.

PR #108, opened 2026-09-05
fix: saving one IMAP mailbox must not delete another

gw7523:
## Reproduction

1. Add iCloud and Proton as two IMAP mailboxes.
2. Leave iCloud selected.
3. Open settings and re-auth Proton (or save Proton’s address/servers on the form).
4. iCloud disappears from the account list. Its keyring entry is still present.

`configureAccount` rebuilt the list through `Accounts.add`. A colliding id is treated as “replace that other row”, so the edited slot is gone.

## Fix

- `replaceAt` writes only the edited index.
- A colliding id is refused (`duplicateAccount`) instead of merging.
- `saveAccounts` also refuses a payload that drops an id last written to disk, unless the caller is an explicit remove.

Test: `node tests/test_accounts.js` and `bash tests/test_service_source.sh`.


Service.qml
```diff
@@ -384,11 +492,14 @@ Item {
     if (raw.imap !== undefined) entry.imap = raw.imap
     if (raw.label !== undefined) entry.label = raw.label
 
-    var updated = Accounts.emptyList()
-    updated.activeId = accountList.activeId
-    for (var i = 0; i < accounts.length; i++)
-      updated = Accounts.add(updated, i === index ? entry : accounts[i])
+    // Never rebuild through `add`: a colliding id replaces the *other* row
+    // and drops this one, which is how re-authing Proton deleted iCloud.
+    if (Accounts.collidingId(accountList, index, entry)) {
```
huacnlee, line 634: [P1] Stop sign-in when duplicate configuration is refused

The new collision guard returns without changing the account, but configureCurrentAccountAndSignIn still unconditionally schedules signInWithPassword. Editing A to the already-added address/server B and entering B’s password therefore authenticates the unchanged A server with B’s secret. A Qt 6.11.2 test through the real Service, MailAccount, ImapAuth and ImapClient recorded a transport request for imaps://a.example.org:993 carrying a:synthetic-B-secret after the duplicate refusal. Processes were stubbed; no real secret/network request was used. Return an explicit failure through the configuration calls and do not schedule authentication unless the requested configuration was accepted.

App.qml
```diff
@@ -872,9 +1069,13 @@ Item {
       root.resumeHeldDraft()
     }) }
 
+    function onSidebarWidthChanged() { root.sidebarWidth = root.service.sidebarWidth }
+    function onListWidthChanged() { root.listWidth = root.service.listWidth }
     function onMessagesChanged() {
       root.cursorId = Model.cursorAfterReload(
         root.service ? root.service.messages : [], root.cursorId)
+      root.checkedIds = Model.retainIds(root.checkedIds,
```
huacnlee, line 1387: [P1] Clear or account-scope bulk selection on account changes

checkedIds stores bare message IDs and this handler retains any matching ID in the newly selected account. Two IMAP inboxes can both contain 42:INBOX, so selecting that row in A then switching to a warmed B inbox leaves B’s unrelated message checked. A Qt 6.11.2 test on this head switched real Service accounts and recorded B’s provider receiving the bulk trash request from A’s selection. Reset selection on every account switch or attach account ownership and reject a mismatch before dispatch. The same issue previously reported in #109 is present here.

Service.qml
```diff
@@ -443,11 +573,16 @@ Item {
     // row. Removal never arrives here as an omission: it goes through
     // `remove` or `removeAt`, so the row is gone from `accountList` too.
     if (Accounts.dropsNamedMailbox(accountList, writable)) return
+    var allowDrop = !!(opts && opts.allowDrop)
+    if (!allowDrop && Accounts.dropsAnyId(lastPersistedIds, writable)) return
```
huacnlee, line 732: [P2] Allow an explicit address correction to replace its persisted ID

Changing an existing IMAP account’s email legitimately changes its account ID. configureAccount accepts that replacement in memory, but this check treats the missing old ID as an unauthorized removal and silently skips every save. In a real Service QML test, correcting a@example.org to corrected@example.org updated accountList while accountsWritePayload remained empty. The correction disappears on reload, and subsequent ordinary settings saves remain blocked by the old snapshot. Distinguish an authorized replacement of the edited row from accidental omission of another account.

account/MailAccount.qml
```diff
@@ -1575,27 +1651,58 @@ Item {
...
+    pendingAction = action
+
+    var done = function(payload, error) {
       root.pendingAction = ""
       root.pendingActionQuery = ""
       if (error) {
         if (root.cacheKey === actionQuery
             && !root.deferredLoadCleared(actionQuery)) {
           root.nextPageToken = actionToken
           root.messages = before
```
huacnlee, line 1986: [P2] Reconcile successful members of a partially failed bulk action

Trash/untrash dispatch separately for each ID and report only the first error. If one succeeds and another fails, this branch restores the whole pre-action list and cache, including the successfully moved message, without a reload when no list load was interrupted. A Qt 6.11.2 test on this head reproduced the successful row returning after the other request failed; the function-level probe recorded zero reloads. Track per-message outcomes or revalidate the list/cache on partial failure instead of presenting every operation as rolled back.

providers/ImapProtocol.js
```diff
@@ -1035,6 +1070,64 @@ function decodeMailbox(name) {
...
+
+// The three commands that change the folder list. Names travel encoded and
+// quoted; a RENAME carries every folder beneath the old name with it, which
+// is what moving a folder under another parent is.
+function createCommand(name) {
+  return "CREATE " + quote(encodeMailbox(name))
+}
+
+function renameCommand(from, to) {
+  return "RENAME " + quote(encodeMailbox(from)) + " " + quote(encodeMailbox(to))
```
huacnlee: [P2] Keep existing IMAP folder IDs in their wire encoding

getLabels exposes the server’s modified-UTF-7 name as id/rawName, and renameLabel/deleteLabel pass that ID here. Encoding it again changes which folder is addressed: the actual helpers turn 旅行 → &ZcWITA- into DELETE "&-ZcWITA-" or RENAME "&-ZcWITA-" "Trips". These normally fail, and can affect a different literal-named folder if it exists. Creating/moving beneath an encoded parent has the same problem. Preserve existing wire IDs and encode only new Unicode names, with non-ASCII parent/rename/delete regressions.

account/Model.js
```diff
@@ -992,3 +1018,360 @@ function settingsScrollTarget(sections, key, contentHeight, viewportHeight) {
...
+    var full = String(label.name || label.rawName || "")
+    if (full === "") continue
+    var parts = full.split(delimiter)
+    var node = root
+    var path = ""
+    for (var p = 0; p < parts.length; p++) {
+      var part = parts[p]
+      if (part === "" && p > 0) continue
+      path = path === "" ? part : path + delimiter + part
+      if (!node.children[part]) {
```
huacnlee, line 1565: [P2] Do not resolve label names through Object.prototype

children is a plain object, so valid names such as constructor, toString and __proto__ resolve inherited properties and skip node creation. Running the real labelTree with each standalone name produced no visible row; constructor/child threw a TypeError and can break the sidebar’s entire label-tree binding. Use a dictionary representation that safely stores these names and check own entries, including the __proto__ assignment case.

agent/Agent.js
```diff
@@ -0,0 +1,420 @@
...
+// a line that is not JSON both mean "no jobs", not an error worth a notice.
+function parseJobs(text) {
+  var parsed = null
+  try { parsed = JSON.parse(String(text || "")) } catch (e) { parsed = null }
+  if (!Array.isArray(parsed)) return []
+  var out = []
+  for (var i = 0; i < parsed.length; i++) {
+    var job = parsed[i]
+    if (!job || typeof job !== "object") continue
+    if (String(job.id || "") === "" || String(job.messageId || "") === "") continue
```
huacnlee: [P1] Retain active scope and multi-message jobs

Scope jobs and selectionPayload jobs deliberately have an empty messageId; they instead carry scope or messageIds. This filter drops both kinds from every runner listing. Real parseJobs probes returned [] for each shape, so those jobs continue running while disappearing from the pane, no longer keeping polling active, and becoming unavailable to the UI cancellation controls. Validate the supported job shapes rather than requiring a single-message ID for every job.

scripts/agent-job.py
```diff
@@ -0,0 +1,487 @@
...
+    messages = payload.get("messages")
+    messages = [m for m in messages if isinstance(m, dict)] if isinstance(messages, list) else []
+    parent = None
+    if parent_id:
+        parent_dir = os.path.join(state_dir(), parent_id)
+        if not os.path.isfile(job_path(parent_dir)):
+            sys.stderr.write("agent-job.py new: no such parent job\n")
+            return 2
+        parent = read_job(parent_dir)
+        scope = scope or clean_text(parent.get("scope")).strip()
```
huacnlee, line 145: [P2] Preserve the complete target context for follow-up jobs

continuationPayload provides the parent ID and answer but leaves account/folder/selection empty. This inheritance copies only scope and messageId. A follow-up to a multi-message job is consequently rejected by the validation below because it has neither messages, messageId nor scope; single-message/account-scope follow-ups proceed with blank account targeting in the generated prompt. Inherit the parent’s account, folder, account list and selected messages as appropriate, and test continuations for all supported scopes instead of only a single-message job.

huacnlee (changes requested):
Reviewed head 6fff07c4a4c76f70569330e97073da867ef3b458. Requesting changes for the inline findings, particularly the wrong-account credential and bulk-action dispatches.

The current diff is substantially broader than the title/body: it also introduces agent execution, bulk selection, label management/watching, pane controls and OAuth changes. Please separate the account-save fix or update the title and description to cover the final scope. GitHub currently reports conflicts with main.

Security verdict: BLOCK. Refusing a duplicate account configuration still dispatches the newly entered secret to the unchanged account’s server, and bulk selection can send a destructive action to a different account. Both routing violations were reproduced with real QML objects and synthetic/stubbed provider transports; no real mailbox or secret was used.

Validation: the existing make validate passes, including 269 QML results, shell/JS checks and plugin validation. Four additional Qt 6.11.2 regressions fail for address persistence, duplicate-refusal authentication, cross-account bulk selection and partial bulk rollback. Protocol/model/agent probes exposed the remaining encoding, label-tree and job-state issues. Scoped transport checks also rejected twenty control-character token cases before connection, and real curl 8.21.0 successfully exercised synthetic loopback XOAUTH2. These results do not establish live-provider compatibility, model prompt-injection resistance, or production systemd cancellation behavior.

agent/Agent.js
```diff
@@ -0,0 +1,499 @@
...
+
+// The job a message shows: the one still running if there is one, else the
+// newest. A message can have been asked about twice; the row has one glyph.
+function jobFor(jobs, messageId) {
+  var list = Array.isArray(jobs) ? jobs : []
+  var id = String(messageId || "")
+  if (id === "") return null
+  var newest = null
+  for (var i = 0; i < list.length; i++) {
+    if (messageIdsOf(list[i]).indexOf(id) < 0) continue
```
huacnlee, line 48: [P1] Include account ownership in Agent job lookup and message actions

`jobFor` matches only `messageId`, and `jobsByMessage` likewise indexes bare IDs. Two IMAP accounts can both have `42:INBOX`. A runtime probe seeds a running `synthetic-A-job` for A, switches the real Service to B, then calls `cancelAgent("42:INBOX")`: it returns true and the Process substitute records `["python3", "/tmp/omamail-test/scripts/agent-job.py", "cancel", "synthetic-A-job"]`. B's message action is cancelling A's job. The same unscoped map also supplies the message popup's job/question and continuation target.

Carry a stable, provider-qualified account identity through persisted jobs, message indexes, attention state and action lookup, and verify ownership before dispatching a message-scoped cancellation or continuation. Filtering only what the pane draws is insufficient. Add colliding-ID regressions for display, cancellation and continuation; email alone cannot distinguish two providers using the same address.

account/BatchAction.qml
```diff
@@ -0,0 +1,262 @@
...
+            && !account.deferredLoadCleared(actionQuery)) {
+          account.messages = Model.restoreRows(account.messages, before, failedIds)
+          account.previewMessages = Model.restoreRows(account.previewMessages, beforePreview, failedIds)
+          restoreMembersOf(failedIds)
+          if (!selectedGone && beforeSelected && beforeSelected.id === account.selectedId)
+            account.selectedMessage = beforeSelected
+          if (!interrupted) account.rememberList()
+          account.refreshCounts()
+          account.fail(Model.batchFailureNote(listed.length, failedIds.length, account.actionLabel(action), error))
+          return
```
huacnlee, line 192: [P2] Complete list-load cleanup after restoring partially refused rows

This return bypasses both pagination revalidation and `resumeDeferredListLoad`. With `nextPageToken = "older-page"`, trash two rows and have one succeed and one fail: the optimistic update clears the token, the refused row returns here, but the token stays empty and no list request follows. The real-QML probe records `hasMore === false` and a list-call delta of 0, leaving no way to load older messages without manually refreshing.

A second probe calls `loadMessages(false, true, "")` while that batch is pending. The refresh enters `deferredListLoad`, but is never resumed after the partial failure. Keep the per-row restoration while completing deferred/interrupted loads and revalidating invalidated pagination; simply restoring an old page token can be wrong after successful deletions. Both cases reproduce independently on this head and on #109.

huacnlee (changes requested):
Re-reviewed `c213d9cbc1aec0aedef5864351cb4b0949f5608b`. The duplicate-account refusal now stops sign-in: a focused QML regression returns false and confirms the unchanged account does not receive the synthetic password entered for the refused account. The existing cross-account batch-selection, per-row restoration, and aggregate-batch revalidation tests also pass.

Two findings remain, detailed inline: Agent job lookup/cancellation still crosses account ownership when IMAP IDs collide, and partial batch failure skips pagination/deferred-refresh cleanup.

The scope mismatch also remains: this PR changes 83 files, including Agent execution, bulk actions, signatures, Graph/OAuth and a send queue, while its title/body describe only account persistence. Please use the separate PRs for those changes, or bring the title, description and Release Notes into agreement with the full diff.

Validation: `QT_QPA_PLATFORMTHEME= make validate` exits 0 (454 QML results plus 4 Outlook HTTP results). Additional focused review probes report 6 passed / 3 failed, including setup/cleanup: wrong-account Agent cancellation, lost pagination and an unresumed refresh. Runtime: Qt/QtTest 6.11.2, Omarchy 4.0.2. The probes use real QML objects and controlled provider/Process substitutes, with synthetic accounts and credentials.

Security verdict: BLOCK. The Agent regression demonstrates an incorrect account-to-job command dispatch. It records the cancellation command without executing systemctl or touching any real mailbox; it is not evidence of actual mail deletion or credential exfiltration. The wider Agent permission/prompt-injection boundary, production systemd lifecycle and new Graph/OAuth paths were not fully verified in this follow-up and must not be treated as cleared by the general test suite.

gw7523:
Closing this one: it grew past its title. The account-save fix is #116 on its own, and the rest of what this branch carried is the reviewed PRs #109, #112, #113, #114 and #115, where the findings from this thread (agent job ownership, partial-batch cleanup, folder wire names, label tree names, continuation context) are being addressed.

PR #109, opened 2026-09-05
Name the open mailbox and act on selected messages

gw7523:
## What changes

**The header says what the window is looking at, and switches it.** The "Omamail" title becomes a scope line: the account, then the mailbox or label open in it — `iCloud › Inbox`, `Work › Receipts`. Each half is a control. The account half opens the existing account switcher; the mailbox half opens a new `MailboxSwitcher` popup over the rail's own numbered slots, also bound to `Alt+M`. Inside it `j`/`k`/`Enter`/`o` walk the rows and a bare digit is the Ctrl digit from the list. This is the answer for a window whose rail is collapsed to icons or folded into tabs, where nothing on screen said which account or folder was open.

**Several messages can be selected and acted on with the keys they already have.** A row gains a tick in the lane the unread dot lives in: under the pointer or cursor, and on every row while any is ticked, so the text never moves. `x` ticks the cursor row, `Ctrl+A` ticks every loaded row or none, Ctrl+click and Shift+click do the same with the mouse, and Escape unticks. While rows are ticked — the list alone, or beside the reader in a wide window — `e`, `d`, `s`, `v`, `Shift+I` and `Shift+U` act on all of them; a row's own buttons, the reader's archive, trash and star buttons and the row menu mean the selection on a ticked row and that row alone on any other. The status line counts the selection; the list's hints and the reader's blank slate name `x`.

## How

- `Model.currentScope` names what is open from the same facts the rail's selected row is drawn from: the mailbox key, or the label matched by the id it was selected with and then by the provider's own query rule. `Model.switcherRows` turns `sidebarSlots` into the popup's rows with the open one marked — the same list as the badges and the Ctrl digits.
- The selection is a list of ids held in `App.qml` beside the cursor, pruned by the same `onMessagesChanged` path when the list under it changes. Every acting key goes through the existing `actOnCursor`, which routes to `actOnChecked` when rows are ticked, so the provider guard in `MailAccount.act` still refuses what the mailbox cannot do before any optimistic update. Star over a mixed selection stars everything; only an all-starred selection unstars.
- `MailAccount.actMany(ids, action)` is the batch: one `batchModify` where the client takes a list, one `trashMessage` call per message where it does not, with the same optimistic update, restore-on-failure, live-list interruption and cache repair as `act`. `markAllRead` now goes through it, and the source regression that guarded mark-all follows it there.
- Two fixes that the scope line surfaced: a row hovers through a passive `HoverHandler`, so a button's own MouseArea no longer takes the hover and flickers the action lane; and every popup row takes the exclusive grab on press (`TapHandler.ReleaseWithinBounds`), because the app menu now drops over the message list and a passive grab let the row underneath receive the release.

## Verification

Reviewer validation on `a502077e74ed04c5656645f8a9be7144cced836d`: `QT_QPA_PLATFORMTHEME= make validate` passed, including JavaScript/shell, QML, native HTTP and sidebar text checks, qmllint and plugin validation. The reader star button now applies to the visible checked selection when it contains the open message. Five added runtime cases cover star/unstar, an open message outside the selection, compact mode and no selection; the two selection cases failed before the fix. The focused suite passed 13 cases. A fresh independent agent reviewed the repair and found no defects; Security: PASS for the affected scope.

Manual light/dark and wide/compact on-screen validation remains outstanding, so this revision is not yet approved.

The author previously reported live IMAP exercises on Office 365 via XOAUTH2 and iCloud, plus a Gmail-shaped label set: header scope, mailbox switching, selection keys and modifiers, and two-row read/unread actions. HEY was not exercised live. These are author-reported checks, not new reviewer live-mailbox tests.

## Release Notes

- The header now names the open account and mailbox, and each half opens a switcher; `Alt+M` opens the mailbox list from the keyboard.
- Select several messages with `x`, `Ctrl+A`, Ctrl+click or Shift+click, then archive, trash, star, move or mark them read or unread together with the usual keys and buttons.

### DIFF
--- a/App.qml
+++ b/App.qml
@@ -1987,12 +1987,17 @@ Item {
             // from a member the action lands on the row the conversation was
             // opened from, which is where the reader came from. Which members
             // an action reaches from there is "Actions on a conversation row".
-            // The toolbar emits archive and trash and nothing else; the star
-            // is a button of its own, through `toggleStar`, which acts on the
-            // open message. Unless the open message is one of the ticked rows,
-            // in which case the buttons mean all of them, by the same rule as
-            // a row's buttons.
             var outside = root.checkedIds.indexOf(root.service.selectedId) < 0
+            // A star stays on the open member unless it belongs to the visible
+            // selection. It never moves the list's independent cursor.
+            if (action === "star") {
+              if (root.selectionActive && !outside)
+                root.actOnChecked(Model.starActionFor(Model.summariesById(
+                    root.service.messages, root.checkedIds)))
+              else
+                root.service.toggleStar(root.service.selectedId)
+              return
+            }
             if (!Conversation.holdsMember(root.service.selectedThread,
                 root.service.selectedId) || root.cursorId === "")
               root.cursorId = root.service.selectedId
--- a/App.qml
+++ b/App.qml
@@ -12,6 +12,7 @@ import "account/Accounts.js" as Accounts
 import "account/Navigation.js" as Nav
 import "compose/Recovery.js" as Recovery
 import "keys/Keymap.js" as Keymap
+import "providers/Registry.js" as Provider
 import "message/Mailto.js" as Mailto
 import "message/Message.js" as Message
 import "components"
@@ -206,6 +207,100 @@ Item {
   readonly property bool compact: window.width < Style.space(760)
 
   property string cursorId: ""
+  // The rows ticked for a bulk action, by id. Like the cursor, a fact about
+  // this window rather than about the mailbox, and pruned the same way when
+  // the list under it changes. Only the list acts on it: in the reader there
+  // is one message and it is the one open.
+  property var checkedIds: []
+  // The mailbox the ticks belong to. An IMAP id is a UID and a folder, unique
+  // only inside one account, so a tick carried across an account switch
+  // would name whatever the other mailbox keeps under the same id — and a
+  // batch dispatched through it would act on that. The ticks are bound to
+  // the account they were made in and dropped the moment it changes, from
+  // any path, before a batch can run.
+  property string checkedAccountId: ""
+  // Ticked rows mean the selection whenever the list is on screen: alone, or
+  // beside the reader in a wide window. A plain click opens a message and
+  // puts the window in the reader view, so a click, a Shift+click and `d`
+  // would otherwise trash only the open message with the ticks ignored.
+  readonly property bool listOnScreen: currentView === "list"
+    || (currentView === "reader" && !compact)
+  readonly property bool selectionActive: checkedIds.length > 0 && listOnScreen
+
+  function toggleCheck(id) {
+    var key = String(id || "")
+    if (key === "") return false
+    claimChecks()
+    checkedIds = Model.toggleId(checkedIds, key)
+    return true
+  }
+
+  // Ticks made in another mailbox are not this one's: start over.
+  function claimChecks() {
+    var owner = service ? String(service.activeAccountId || "") : ""
+    if (checkedAccountId !== owner) checkedIds = []
+    checkedAccountId = owner
+  }
+
+  function clearChecksIfForeign() {
+    var owner = service ? String(service.activeAccountId || "") : ""
+    if (checkedIds.length > 0 && checkedAccountId !== owner) checkedIds = []
+    checkedAccountId = owner
+  }
+
+  // Shift+click: everything from the cursor to here joins the selection, and
+  // the cursor moves here so the next stretch starts where this one ended.
+  function checkRange(id) {
+    if (!service) return false
+    var key = String(id || "")
+    if (key === "") return false
+    claimChecks()
+    checkedIds = Model.unionIds(checkedIds, Model.idsBetween(service.messages, cursorId, key))
+    cursorId = key
+    return true
+  }
+
+  // Every loaded row, or none when every one is already ticked: one key that
+  // means "all of these" has to be able to mean "none of them" as well.
+  function checkAll() {
+    if (!service) return false
+    claimChecks()
+    var all = Model.allIds(service.messages)
+    checkedIds = Model.retainIds(checkedIds, service.messages).length === all.length ? [] : all
+    return true
+  }
+
+  // Acting on the selection acts on every ticked row at once, then drops the
+  // selection: the rows it named have moved or changed, and a selection that
+  // outlived the action would be one keystroke from repeating it.
+  function actOnChecked(action) {
+    if (!service || checkedIds.length === 0) return false
+    // Never through another mailbox: a switch that reached the service by
+    // any road drops the ticks before a batch can be built from them.
+    if (checkedAccountId !== String(service.activeAccountId || "")) {
+      checkedIds = []
+      return false
+    }
+    var ids = checkedIds.slice()
+    var leaves = !Model.survivesAction(service.mailboxKey, action,
+      service.rawQuery, service.hasLabels, service.rawLabelId)
+    var next = leaves ? Model.cursorAfterRemovals(service.messages, ids, cursorId) : cursorId
+    // The open message going with the selection closes the reader, the way
+    // acting on it alone does: it is about to leave this list.
+    var wasOpen = currentView === "reader" && ids.indexOf(service.selectedId) >= 0
+    if (!service.actMany(ids, action)) return false
+    checkedIds = []
+    if (!leaves) return true
+    if (wasOpen) {
+      if (next !== "") openMessage(next)
+      else backToList()
+      return true
+    }
+    cursorId = next
+    revealCursorRow()
+    return true
+  }
+
   // Kept across messages, and across the window being closed: how somebody
   // reads their mail is a fact about them, not about the message that made them
   // reach for it. The service holds it because that is what writes it to disk.
@@ -852,7 +947,7 @@ Item {
   // move before asking for a destination, through the same provider guard that
   // checks the final action before its optimistic update.
   function openLabelPicker() {
-    if (!service || cursorId === "") return false
+    if (!service || (cursorId === "" && !selectionActive)) return false
     // A merged list draws no labels, so there is nothing to offer and the
     // picker would open empty on a destination list it cannot fill — and a
     // chosen id would belong to whichever mailbox happened to be active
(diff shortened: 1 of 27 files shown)


App.qml
```diff
@@ -875,6 +957,8 @@ Item {
     function onMessagesChanged() {
       root.cursorId = Model.cursorAfterReload(
         root.service ? root.service.messages : [], root.cursorId)
+      root.checkedIds = Model.retainIds(root.checkedIds,
```
huacnlee, line 1261: [P1] Reset or scope checked IDs when the account changes

Pruning by bare message ID cannot distinguish accounts. Reproduction with real Service/App switching: A and B each contain `42:INBOX`; check A's row, switch to B, then invoke the batch trash action. The check survives and the controlled provider records a trash request for B's unrelated row. IMAP UID/folder IDs are only unique inside an account. Bind the selection to its owning account and clear it on account/host changes before a batch can execute, including changes initiated outside App.switchAccount. Add a regression with colliding IDs across two accounts that asserts no B mutation occurs from A's selection.

account/MailAccount.qml
```diff
@@ -1575,27 +1601,58 @@ Item {
...
+    pendingAction = action
+
+    var done = function(payload, error) {
       root.pendingAction = ""
       root.pendingActionQuery = ""
       if (error) {
         if (root.cacheKey === actionQuery
             && !root.deferredLoadCleared(actionQuery)) {
           root.nextPageToken = actionToken
           root.messages = before
```
huacnlee, line 1978: [P2] Reconcile partial batch success instead of restoring every row

Trash/untrash launches one request per ID but reduces all results to a single firstError. If message one succeeds and message two fails, this branch restores the entire pre-action list, including message one, and normally returns without a fresh list load because interrupted is false. The offscreen regression confirms the successfully trashed message reappears. Track per-message outcomes and restore only failed rows, or invalidate/reload authoritatively when the provider cannot report an atomic outcome; keep cache, preview state and counts consistent as well. Cover mixed-success trash/untrash and non-atomic provider batches rather than treating an aggregate error as proof that no mutation succeeded.

huacnlee (changes requested):
The header/switcher and bulk-selection features are useful, but the current head is not ready to merge. I independently ran make validate successfully (269 QML tests, Qt 6.11.2) and then added two focused offscreen probes outside the checkout. Both fail on the actual App/Service/MailAccount behavior.

The blocking issues are account ownership of checked IDs and partial-success reconciliation, detailed inline. Please add runtime regressions for these paths; the current model/source assertions cannot detect them.

Security verdict: BLOCK. The account-switch regression demonstrates selection from A being used to dispatch a trash request through B's provider when both accounts contain the same IMAP UID/folder ID. The fixture uses real App and Service account switching, with only the provider replaced through its existing Loader; it records B receiving the mutation. No live mailbox or real credential was used, so this is a demonstrated account-ownership boundary violation, not a claim that real mail was deleted during review. Scope/folder text is explicitly plain text, and this patch introduces no new transport, but those facts do not compensate for the wrong-account mutation.

Separately, the partial-trash probe records one successful callback and one failure, then observes both rows restored. This contradicts the PR description's stated behavior that successful rows are not restored. General validation remains green because there are no new QML interaction/batch lifecycle tests in the PR.

gw7523:
Thank you for the probes — both were right, and both are fixed in `49493a7`, with runtime regressions that drive the real App, Service and MailAccount against a controlled client.

**P1, account ownership of checked ids.** The selection now belongs to the account it was made in (`checkedAccountId`). Any switch that reaches the service — through `App.switchAccount`, `Service.switchToIndex`, or the account list changing — drops the ticks in `onActiveAccountIdChanged`, and `actOnChecked` refuses to build a batch when the owner differs, so the batch cannot be dispatched through the other account's provider. The cursor is an id too and is dropped with them, so the next `d` after a switch acts on nothing rather than on whatever shares the id. `tst_batch_actions.qml::test_ticks_do_not_cross_an_account_switch` seeds two IMAP accounts that both hold `42:INBOX`, ticks it in A, switches to B both ways, and asserts B's client records no trash request.

**P2, partial-success reconciliation.** Trash and untrash now answer per id: only the rows whose request failed are put back, each at its old place (`Model.restoreRows`), the accepted rows stay gone, counts are refreshed and the note says "1 of 2 could not be moved to trash: …" (`Model.batchFailureNote`). A whole-batch refusal from a client that answers with one word (`batchModify`, and IMAP's plan across folders) is no longer treated as proof that nothing changed: the list is read again from the server when it is on screen, and the cache is repaired when it is not. `test_a_partly_refused_trash_restores_only_the_refused_rows` and `test_a_refused_batch_reads_the_list_again` cover both.

The controlled client is reached through a small `clientOverride` seam on `MailAccount` (a `Component` the Loader prefers when set), which is what the harness needed to replace the provider without a server or a credential. The PR description is updated. `make validate` passes here (274 QML tests).

gw7523:
Rebased onto `d5a5d03` as one commit, over the JMAP conversations and the action line from #107.

- The batch now lives in `account/BatchAction.qml`, a second caller of the account's optimistic and restore machinery: `MailAccount.qml` sits within a few hundred bytes of the new 128 KiB ceiling, so `actMany` forwards to it. It follows the conversation model: each ticked row expands to its counted members through `Model.actionTargets`, a conversation-scoped verb asserts the row's block with `threadAfterAction` and a message-scoped one recomputes it from the members, the rail's member summaries take the change and go back on a refusal, and `survivesAction` is asked per row. `trashMessage` and `untrashMessage` are called once per row with that row's member list, so a refusal is still answered per row.
- A batch asked for while another action finishes joins the action line one row at a time rather than being refused.
- `markAllRead` stays as `main` has it.
- `make test-js`, `make test-shell` and `make test-qml` pass on the branch.

gw7523:
Rebased onto `062f0f0`, over the All Mailboxes view and preview-on-cursor.

- In a merged list a batch is refused the way a move is ("Acting on several messages needs one mailbox on screen"): the ticked rows can come from several mailboxes and a batch is one mailbox's request. The label-move guard keeps `main`'s merged-list refusal and its `refuseUnavailableAction(action, id)` shape, with the selection standing in for the cursor.
- `main`'s `MailAccount.qml` is now 109 bytes under the 128 KiB ceiling, so this branch moves the RSVP block beside it as `account/Rsvp.qml` (its state stays on the account, where the reader reads it) to make room for the batch forwarder and the test seam. If you would rather raise the ceiling than split the file, say so and I will drop that part.
- `make test-js`, `make test-shell` and `make test-qml` pass on the branch.

account/BatchAction.qml
```diff
@@ -0,0 +1,262 @@
...
+            && !account.deferredLoadCleared(actionQuery)) {
+          account.messages = Model.restoreRows(account.messages, before, failedIds)
+          account.previewMessages = Model.restoreRows(account.previewMessages, beforePreview, failedIds)
+          restoreMembersOf(failedIds)
+          if (!selectedGone && beforeSelected && beforeSelected.id === account.selectedId)
+            account.selectedMessage = beforeSelected
+          if (!interrupted) account.rememberList()
+          account.refreshCounts()
+          account.fail(Model.batchFailureNote(listed.length, failedIds.length, account.actionLabel(action), error))
+          return
```
huacnlee, line 221: [P2] Finish pagination and deferred-refresh cleanup after partial failure

This early return skips the list-load cleanup performed by the other completion paths. With `nextPageToken = "older-page"`, trash two rows and let one request succeed and the other fail: the optimistic update clears the token, the refused row is restored here, but the token stays empty and no `listMessages` call follows. The runtime probe records `hasMore === false` and a list-call delta of 0, so the user loses the ability to load older messages.

A second probe starts the same batch and calls `loadMessages(false, true, "")` before the callbacks arrive. That refresh is stored in `deferredListLoad`; this branch returns without `resumeDeferredListLoad`, again producing no list request after the action finishes.

Keep the per-row restoration, but also resume deferred/interrupted loads and revalidate an invalidated page before returning. Simply restoring the old token may be incorrect after successful deletions. Please add runtime regressions for both paths; the current partial-trash test checks the restored rows but not pagination or a waiting refresh.

huacnlee (changes requested):
Re-reviewed head `ca44baff8cfdfb33366ca77d75b400b02687398b`. The original cross-account selection regression now passes, as do the tests for restoring only refused rows and re-reading after an aggregate batch refusal. Switching accounts through either App or Service clears the ticks, and `actOnChecked` refuses a deliberately retained selection owned by another account.

One P2 remains in the partial-failure completion path, detailed inline: it restores the refused rows but leaves pagination invalidated and skips a deferred refresh. I reproduced both cases independently on this PR's head with the real App, Service and MailAccount and a controlled provider. Please complete that list-load cleanup before merging.

Validation: `QT_QPA_PLATFORMTHEME= make validate` exits 0 (450 QML results plus 4 Outlook HTTP results, with no failures). The additional focused review harness reports 5 passed / 2 failed, including setup/cleanup; the two failures are the pagination and deferred-refresh cases below. Runtime: Qt/QtTest 6.11.2, Omarchy 4.0.2. No live mailbox or real credential was used.

Security verdict: PASS for the account-ownership and text-rendering boundaries examined in this change. The account-switch test records no trash dispatch to B from A's selection; the merged-mailbox entry point refuses batches; the new account/label text is explicitly plain text; the RSVP extraction retains the existing MIME construction and account provider path. This is a scoped verdict, not a complete application security audit or evidence of live-provider compatibility. The remaining functional defect still warrants changes.

gw7523:
Rebased on current main (`85eb6c2`, so the A-to-Z rail from #128 is underneath) and the remaining P2 addressed.

**Partial-failure cleanup** (`account/BatchAction.qml`): the branch that restores only the refused rows no longer returns early. After the rows are back it resumes a refresh that was deferred behind the action (`resumeDeferredListLoad`) and, when the optimistic update invalidated the page, reads the page again from the server (`loadMessages(false, true, note)`) instead of putting the old token back — the accepted rows are gone, so the old token names a page that no longer starts where it did. The failure note survives the reload.

The same holds when the view moved on while the batch ran — to another query, or cleared for a reload of this one: the pre-action page no longer comes back on screen or into the cache (the accepted rows are gone from the server); only the refused rows are put back into the cached page, the token is left empty, and the deferred refresh or the reload runs with the batch note.

**Regressions** (`tests/qml/tst_batch_actions.qml`, real App/Service/MailAccount with a controlled client): a partly refused trash with `nextPageToken = "older-page"` now produces a list request and keeps the restored rows on screen while it runs; a `loadMessages(false, true, "")` made while a partly refused trash is pending is deferred and runs once the refused row is back, with the accepted one gone and the note up; the same for a batch that finishes whole. These fail on the previous head.

Also updated the switcher-order expectations in `tests/test_model.js` for #128's sorted rail. `QT_QPA_PLATFORMTHEME= make validate` passes (452 QML results, sidebar text and Outlook HTTP suites included).

Please take another look.

huacnlee:
@gw7523 After trying this in the running app, I found that the title-bar Inbox/folder dropdown duplicates the navigation already available in the sidebar. In practice, the extra controls make the header busier and the interaction less clear, so I've removed the header scope switchers and the separator chevron in #148 and restored the Omamail title.

This is a scoped rollback of the header interaction, not a revert of the whole PR. Multi-message selection is useful and is being kept, with refinements to its controls and keyboard/mouse behavior. #148 is still open; this comment records why the header portion is being rolled back.

PR #111, opened 2026-09-06
Add a JMAP provider, with conversations, a reader rail and push

Depodra:
## What this adds

A fourth mail provider: **JMAP** (RFC 8620/8621), as spoken by Fastmail, Stalwart and anything else that implements it. A mailbox is an address plus an app password or an API token. The server is discovered from the address's domain (the `_jmap._tcp` SRV record, then the domain's `/.well-known/jmap`), or typed into a **Server settings** field when discovery finds nothing — which is the reference Stalwart's case. Sign-in verifies the secret by using it: the session GET under Basic and, only on a 401, once more under Bearer, then one `Mailbox/get`; whichever scheme answered is recorded on the account.

On top of reading, triaging and sending, a JMAP mailbox gets two things the other providers do not have:

- **Conversations as rows.** The listing collapses to one row per thread, with a badge for the count. Each row carries a `thread` block (`{ id, count, unread, flagged, memberIds }`) computed from the members the server returned for the view, so an unread reply keeps its row bold and unread even when the representative is read.
- **A conversation rail beside the reader.** Every member of the open conversation, newest first, as a stop that opens in the same reader. `n` and `p` walk it; a right-click on a stop opens the row's own menu (reply, archive, trash, spam, read, star) scoped to that one message, and the reader moves to the neighbouring stop if the open message leaves the view.
- **Push.** One `text/event-stream` connection per account under a Quickshell `Process`, parsed line by line, with a watchdog at twice the server's ping, a backoff from a second to five minutes, a resume-from-suspend detector on the wall clock, and a stop on a rejected credential. A `StateChange` becomes a plan (`{ mail, mailboxes }`) that runs the account's own `loadLabels()` and `refresh()`; the two-minute poll stays the floor.

Sending is one upload of the raw RFC 5322 bytes `Message.buildRawMessage` already produces, then `Email/import` + `EmailSubmission/set` in one request, with `onSuccessUpdateEmail` moving the copy to Sent. Drafts are import + destroy. No `Email` is ever built from parts, so the direction twin, calendar replies and nested boundaries are byte for byte what the other providers send.

## How it is built

The seam the existing providers use is kept exactly: one object that signs in (`providers/JmapAuth.qml`) and one that fetches (`providers/JmapClient.qml`), built by `MailAccount` from the provider id, with the decisions in a pure library (`providers/JmapProtocol.js`, `JmapThreads.js`, `Jmap.js`) that node tests exercise without a compositor.

Every request crosses the network through `scripts/jmap-transport.sh`, the JMAP twin of `mail-transport.sh`: five verbs (`session`, `call`, `download`, `upload`, `stream`), fields base64 on one stdin line, the credential assembled by the script and never in QML, curl fed its config on stdin with `-q --globoff`, `proto`/`proto-redir` pinned to HTTPS, `max-redirs 0`, `noproxy`, per-verb deadlines and a 20 MB ceiling on every read answer. QML's `XMLHttpRequest` is not used, because it follows a 3xx by itself re-sending the Authorization header. The session object — which is where every later URL comes from — is cached beside the query cache and re-verified when read back.

Secrets: the app password or token lives in GNOME Keyring under `service omamail`, `kind jmap-secret`, `account <id>`, written over stdin. `accounts.json` holds `{ sessionUrl, username, authScheme, accountId }` and nothing that could authenticate.

## What it touches in the existing architecture

Most of the diff is new files. The shared changes, each with the reason:

- **`message/Message.js`** — every summary now carries a `thread` block (`threadOf` / `normalizeThread`), absent for the three existing providers and read as a count of 0, so one list, one reader, one cache and one set of actions serve all four. The raw-message builder writes `Date` and `Message-ID` (falling back to the signed-in address's domain), because a JMAP server stores exactly the bytes it is handed; Gmail and SMTP servers keep or replace their own. `buildSendPayload` carries `draftId` so a provider that can destroy the old draft in the same request does.
- **`account/Model.js`** — action scope: `actionTargets` expands a conversation row to its counted members above the seam, so no client ever calls `Thread/get` for an action; `survivesAction` reads the row's recomputed block, so a mark-read in Unread keeps a row while a reply is unread. `Model.setupState` is unchanged.
- **`account/MailAccount.qml`** — member summaries for the rail (`getSummaries` on every client; the three existing ones answer an empty list), the row-versus-member distinction in `act` (a new fourth argument scopes an action to one message), and the push wiring (`remoteChanged` → `loadLabels()` then `refresh()`). Actions on a row still go through the same optimistic-update-then-restore path.
- **`account/Accounts.js`, `providers/Registry.js`, `providers/Credentials.js`** — `jmap` in the provider list with `makeJmapSettings`; two new capability words, `conversations` (the listing collapses) as distinct from `threads` (the server has thread ids); a per-account `refusals` argument to `Registry.can`, because on JMAP a missing Archive mailbox is a fact about the account rather than the protocol; a keyring kind of its own.
- **`cache/Cache.js`, `cache/CacheStore.qml`** — a `session` slot keyed on the URL it came from and the server's `state`.
- **`App.qml`, `components/MessageReader.qml`, `MessageRow.qml`, `SettingsPage.qml`, `keys/Keymap.js`** — the JMAP setup page, the rail and its `n`/`p` keys (the list cursor stays put), the count badge, a `JMAP · host` detail line under the mailbox, and the right-click menu opened member-only.
- **`providers/GmailApiClient.qml`, `HeyClient.qml`, `ImapClient.qml`** — the `getSummaries` interface and nothing else in behaviour.
- **`AGENTS.md`, `README.md`, `docs/KEYS.md`** — the provider seam and the two new capability words are written down where the others are.

Nothing changes for an existing Gmail, HEY or IMAP account: rows carry no block, the rail does not draw, and the transport scripts they use are untouched.

## Security

Reviewed against AGENTS.md's release-blocking rubric, with the transport script run against a local HTTPS lab under real curl 8.21.0 (redirects reported and never followed, the size ceiling holding mid-transfer on chunked bodies, proxies ignored, upload directory 0700/0600, TERM cleaning up mid-transfer and mid-retry) and the protocol library exercised in node. Verdict PASS; the review's findings are fixed in the last commits of the branch, each with a regression test that fails on the previous code:

- no request carries a refused credential until sign-in clears the flag (the stream already stopped; the poll did not);
- a stream is believed only once it has lasted a whole ping interval, so a server that closes cleanly after one line backs off instead of reconnecting at once;
- a reply past the 20 MB ceiling sends no body up the pipe;
- a session naming a non-HTTPS address is refused where the session is judged;
- the plain-text guard follows the rail's and the badge's strings;
- a request's work directory lives under `XDG_RUNTIME_DIR` and stale ones are swept;
- a discovery hop carrying userinfo is not followed.

Accepted by design and recorded as such: the credential goes to the URL that answered the session request and to the addresses inside its own session object, which may be other hosts (Fastmail's download host is).

## Tests

`make validate` exit 0 on the tip: 313 QML tests, the node suites (`test_jmap.js`, `test_jmap_helpers.js`, `test_jmap_threads.js`, `test_conversation.js`, and the extended `test_message.js`, `test_model.js`, `test_accounts.js`, `test_credentials.js`, `test_provider.js`), the shell suites (`test_jmap_transport.sh` against a faithful curl stub, `test_jmap_srv.sh`, `test_curl_config.py` extended to the JMAP verbs), and new `test_source.sh` rules: no `Authorization` value assembled in QML, no credential word in the account's settings maker, `getSummaries` on every client.

Exercised live against a Stalwart 0.16 server and a Fastmail account during development.


huacnlee (changes requested):
Thanks for the substantial work on this. My position remains the same as in #39: I don't want to introduce a separate JMAP protocol stack into Omamail at this stage.

This proposal goes beyond the earlier Fastmail-only, read-only implementation: it describes a generic provider with discovery, core write and submission actions, conversations, and push. Those address limitations of the earlier PR, and I don't want to repeat objections that no longer apply.

However, generic support and complete core actions alone are not the acceptance bar. We still need a concrete product need and a demonstrated advantage over the existing IMAP/SMTP path that justifies the additional authentication, protocol, discovery, synchronization, setup, and maintenance surface. The proposed conversations and push features do not by themselves establish that justification, and the PR does not provide measured comparative evidence of a synchronization or performance advantage.

As I explained in #39, Fastmail is already supported through the generic IMAP/SMTP provider with normal mail actions. I am not taking on this additional protocol and maintenance commitment now. We can reconsider when broader adoption or a concrete Omamail need changes that tradeoff. Please close this PR; this is a product-scope decision, not a request for another implementation revision.

Security verdict: NOT VERIFIED. This review is limited to product scope; I have not independently audited the new credential/network boundaries or reproduced the reported security tests. The PR's reported PASS is not an independent security approval.

Depodra:
Apologies for missing #39. I built this for my own Stalwart account and only went looking for prior art after opening the PR, so the "Fastmail-only, read-only" objections were news to me. Fair enough that a product-scope decision is yours to make, and I will not argue the scope itself.

I do want to fill the one gap the review named, since "the PR does not provide measured comparative evidence" was a fair criticism at the time. I have now measured it, and the write-up is here: https://staging.rig2dig.com/reports/jmap-vs-imap.html

**What was measured.** The same mailbox on the same Stalwart server, over IMAP and over JMAP, driven through Omamail's own transport scripts (`mail-transport.sh` and `jmap-transport.sh`) with the exact command strings and JSON bodies the two providers produce. So it is not a protocol benchmark in the abstract; it is what the panel waits on. Median of five, inbox of ~122k messages:

| Operation, as the panel issues it | IMAP | JMAP |
|---|---|---|
| Inbox page, 25 rows | 8.4 s, 3.3 MB, 6 connections | 0.7 s, 38 KB, 2 |
| Unread badge poll (every 120 s) | 8.7 s, 3.7 MB | 0.6 s, 4.7 KB |
| Sidebar counts, 11 folders | 3.0 s | 1.1 s (0.11 s if fetched in one call) |
| Open a 3 MB message with attachments | 0.8 s, 2.97 MB | 0.11 s, 15 KB |
| Mark read | 283 ms | 89 ms |
| New mail reaches the panel | 0–120 s (poll) | ~90 ms (push) |

**Why the gap is structural rather than a Stalwart quirk.** Omamail's transport is a fresh `curl` process per request, so every IMAP call pays TLS + `LOGIN` + `SELECT` (~240–300 ms here against ~90 ms for one HTTPS `POST`), and a stateless client has no "newest 25" without first pulling the folder's whole UID list (`UID FETCH 1:* (UID)`), which is the 3.3 MB and ~5.6 s above and runs again on every badge poll. A persistent IMAP client with `QRESYNC` would avoid that, but that is a different architecture from the one Omamail has. JMAP's `Email/query` takes a sort and a limit, chains the row fetch into the same round trip, and pushes state changes over one open stream, which is exactly the shape a process-per-request client needs.

**Where it is fair to IMAP.** Sign-in is level, typed search is close (646 vs 457 ms), a plain message costs the same bytes, and on an 877-message folder the gap narrows to ~2.4× (the remaining connection count). The report says all of that too; I did not want to hand you a one-sided number.

**On product need.** The two user-visible outcomes are new mail arriving in under a second instead of on a two-minute poll, and a badge that costs kilobytes rather than megabytes per poll on a large inbox. Those are the reasons I wanted it for my own use.

If that changes the tradeoff, I am happy to trim the PR to whatever surface you would actually accept, or to split conversations and push out so the core provider is a smaller diff. If it does not, I will close this and keep the branch on my fork for anyone else on Stalwart or Fastmail who wants it. Either way, thanks for the clear acceptance bar; it is what made the measurement worth doing.

providers/JmapClient.qml
```diff
@@ -0,0 +1,1876 @@
...
+      }
+      function afterCandidate(candidate) {
+        if (candidate === "") walk(index + 1)
+        else attempt(candidate, 0, function() { walk(index + 1) })
+      }
+      if (step.kind === Jmap.STEP_SRV) {
+        lookupSrv(plan.domain, handle, function(url) {
+          if (handle.aborted) return
+          if (url === "") walk(index + 1)
+          else probe(url, 1, afterCandidate)
```
huacnlee: [P1] Authenticate the discovered destination before sending the secret

The SRV lookup accepts records without verified provenance (`resolvectl --legend=no` / `dig +short`), and this path promotes their target to a credential destination after only a 200 or 401 probe. A substituted SRV target with a valid certificate for its own hostname can therefore receive the mailbox password/token; TLS here authenticates that target, not its authority to serve the original email domain. In a Qt 6.11.2 test of the real client with mocked resolver/HTTP results, an unsigned `attacker.example.net` SRV answer followed by 401 produced a credential-bearing request to that host. DNS poisoning itself was not staged. Please require authenticated discovery provenance or establish delegation through the original domain’s HTTPS well-known endpoint before sending credentials. This is the risk described in RFC 8620 §8.3: https://www.rfc-editor.org/rfc/rfc8620.html#section-8.3

providers/JmapPush.qml
```diff
@@ -0,0 +1,374 @@
...
+    // that is measured rather than chosen.** Quickshell's `SplitParser` never
+    // hands over an empty segment: given `a\n\nb\n` it emits `"a"` and then
+    // `"\nb"`, gluing the delimiter it skipped onto the front of the next
+    // chunk. Splitting on `"\n"` therefore never delivers the blank line that
+    // ends an SSE event — the event completes only when the *next* one arrives,
+    // and the last event before a quiet period never completes at all. On the
+    // reference server that looked exactly like push not working: the two lines
+    // of the event arrived and nothing happened. Splitting on `"\n\n"` hands
+    // over one whole event per read, the moment the server finishes it.
+    stdout: SplitParser {
```
huacnlee, line 308: [P1] Enforce the stream size limit before buffering a complete event

`SplitParser` retains everything up to the next blank-line delimiter before invoking `readBlock`, so the downstream `MAX_EVENT_CHARS` check cannot bound that allocation. The stream transport has no byte ceiling. A native Quickshell 0.3.1 Process/SplitParser probe buffered 8,388,614 bytes without a callback, then delivered the whole block when the delimiter arrived, despite the intended 65,536-character limit. A malicious or broken configured server can consume desktop-process memory before the policy runs; the watchdog bounds elapsed time, not bytes. Please enforce an incremental byte limit before this unbounded accumulation. This was a local synthetic stream test; no OOM was attempted.

providers/JmapProtocol.js
```diff
@@ -0,0 +1,2898 @@
...
+//
+// An Email is immutable apart from its keywords and its mailboxes — a subject
+// patch is refused `invalidProperties` — so a saved draft is always a new id,
+// as it is on IMAP. There is no `updateDraft`.
+function saveRequest(accountId, blobId, roles, draftId) {
+  var map = roles && typeof roles === "object" ? roles : {}
+  var account = trimmed(accountId)
+  var calls = [importCall(account, blobId, trimmed(map.drafts), "0")]
+  var stale = trimmed(draftId)
+  if (stale !== "") calls.push(["Email/set", { accountId: account, destroy: [stale] }, "1"])
```
huacnlee: [P1] Preserve the old draft until replacement or submission succeeds

This unconditional `Email/set destroy` still executes when the preceding `Email/import` returns `notCreated` (for example `overQuota` or `invalidEmail`). The sibling destroy in `sendRequest` has the same problem when import or submission fails; submission failure also causes the client to clean up the new imported copy. JMAP method calls are ordered but are not a transaction: later calls continue after earlier errors (RFC 8620 §3.6.2). A probe of the real request builders with synthetic sequential responses confirms that a refused replacement still includes deletion of the original server draft. Please defer deleting the old draft until successful import for Save, and successful submission for Send, and cover both failure cases. https://www.rfc-editor.org/rfc/rfc8620.html#section-3.6.2

account/MailAccount.qml
```diff
@@ -1323,15 +1472,23 @@ Item {
...
+    var refused = Provider.refusal(providerId, needs, capabilityRefusals)
+    note(refused !== "" ? refused : Model.actionUnavailable(action, Provider.badge(providerId)))
     return true
   }
 
-  function act(id, action, quiet) {
+  // `memberOnly` is the rail's: a stop names its one message, not the row's.
+  function act(id, action, quiet, memberOnly) {
     var messageId = String(id || "")
+    var oneMessage = memberOnly === true
```
huacnlee, line 1493: [P1] Preserve member-only scope through the action queue

A rail action calls `act(id, action, false, true)`, but the pending-action branch below drops `memberOnly`; neither the queue record nor `runQueuedAction` restores it. If the selected rail member is also the conversation representative, queuing its Trash while another mutation finishes replays it as a conversation-row Trash. A Qt 6.11.2 regression test on the real MailAccount/JmapClient recorded `Email/set` updates for `m1,m2,m3` where the requested single-message action must name only `m3`. Carry the scope through enqueueing, coalescing and replay so latency cannot expand which messages a destructive action targets.

account/MailAccount.qml
```diff
@@ -1352,7 +1509,33 @@ Item {
...
+    // mark-read through the queue, and star or unstar through the branch below.
+    if (index < 0 && previewIndex < 0) {
+      if (!Conversation.holdsMember(selectedThread, messageId)) return false
+      var memberChange = Model.labelChangesFor(action)
+      if (!memberChange) return false
+      if (quiet !== true) return actOnDetachedMember(messageId, action, memberChange)
+      applyMemberChange(messageId, action)
+      if (selectedId === messageId && selectedMessage)
+        selectedMessage = Model.applyLabelChange(selectedMessage, action)
+      queueQuietAction(messageId, action, cacheKey)
```
huacnlee: [P2] Use the current queue API for detached members

When the conversation row is no longer in messages/previewMessages but its rail is still open, opening another unread member reaches this branch. `queueQuietAction` and `runQueuedQuietAction` no longer exist after the queue rename in this PR. A real Qt 6.11.2 MailAccount test throws `ReferenceError: queueQuietAction is not defined`, after the local summary has already been changed, and dispatches no mark-read request. This can occur when a refresh removes a now-read conversation from the Unread view while the reader retains its rail. Route this through the current queue functions and cover the detached-member path.

providers/JmapPush.qml
```diff
@@ -0,0 +1,374 @@
...
+      splitMarker: "\n\n"
```
huacnlee, line 309: [P2] Accept CRLF event boundaries in the live stream

A legal SSE event ending in `\r\n\r\n` never matches this literal LF-LF delimiter. The CR stripping in `readLine` cannot help because it runs only after splitting. With native Quickshell 0.3.1, a complete CRLF ping on a still-open stream produced zero callbacks; appending LF LF later delivered the buffered event. CRLF servers therefore deliver neither prompt StateChange notifications nor pings to the watchdog. The stream parser needs to recognize SSE line endings while preserving the incremental size bound requested separately.

providers/JmapClient.qml
```diff
@@ -0,0 +1,1876 @@
...
+
+  function readCall(reply, callback) {
+    if (reply.exit !== 0 || reply.status !== 200) {
+      // A method that failed comes back inside a 200 with an `error`
+      // invocation, so a 404 from the API URL is not a missing message — it is
+      // a URL that is no longer there, which is what a restored session whose
+      // server has moved since looks like. Dropping it has the next read fetch
+      // a fresh one rather than failing against the same stale address for as
+      // long as the account is open.
+      if (reply.status === 404) forgetServer()
```
huacnlee: [P2] Invalidate the persisted session when its API URL returns 404

`forgetServer()` clears only in-memory state. The next `ensureSession()` immediately reloads the same valid-looking session from `cache.getSession(url)` and returns without calling `fetchSession`. A server that moves its API endpoint therefore leaves this account repeatedly requesting the old 404 URL, including after restart, rather than recovering as the comment promises. A probe executing the unmodified reset/ensure functions restored the stale endpoint three consecutive times without a session fetch. Invalidate or bypass that cached entry on this failure so the next read discovers the new API URL.

huacnlee (changes requested):
Thanks for the clarification and the Stalwart comparison. I am open to adding JMAP as an optional provider for servers that natively support it; the earlier product-scope objection should not be treated as a request to close this PR. It does not require an extra relay for those servers. The current chooser places JMAP before IMAP (Gmail → Outlook → HEY → JMAP → IMAP).

I reviewed head 6c458eee5e3863ca65b7c931a3fe65a10d6d9686 for merge readiness. Please address the inline correctness and security findings before merging, especially credential discovery, stream buffering, failed draft replacement, and queued single-message actions expanding to a conversation.

Security verdict: BLOCK. The new discovery path can route credentials to an unauthenticated SRV-selected origin, and the live stream’s size check runs after unbounded parser buffering. These are separate from product scope.

Validation: existing JS and shell tests passed, as did 345 QML tests, the 4-result native Outlook HTTP suite, and qmllint (existing warnings remain). The shell suite needed TMPDIR=/tmp in this sandbox. Additional targeted Qt 6.11.2 tests failed on the queued-member scope, detached-member automatic read, and SRV credential-routing cases. Native Quickshell 0.3.1 probes demonstrated the oversized block and CRLF framing behavior. Draft failure/session recovery probes used the real functions with synthetic responses, not live-server tests. No real credentials or mailbox mutations were used; this is a scoped PR review, not a complete audit of every existing provider.

huacnlee (approved):
I have pushed fixes for all seven findings and independently re-reviewed the changes.

- Discovery now starts at the mailbox domain’s HTTPS well-known URL or an explicitly entered server URL. Unsigned SRV discovery has been removed; domains without that HTTPS endpoint need the server field filled in.
- The stream helper bounds events before desktop buffering, handles CR/LF/CRLF, drops empty frames, and terminates curl on rejection or cancellation.
- Original drafts are deleted only after confirmed replacement or submission. A failed cleanup preserves the successful save and its new draft ID.
- Queued rail actions preserve single-message scope through coalescing and replay. Detached automatic reads dispatch normally and roll back on failure.
- An API 404 invalidates the persisted session so the next read fetches a fresh endpoint.

Validation: `TMPDIR=/tmp QT_QPA_PLATFORMTHEME= make validate` passed, including 357 QML results, the native HTTP suite, seven new stream tests, JS and shell checks, and qmllint. Regression tests failed on the original defects before the fixes. Tests use synthetic credentials and controlled/stubbed transports; no live mailbox was modified.

Security verdict: PASS for the affected boundaries. The independent review found no remaining blocker in these seven fixes. Ready to merge from this review’s perspective.

Depodra:
Thanks for taking this on, and for the fixes to discovery, the stream bounds and mutation recovery before merging. Much appreciated! I'll keep testing against my JMAP and will open issues for anything that turns up.

PR #112, opened 2026-09-06
Nest folders and remember resized pane widths

gw7523:
**Stacked on #109.** This branch contains #109's commits; only the last commit, `Fold folders under their parents and drag the rail's width`, is this PR. Once #109 merges, this rebases to its own diff.

## What changes

The rail draws labels as a tree. "Archive/2026" sits indented under "Archive", siblings sort by name, a chevron folds a parent and a folded parent shows its subtree's unread count. An ancestor no label names — a `\Noselect` container, or a Gmail label whose parent was never made — is drawn dim and only folds. IMAP folders nest by the delimiter the server reported in LIST; Gmail labels nest by `/`. The fold state persists in `window.json` and the Ctrl digits follow what is on screen, so a digit never opens a folded row.

A divider beside the rail drags its width the way the list's divider does; double-click resets. Both widths persist in `window.json` and are restored on open.

## How

- `Model.labelTree(labels, collapsedPaths)` builds the rows depth-first, siblings in name order, folded subtrees left out; `Model.visibleLabels` is the same order for the digits. Children are keyed on a prototype-less object, so a label named "constructor" is a label.
- `ImapClient` carries the server's `delimiter` on each label; `Model.labelDelimiter` reads it, defaults missing metadata to `/`, and keeps an explicit NIL delimiter flat.
- `Model.paneWidth` and `windowPrefs` carry `sidebarWidth`, `listWidth` and `collapsedFolders`; the service writes them when a drag ends, not while it runs.

## Verification

Reviewer validation on `3096829b954da720c4447eb1529257e32dca0783`: `QT_QPA_PLATFORMTHEME= make validate` passed, including JavaScript/shell checks, QML tests, native Outlook HTTP tests, sidebar text tests, qmllint and plugin validation. Tests cover folder folding, provider IDs and delimiters, shortcut numbering and plain-text tree labels.

Pane allocation now subtracts the actual sidebar and both dividers and reserves reader space while preserving the saved preferred widths. Two new Qt cases reproduced a reader width of -10 before repair. After repair, all three layout cases pass across both update orders, collapsed rail, reopening and resizing across the compact threshold. These tests set width properties directly. This branch also contains the reader-star selection repair from #109; the combined focused suite passes 16 cases. Fresh independent review found no repair defects; Security: PASS for the affected scope.

The author previously reported live nested-folder exercises on Office 365 IMAP and iCloud. Manual light/dark and wide/compact visual validation remains unverified by the reviewer.

## Release Notes

- Labels and folders nest under their parents in the rail; a chevron folds a parent, and a folded parent counts its unread.
- The rail's width can be dragged, and both pane widths are remembered.


gw7523:
Rebased onto `d5a5d03` over the rebuilt #109; no changes beyond the rebase. `make test-js`, `make test-shell` and `make test-qml` pass on the branch.

gw7523:
Rebased onto `062f0f0` over the rebuilt #109; no changes beyond the rebase. `make test-js`, `make test-shell` and `make test-qml` pass on the branch.

huacnlee (changes requested):
Product decision: accepted for this PR’s folder hierarchy/folding and adjustable, persisted pane widths. These fit nested mailboxes and long folder names. Approval is blocked by the findings below, not by that product scope.

Reviewed commit `8577c39912c9ae6a21a40ef351127e30663900c9`.

1. **P1 — Preserve and verify the sidebar’s plain-text boundary** (`components/MailboxSidebar.qml`, label Text binding). The affected server-label consumer still uses `Text.AutoText` on this head. The real sidebar fixture observes `textFormat = 2`, rather than `Text.PlainText = 0`. Current main already adds explicit plain text here; preserve that protection when rebasing and cover the tree-label and tooltip consumers with appropriate rendering regressions. This is a pre-existing boundary affected by the new tree, not a claim that this PR introduced it. No live resource-fetch exploit was attempted.

2. **P2 — Folding a parent also opens it** (`components/MailboxSidebar.qml:235`, with the ancestor row’s TapHandler). Clicking the chevron emits both `folderToggled` and `labelSelected`. A Qt test using the actual sidebar records one of each, so folding unexpectedly navigates to that folder. Make the two actions mutually exclusive and add a pointer regression asserting that folding emits no selection.

3. **P2 — Respect an explicitly absent IMAP hierarchy delimiter** (`account/Model.js:1552`). LIST’s NIL delimiter becomes an empty string, but the tree replaces it with `/`. A flat folder named `a/b` therefore becomes a synthetic parent and child. Distinguish missing provider metadata from an explicit no-hierarchy delimiter. Cover NIL, slash and non-slash delimiters.

4. **Prerequisite blocker — #109’s partial-batch completion remains unresolved** (`account/BatchAction.qml:185–192`). This branch includes the exact previously reviewed parent `ca44baf` and does not change its partial-failure branch. It restores refused rows but returns before pagination/deferred-refresh cleanup. Resolve that prerequisite or remove the dependency before integration; retain the corresponding runtime regressions.

**Security: NOT VERIFIED** for the affected label-rendering boundary. There is no new transport or credential destination in this PR’s own delta, but that does not establish safe rendering of server-controlled labels.

Validation on the reviewed head: `QT_QPA_PLATFORMTHEME= make validate` exited 0, including JavaScript/shell checks, 450 QML tests, 4 Outlook HTTP tests, qmllint and plugin validation. The separate review fixture reports 2 passed / 2 failed for folding and plain-text handling; the Node delimiter probe demonstrates the incorrect hierarchy. Runtime: Qt/QtTest 6.11.2, Omarchy 4.0.2. Tests used synthetic data and no real mailbox or credentials. Manual light/dark and wide/compact visual validation remains unverified.

Performance, non-blocking observation: `visibleLabels` performs a label scan per row; Node samples took approximately 8/42/157 ms for 1,000/5,000/10,000 labels. These are not Qt UI timings. Prefer indexed lookup and bounded tree traversal when revising this code; no live-provider scale claim is made.

gw7523:
Rebased on current main (`85eb6c2`) and the four findings addressed.

1. **Plain text preserved.** The tree's row Text carries `textFormat: Text.PlainText` from #128; the rebase keeps it. `tests/test_sidebar_text.py` now drives the tree: a server name with the delimiter in it becomes several rows and every one of them must be a `Text.PlainText` (the leaf and the parents the server never named), and the tooltip receives the row's path as one unchanged string (the shell's `PanelToolTip` draws `Text.PlainText`; the stub here records what it was handed). A NIL-delimiter row is added for every fixture name.
2. **Folding no longer opens** (`components/MailboxSidebar.qml`). The fold's `TapHandler` sits inside the row's and Qt hands the tap to both, so the row's handler now checks where the tap landed and stands down on the fold. `tests/qml/tst_sidebar_fold.qml` drives real pointer events on the real sidebar: a click on the chevron emits one `folderToggled` and no `labelSelected`; a click on the name emits one `labelSelected` and no fold; a child row opens by the server's id.
3. **NIL delimiter respected** (`account/Model.js`). New `labelDelimiter(label)`: undefined/null → `/` (Gmail nests on it), `""` (what the IMAP provider passes for LIST's NIL) → no hierarchy at all, so `a/b` is one folder. `labelTree` uses it; `tests/test_model.js` covers NIL, slash and dot delimiters.
4. **Prerequisite**: #109's partial-batch cleanup is fixed on its head and this branch is rebuilt over it.

One more, found while re-checking: the key numbers and the switcher were still sorted A to Z by full name (`sidebarSlots` re-sorted what App handed it), while the rail draws the tree parent-first, so with `Work`, `Work/Invoices` and `Work (old)` the digit beside a row could open a different one. `sidebarSlots` now keeps the order it is given — App hands it `visibleLabels`, the tree with folded rows left out — and `tests/test_model.js` asserts a child is numbered right after its parent and a folded child has no number.

On the performance note: `visibleLabels` now indexes the labels by id once rather than scanning per row.

`QT_QPA_PLATFORMTHEME= make validate` passes. Please take another look.

PR #113, opened 2026-09-06
Manage labels, search by address, and monitor label counts

gw7523:
This PR adds label management, address shortcuts and watched-label counts. Prerequisites #109 and #112 have merged; this branch now contains its own changes over main.

## What changes

**Labels are managed where they are.** A right-click on a label in the rail opens a menu: Rename, New label beside, New sub-label, Move to (a picker with Top level and every other label except the label itself, its children and its current parent), Monitor for new mail, Delete (confirmed, with the note that IMAP deletes the folder's mail while Gmail keeps the messages' other labels). The label being viewed follows its own rename; a deleted label sends the window to the inbox.

**An address can be acted on from the reader.** A right-click on the From line or the To line — To, Cc and Bcc, which sent mail carries — offers Copy address, Search mail from and Search mail to. A line with several addresses asks which one first.

**A label can be watched.** A monitored label is counted on every refresh, shows an eye and its unread count in the rail, and the status line says "N new in Label" when it gained mail.

## How

- A `manageLabels` capability on the registry: Gmail and IMAP true, HEY false (its labels are HEY's own), so HEY's menu offers only the watch.
- Gmail: `labels.create`, `patch` and `delete`; a nested label is a path with `/`, so a move is a rename. IMAP: `CREATE`, `RENAME`, `DELETE` — a typed name encoded in modified UTF-7 by a new `encodeMailbox` (round-trip tested against `decodeMailbox`), a name the server listed sent exactly as LIST spelled it, and `RENAME` carrying the subtree, which is what a move is. Paths are composed and shown from decoded names.
- `Model.labelLeaf`, `labelParent`, `labelPathJoin`, `labelNameProblem` and `labelMoveTargets` hold the naming rules under test.
- `Provider.addressQuery(id, field, address)`: `from:`/`to:` on Gmail, `FROM "…"`/`TO "…"` on IMAP, a word search on HEY. Copy goes through `wl-copy` as one argument, no shell.
- Monitored ids live on the account entry (`Accounts.toggleMonitored`); the account counts them through each provider's existing `getLabelCounts` on refresh.

## Verification

Reviewer `QT_QPA_PLATFORMTHEME= make validate` passed on `f58b3cee26145d5a82756192eb8451bc4c6eae97`: JavaScript/shell checks, 491 QML cases, 4 native Outlook HTTP cases, 64 sidebar text cases, qmllint and plugin validation.

Reviewer re-review of `f58b3cee26145d5a82756192eb8451bc4c6eae97`: the previously reported destructive IMAP folder-identity boundary and both watched-label migration defects remain unresolved. A controlled real-curl probe confirms changed folder identity; separate actual App/Service/MailAccount probes reproduce loss of a surviving child's watch and loss of one watch after two queued renames. Security: BLOCK. Reproduction details remain local under CONTRIBUTING.md.

Before-and-after screenshots and a visible-difference explanation are missing for the changed UI. UI evidence: BLOCK. Reviewer manual light/dark and wide/compact verification remains outstanding. Product acceptance for label management, address shortcuts and monitoring is pending.

The earlier account-ownership repair passes its tests. The former #109/#112 prerequisites have merged and their old findings are no longer treated as inherited blockers. The author previously reported live IMAP address-menu and monitored-label exercises; no reviewer live-provider mutation testing is claimed.

## Release Notes

- Right-click a label to rename it, create a label beside or beneath it, move it, delete it, or watch it for new mail.
- Right-click an address in the reader to copy it or search for mail from it or to it.

### DIFF
--- a/App.qml
+++ b/App.qml
@@ -1833,6 +1833,8 @@ Item {
           visible: !root.compact && !root.showPage && !root.composing
           collapsed: root.sidebarCollapsed
           calendarSelected: root.calendarVisible
+          menuLabelPath: labelMenu.opened && !!root.service
+            && labelMenu.accountId === root.service.activeAccountId ? labelMenu.labelPath : ""
           service: root.service
           textColor: root.foreground
           accentColor: root.accent
--- a/App.qml
+++ b/App.qml
@@ -1255,6 +1255,7 @@ Item {
     function onActiveAccountIdChanged() {
       root.clearChecksIfForeign()
       root.cursorId = ""
+      root.closeLabelPopups()
     }
     function onSidebarWidthChanged() { root.sidebarWidth = root.service.sidebarWidth }
     function onListWidthChanged() { root.listWidth = root.service.listWidth }
@@ -1463,12 +1464,82 @@ Item {
     })
   }
 
+
   function confirmDelete(request) {
     if (!service) return
     if (request.kind === "event" && request.event) {
       service.calendarController.deleteEvent(request.sourceId, request.event)
       calendarView.closeDetail()
     }
+    if (request.kind === "label" && request.labelId) service.deleteLabel(request.labelId, request.accountId)
+  }
+
+  // ------------------------------------------------------------ labels
+
+  // Every label popup — the menu, the name prompt, the move picker, the
+  // delete confirmation — is opened for one account and answers to it by
+  // id, whatever the window has switched to since. An IMAP label id is a
+  // folder name, which another account may well have too, so the open
+  // account is no guide to whose folder was meant. And a switch closes
+  // them: a prompt about a mailbox no longer on screen is a trap.
+  function openLabelMenu(labelId, path, sceneX, sceneY) {
+    if (!service) return
+    labelMenu.accountId = service.activeAccountId
+    labelMenu.canManage = service.canManageLabels
+    labelMenu.monitored = service.monitoredLabelIds.indexOf(labelId) >= 0
+    labelMenu.openAt(labelId, path, sceneX, sceneY)
+  }
+
+  function closeLabelPopups() {
+    labelMenu.close()
+    namePrompt.close()
+    labelMovePicker.close()
+    if (confirmDeleteDialog.request && confirmDeleteDialog.request.kind === "label") confirmDeleteDialog.close()
+  }
+
+  function labelDelimiterFor(id, accountId) {
+    var label = service ? service.labelById(id, accountId) : null
+    return Model.labelDelimiter(label)
+  }
+
+  function labelPathFor(id, accountId) {
+    var label = service ? service.labelById(id, accountId) : null
+    return label ? String(label.name || label.rawName || "") : ""
+  }
+
+  // A name asked for, then handed to the service with what it was for. The
+  // prompt is one component serving three asks, told apart by `kind`.
+  function askLabelName(kind, subject, title, initial, hint, accountId) {
+    namePrompt.accountId = String(accountId || "")
+    namePrompt.openFor(kind, subject, title, initial, hint)
+  }
+
+  function labelNamed(kind, subject, text, accountId) {
+    if (!service) return
+    if (kind === "rename") service.renameLabel(subject, text, accountId)
+    else if (kind === "create") service.createLabel(subject, text, accountId)
+  }
+
+  function copyAddress(address) {
+    var text = String(address || "").trim()
+    // Straight to wl-copy as one argument: no shell, and an address that
+    // starts with a dash is not an address.
+    if (text === "" || text.charAt(0) === "-") return
+    Quickshell.execDetached(["wl-copy", text])
+    notice = "Copied " + text
+    noticeTimer.restart()
+  }
+
+  function searchAddress(field, address) {
+    if (!service) return
+    var query = Provider.addressQuery(service.providerId, field, address)
+    if (query === "") return
+    // In the provider's own words, so it must not pass through the typed
+    // search's wrapping a second time; the box shows the words for it.
+    var text = (field === "to" ? "to: " : "from: ") + String(address || "").trim()
+    service.searchAddress(query, text)
+    searchBar.setQuery(text)
+    backToList()
   }
 
   FloatingWindow {
@@ -1773,6 +1844,9 @@ Item {
             root.backToList()
           }
           onFolderToggled: function(path) { root.service.toggleFolder(path) }
+          onLabelMenuRequested: function(labelId, path, sceneX, sceneY) {
+            root.openLabelMenu(labelId, path, sceneX, sceneY)
+          }
         }
 
         // Narrow windows lose the sidebar; the same mailboxes come back as a
@@ -1981,6 +2055,9 @@ Item {
           onComposeRequested: function(mode) { root.startCompose(mode) }
           onMailtoRequested: function(url) {
             root.openDraft(Mailto.parse(url))
+          }
+          onAddressMenuRequested: function(addresses, sceneX, sceneY) {
+            addressMenu.openAt(addresses, sceneX, sceneY)
           }
           onActionRequested: function(action) {
             if (!root.service || root.service.selectedId === "") return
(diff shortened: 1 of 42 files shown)


gw7523:
Rebased onto `d5a5d03` over the rebuilt #109 and #112. Two things changed in the rebase:

- Label changes and the watched labels' counts now live in `account/LabelActions.qml`, a small object beside the account (`MailAccount.qml` is within a few hundred bytes of the new 128 KiB ceiling); the service reaches them through `current.labelActions`.
- "Search mail from" and "to" from the address menu were being wrapped a second time by the typed-search path, so on IMAP they searched for the criterion's own text. `search(text, raw)` now takes an app-built query in the provider's words beside the words the box shows for it.

`make test-js`, `make test-shell` and `make test-qml` pass on the branch.

gw7523:
Rebased onto `062f0f0` over the rebuilt #109 and #112. In a merged list the address search is refused the way a move is (it is built in one provider's words); `main`'s merged-view `selectMailbox` and `search` are kept. `make test-js`, `make test-shell` and `make test-qml` pass on the branch.

huacnlee (changes requested):
Product decision: pending for this PR’s label management, address menu and label monitoring. The folder hierarchy and pane-width scope inherited from #112 has been accepted. The change request below is for technical blockers, independently of the pending product decision.

Label management fits mailbox organization but adds destructive provider-specific operations; recommend deferring it until independently verified. Address copy/search is a useful shortcut over existing functionality and is worth considering with accurate provider-specific wording. Monitoring overlaps existing refresh behavior and adds polling cost; recommend deferring it until its need and request bounds are settled.

Reviewed commit `067263f887fc5cb328660d2d53134f78733c3180`.

1. **P1 — Bind label mutations to the account that opened the interaction** (`App.qml:1485`, label confirmation construction around line 2718, and Service’s label-action forwarding). The confirmation retains a label ID but no owning account, while dispatch resolves the current account. An isolated test with the real App, Service and MailAccount demonstrates a destructive provider call crossing account ownership after the active account changes. Bind the owner when the menu/dialog opens and cancel or refuse stale operations; apply the same invariant to create, rename, move and delete. Regression tests must assert that no mutation reaches another account. No real mailbox was modified; this establishes the dispatch boundary violation, not a live-server exploit. Please coordinate any further security reproduction details privately under CONTRIBUTING.md.

2. **P2 — Migrate monitored IDs when IMAP folders are renamed or moved** (`account/LabelActions.qml:157`). After a successful rename and a fresh listing containing the new folder ID, the persisted monitored list still contains the old ID. The renamed row loses its monitor and subsequent polls target the old folder. Update the account’s persisted IDs, including affected descendants, and clean up deleted IDs. Cover successful/refused operations and persistence across restart. The controlled-provider test confirms the new ID is in account.labels before the stale monitored-ID assertion fails; live-server rename semantics were not tested.

3. **Resolve inherited blockers before integration.** This head still contains #112’s chevron double dispatch, NIL-delimiter handling and affected sidebar text-rendering boundary, plus #109’s partial-batch pagination/deferred-refresh cleanup issue. Those prerequisite findings remain applicable.

**Security: BLOCK** for the demonstrated account-ownership violation. New menu text is explicitly plain text and clipboard dispatch uses argv without shell interpolation. The existing transport paths are reused, but these facts do not compensate for wrong-account destructive dispatch. Real-transport coverage of the new folder mutations and the affected sidebar rendering boundary also remains incomplete.

Validation: `QT_QPA_PLATFORMTHEME= make validate` exited 0, including JS/shell checks, 450 QML tests, 4 Outlook HTTP tests, qmllint and plugin validation. The separate review harness reports 3 passed / 2 failed: account ownership and monitored-ID migration fail, while ordinary monitor callback completion passes. Qt/QtTest 6.11.2, Omarchy 4.0.2; synthetic data only. Manual light/dark and wide/compact visual validation remains unverified.

Performance observation: monitoring starts one count request per watched label on each refresh, without a new concurrency cap or cancellation owner. Request growth is inferred from the code, not measured against a live mailbox. Please bound this work if the monitoring feature proceeds.

gw7523:
Rebased on current main (`85eb6c2`) over the fixed #109 and #112, and both findings addressed.

1. **Label changes are dispatched to the account whose menu asked** (`App.qml`, `Service.qml`, the three popups). `openLabelMenu` stamps the menu with `service.activeAccountId`; the name prompt, the move picker and the delete confirmation carry that id; every answer goes through `Service.labelOwner(accountId)`, which resolves the owning `MailAccount` by id — never the account that happens to be open when the answer arrives — and refuses, with a note, when that account is no longer configured. `createLabel`, `renameLabel`, `moveLabel`, `deleteLabel` and `toggleMonitored` all take the owner. A switch closes the label popups. `tests/qml/tst_label_ownership.qml` (real App/Service/MailAccount, controlled clients per account): the menu opens on Ada, the service switches to Bob, and a rename, a delete and a create confirmed afterwards reach Ada's client only; a change for a removed account reaches nothing.
2. **Watched ids follow a rename or move, and leave with a delete** (`account/LabelActions.qml`, `Model.migrateMonitoredIds`). After the fresh listing lands, each watched id is mapped by path — the folder and every folder beneath it — to the id the listing now gives it; Gmail ids are unchanged and kept; ids the listing no longer has are dropped. The result is written to the account's entry through `Accounts.setMonitored` and saved. Covered in `tests/test_model.js` (rename, move, descendants, delete, Gmail, NIL delimiter) and in the QML test above (the entry and the write payload carry the new id).

Found while re-checking, and fixed here too: the moves waiting on a listing are a queue rather than a slot (two changes before a listing lands are both applied, in order); a listing that fails is retried once and otherwise leaves the moves waiting for the next listing rather than dropping them; a rename, move or create onto a name the list already has is refused before it reaches the server, since the row the window follows and watches after the change is found by its path; and a signed-out owner changes nothing and says so.

Inherited blockers: #112's chevron, NIL delimiter and text-format findings and #109's cleanup are fixed on their heads and this branch is rebuilt over them. The NIL case is honoured here too: with no hierarchy the path helpers treat every name as a leaf, and creating or moving under a parent is refused with a word about it (`Model.labelNestProblem`).

On the performance note: monitoring now counts at most three watched labels at a time, keeps the handles, aborts them when the account signs out or goes away, and carries a serial so a late answer from an abandoned poll is not taken off the next one's tally (`tst_label_ownership.qml` asserts the bound).

On the product questions: happy to split the address menu out into its own PR if that helps the decision; say the word.

`QT_QPA_PLATFORMTHEME= make validate` passes. Please take another look.

huacnlee (changes requested):
Three blockers remain on this revision.

1. **P1 — Preserve exact folder identity before destructive IMAP operations** (`providers/ImapProtocol.js:1097–1102`). Validate original server-provided identifiers before normalization, quoting, encoding or process dispatch. The new rename/delete paths reuse a helper that can change the identifier before the transport validator sees it. A controlled test through the actual curl transport demonstrated an operation against a changed identifier. Reject invalid original identifiers without any request, cover sibling mutation paths and Outlook, and retain valid international and escaped names. Security reproduction details remain private under CONTRIBUTING.md.
2. **P2 — Retain watches on surviving child folders** (`account/Model.js:1804`, deletion confirmation in `App.qml`). Deleting one IMAP parent does not delete its inferior names. The fresh listing can still contain a watched child, but migration drops it by ancestry. Reconcile watches against what actually survives and correct the confirmation wording. An actual App/Service/MailAccount regression reproduces this loss.
3. **P2 — Compose queued rename migrations before pruning** (`account/LabelActions.qml:260–266`). Two successful renames before the authoritative listing arrives lose the second renamed watch: the first migration filters out its old ID before the second can map it. Apply all pending transitions before filtering, or serialize the operations and listings. Cover unrelated and dependent renames.

This head also includes the previously identified reader-star selection and pane-allocation defects from #109/#112; incorporate their corrections before re-review.

Product decision: pending for this PR's label management, address shortcuts and watched-label notifications. Label management adds destructive operations; address shortcuts reuse existing search/copy behavior; monitoring adds polling proportional to watched labels. Recommend deferring management and monitoring until their correctness and need are settled. Scope acceptance remains separate from these technical blockers.

Security: **BLOCK** for the affected folder-identity boundary. The demonstrated result is a controlled boundary violation, not a claim about exploit delivery through a live provider. General tests passed on `b00f26c0214d5ff1bb9c2ab4ca0da1272775790e`: `QT_QPA_PLATFORMTHEME= make validate`, including 467 QML cases, 4 native Outlook HTTP cases and 64 sidebar text cases. Additional runtime probes reproduce both watch-loss defects. Qt 6.11.2 and curl 8.21.0; synthetic credentials and controlled local targets only.

Performance: monitor concurrency is bounded at three per account, but total requests remain linear in watched labels; live-provider timings and real cancellation behavior were not verified. Manual light/dark, wide/compact UI validation also remains outstanding. Non-blocking follow-up: describe address-search scope honestly for HEY and IMAP.

gw7523:

gw7523:
Rebased again, now on `68e4243` with #116 and #148 underneath. The mailbox-switcher block #148 removed is gone from this head too; the label popups this PR adds sit where it was. No other change of its own. `QT_QPA_PLATFORMTHEME= make validate` passes on this head.

gw7523:
Rebased onto v0.8.2 (`27c1b4a`, after #151); the head is now `f58b3ce`. The commit's content is unchanged — `git range-diff` shows context lines only — so the evidence in the description still stands. `QT_QPA_PLATFORMTHEME= make validate` passes on the new head.

PR #114, opened 2026-09-06
Add Reply-To, draft editing shortcuts, and imported signatures

gw7523:
**Stacked on #113.** This branch contains label management, address shortcuts and monitoring from #113 plus the compose changes below. #109 and #112 have merged.

## What changes

- **Reply-To.** A toggle beside Cc and Bcc opens a Reply-To field. It is written as one folded header (a line break typed into it cannot start a second header), carried through drafts, send and local recovery, and comes back into the field when a draft is reopened from the server.
- **Drafts round-trip.** In Drafts, Enter, `o` and `c` open the selected draft in the composer with what was written; a click still previews. Leaving the composer saves back over the draft it came from, and the Drafts list is read again from the server so the replaced copy goes and the new one appears at once. Sending removes the draft it came from on Gmail (`drafts.delete`, found the way an update finds it) and IMAP (the same commands a save uses to remove the copy it replaced). HEY draft deletion is currently a success no-op and remains incomplete. A save or a send asks for the mailbox's credentials first rather than refusing a mailbox whose token was refused a moment ago.
- **Numpad Enter.** `Enter` opens and `Ctrl+Enter` sends, beside Return; the help sheet names the keycap once.
- **Signature from a file.** Settings imports a signature from a picture or an HTML file, with a preview and a note of what was removed.

## Security argument for the import

What is stored is never the file. A picture is checked to be the raster it claims by its first bytes — an SVG named `.png` is refused — and wrapped in one `` within a size cap. Markup goes through `Html.sanitize`, the gate the reader already trusts for a stranger's message, and is then walked once more by `message/Signature.js` against a short list of what can never be in a signature: script, style, iframe, frame, object, embed, form and every input, svg, template, link, meta, media; every `on*` handler and every attribute that names an address; every `href` that is not a public http(s) link or a mailto; every `src` that is not an inline raster; and every inline style that fetches (`url(`), runs (`expression(`, `behavior:`), or hides text (`display:none`, `visibility:hidden`, zero opacity or font size, negative indent). What the file carries as formatting — colours, faces, sizes, tables, alignment — is kept, because that is what the owner imported it for. The preview draws the stored markup: the sent thing.

`tests/test_signature.js` holds an adversarial fixture with a script, a style block, an event handler, a `javascript:` link, a loopback link, a tracking image, an SVG passed off as PNG, a frame, an object, a form, a `background=` URL, a `url()` in a style and a zero font size, and asserts none of it survives while the words, the public link, the mailto, the real raster and the formatting do.

## Sending

A message with an HTML signature goes as `multipart/alternative` (`text/plain`, `text/html`), inside `multipart/related` when the signature carries pictures, each picture a part by `cid:` — Gmail's web client draws no `data:` image in HTML mail. The HTML twin escapes the body paragraph by paragraph and puts the signature markup where the plain signature stands; a plain signature the writer took out of the body is not put back as markup, and a picture-only signature sits before the quote. `tests/test_message.js` covers the shapes.

## Verification

Reviewer `QT_QPA_PLATFORMTHEME= make validate` passed on `8c22860da25dd99995e279e673f67151c8c8bfdd`: JavaScript/shell checks, 492 QML cases, 4 native Outlook HTTP cases, 64 sidebar text cases, qmllint and plugin validation.

Reviewer re-review on `8c22860da25dd99995e279e673f67151c8c8bfdd` confirms the two previously reported signature defects remain: changing the selected account during an import writes to the later account, and HTML placement replaces the first matching substring rather than the actual sign-off. The actual SettingsPage ownership probe and Node greeting-placement assertion fail. Qt UTF-8 import and the actual hosted unsubscribe completion probe pass.

HEY's adapter discards Reply-To and HTML/CID signature content while the UI offers them; its new draft-delete method performs no server deletion. Provider capability handling and accurate UI wording remain necessary. No live HEY or other real mailbox was used by the reviewer.

This head retains #113's folder-identity and watch-migration blockers. Security: BLOCK. Before-and-after UI screenshots and a visible-difference explanation are missing; UI evidence: BLOCK. Product decisions for Reply-To, draft key behavior, rich signatures and inherited #113 features remain pending. Required manual light/dark wide/compact screen validation remains unverified. The former #109/#112 prerequisite fixes have merged.

## Release Notes

- Compose has a Reply-To field.
- A draft opens into the composer from Enter, `o` or `c`, and is saved back over itself; Gmail and IMAP drafts are removed after sending.
- The numpad's Enter works like Return.
- A signature can be imported from a picture or an HTML file, with the file's formatting kept and anything that could run, fetch or hide text removed.

### DIFF
--- a/App.qml
+++ b/App.qml
@@ -1255,6 +1255,7 @@ Item {
     function onActiveAccountIdChanged() {
       root.clearChecksIfForeign()
       root.cursorId = ""
+      root.closeLabelPopups()
     }
     function onSidebarWidthChanged() { root.sidebarWidth = root.service.sidebarWidth }
     function onListWidthChanged() { root.listWidth = root.service.listWidth }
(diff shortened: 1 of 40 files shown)


gw7523:
Rebased onto `d5a5d03` over the rebuilt #109, #112 and #113. One change beyond the rebase: `MailAccount.qml` reached the new 128 KiB ceiling, so its unsubscribe block (the two functions and the `unsubscribe.py` process) moved beside it as `account/Unsubscribe.qml`, with its state (`unsubscribing`, `unsubscribeDone`) staying on the account where the reader reads it; the source guard for the public-IP-checked transport follows it there. `main`'s own draft-on-send comment in `ComposeView` is kept.

`make test-js`, `make test-shell` and `make test-qml` pass on the branch.

gw7523:
Rebased onto `062f0f0` over the rebuilt #109 to #113. The signatures now ride on `main`'s routed `send` and `saveDraft`, read from the entry of the mailbox the From address picked rather than the active one. `make test-js`, `make test-shell` and `make test-qml` pass on the branch.

gw7523:
Rebased on current main (`85eb6c2`) over the revised #109, #112 and #113. No change of its own; `QT_QPA_PLATFORMTHEME= make validate` passes on this head.

gw7523:

gw7523:
Rebased again, now on `68e4243` with #116 and #148 underneath. The mailbox-switcher block #148 removed is gone from this head too; the label popups this PR adds sit where it was. No other change of its own. `QT_QPA_PLATFORMTHEME= make validate` passes on this head.

gw7523:
Rebased onto v0.8.2 (`27c1b4a`, after #151); the head is now `8c22860`. The commit's content is unchanged — `git range-diff` shows context lines only — so the evidence in the description still stands. `QT_QPA_PLATFORMTHEME= make validate` passes on the new head.

PR #115, opened 2026-09-06
Integrate Omarchy AI assistance into mail and draft workflows

gw7523:
Omamail integrates the system AI beside messages, selections, and drafts, without an Agent settings page or an external terminal. The outline AI icon beside Compose opens a right dock whose left edge can be dragged to resize. Double-clicking its divider restores the default width. Claude runs silently using its existing system login and streams replies and public execution status into the panel; follow-up questions continue the same native conversation.

The chat uses shaded user messages with a › marker, plain-background AI replies, safe bold/code formatting, and a pure copy icon shown after generation finishes that briefly becomes a check after copying. The full-width input starts at one line, grows with newlines, sends on Enter, and inserts a newline on Shift+Enter. Typing / opens editable scenario commands. Messages submitted while AI is working enter a visible Pending queue, execute in order in the same conversation, and can be taken back for editing or removed. Failed starts retain messages; interruption or failure pauses the queue. Pending messages are session-only and bounded. The more-icon menu provides New chat and contextual conversation history. A timed Working line above the input advertises Escape to interrupt and / to show commands. Older messages remain readable while new output streams. The mail list has no bot icon, and the header AI entry has no breathing animation.

Translation and rewriting operate only on mail titles and bodies. Draft insertion extracts the body section and remains an explicit owner action. Jobs, results, and continuations stay bound to their account, message selection, and draft; retained files and streams are private and bounded. The background adapter currently supports Claude only, with an inline explanation for other system defaults. Existing system tool permissions apply; the bridge is not a sandbox, and detached tools may outlive cancellation.

Validation: full make validate, targeted keyboard/history/streaming/ownership regressions, safe-formatting tests, and native Omarchy-theme previews. A real installed-Claude smoke check passed two synthetic turns, including remembering context from the first turn; no real mailbox content was used.

The homepage and README lead with an updated AI-assistance screenshot, optimized to a 1600 × 871 WebP of approximately 116 KB.


k7cfo:
This is pretty cool

gw7523:
Rebased onto `d5a5d03` over the rebuilt #109 to #114; the agent button on the reader toolbar now sits beside `main`'s member-aware cursor rule. `make test-js`, `make test-shell` and `make test-qml` pass on the branch.

gw7523:
Rebased onto `062f0f0` over the rebuilt #109 to #114; the agent setting sits beside `main`'s new preview settings in the manifest, nothing else changed. `make test-js`, `make test-shell` and `make test-qml` pass on the branch.

gw7523:
Rebased on current main (`85eb6c2`) over the revised #109–#114, and the agent-ownership finding from the #108 thread addressed here, where the code lives.

**A job is its account's** (`agent/Agent.js`, `agent/AgentRunner.qml`, `Service.qml`, `scripts/agent-job.py`). Every payload names the asking account by the id the service gives it — `provider:address`, in `accountId` — and the runner writes it on the job and hands it down to continuations (with the parent's `accounts` list for a scope job). `jobFor`, `jobsByMessage` and `attentionByMessage` take the owning id and match a job to a message only inside it; the runner's `accountId` follows `Service.current`, so a row's glyph, its glow and its cancel read the open account's jobs alone. A job with no owner answers to no row. The pane still lists every job.

The popups follow the same rule as #113's label menus: the agent popup stamps the account it opened on, and its ask, its cancel and the job it shows are read inside that account (a unified row's id names its account itself; a selection across two mailboxes is refused with a note); a switch closes it. The composer's draft-agent card lists the open account's draft jobs only, and a finished job's note goes to the account it was about.

Regressions: `tests/test_agent.js` (colliding ids across two accounts, no owner, every payload carries the owner), `tests/test_agent_job.sh` (the owner is recorded and inherited; a scope continuation keeps its accounts), and `tests/qml/tst_agent_ownership.qml` (real Service with the stub processes: with Ada's job running on `42:INBOX`, Bob's row shows no job, glows for nothing, and `cancelAgent("42:INBOX")` starts no runner; Ada's own row cancels it).

`docs/AGENT.md` says so. `QT_QPA_PLATFORMTHEME= make validate` passes.

gw7523:
One fix since the last push: the Settings page called `Service.setAgentCommand` to save the default agent's command line, and the service had no such function — the field could only be saved through the plugin's generic settings panel. Added beside the other setters, with a Service-level test in `tests/qml/tst_agent_ownership.qml`. `QT_QPA_PLATFORMTHEME= make validate` passes on this head.

gw7523:

gw7523:
Rebased onto `68e4243` (#116 and #148 underneath). The App.qml comments were trimmed to stay under the 128 KiB source ceiling now that #148's row checkboxes sit in App; no change of behaviour. `QT_QPA_PLATFORMTHEME= make validate` passes on this head.

gw7523:
Rebased onto v0.8.2 (`27c1b4a`, after #151); the head is now `6d9536a`. The commit's content is unchanged — `git range-diff` shows context lines only — so the evidence in the description still stands. `QT_QPA_PLATFORMTHEME= make validate` passes on the new head.

huacnlee:
Don't update anything, I'm going to fix details.

PR #116, opened 2026-09-06
Prevent account edits from deleting another mailbox

gw7523:
Replaces #108, whose head branch grew into a development branch. This PR keeps the account-save correction scoped to account identity and persistence.

Editing one account to an address already held by another now refuses the change and preserves both rows. Password and OAuth sign-in stop when configuration is refused, so a newly entered secret cannot be handed to the unchanged account. Account saves also reject unintended omissions of previously persisted IDs.

Intentional identity changes release only the changed row's old ID from that protection. This applies both to settings edits and to a provider reporting its canonical address or recognizing a duplicate after sign-in. The canonical identity and subsequent settings can therefore be saved without weakening protection for other mailboxes, including when an earlier write is still queued.

## Verification

`QT_QPA_PLATFORMTHEME= make validate` passed on `76fa28a4fa364ce071240cbb9823e7fdba859f7b`: 475 QML cases, 4 native Outlook HTTP cases, 64 sidebar text cases, JavaScript/shell checks, qmllint and plugin validation. The new `AccountPersistence` suite passes 9 cases; four failed before the profile-identity repair. It covers queued writes, canonical addresses, duplicate resolution, first naming, invalid identities, later settings, refusal of unrelated omissions, and no password/OAuth dispatch after a collision.

A fresh independent agent reviewed the complete PR diff and found no blockers; it independently passed the account Node tests and all 9 new QML cases. Security: PASS for the affected account identity, authentication-dispatch and persistence boundaries. Tests use synthetic accounts and intercepted Process/sign-in endpoints; no live mailbox or keyring integration is claimed. UI evidence: NOT APPLICABLE — this PR changes no interface.

The branch was validated with main `a669567`; the subsequent main commit `174846d` changes header/selection UI and does not modify `Service.qml` or `account/Accounts.js`. The integration delta was inspected and GitHub reports no conflicts.

## Release Notes

- Saving one mailbox's settings no longer deletes another mailbox, and signing in with a canonical email address no longer prevents later settings from being saved.


gw7523:
Rebased onto `d5a5d03`, and the two findings from #108 are addressed here:

- **P1, sign-in after a refused save.** `configureAccount` answers whether the configuration was taken; a collision answers false and `configureCurrentAccountAndSignIn` / `configureCurrentAccountAndSignInOAuth` stop there, so the secret just typed is never sent to the mailbox still standing at the row.
- **P2, a corrected address.** The write guard keeps refusing a payload that drops a persisted id by accident, but an address corrected on the edited row is a new id for the same mailbox: the old id is released from the persisted set on purpose (`Accounts.withoutId`) before the save, so the correction lands and later saves are not blocked by the stale snapshot.
- `replaceAt` keeps the in-place, refuse-on-collision shape. `main`'s `replaceAt` folds a colliding edit the way `add` folds a re-added address; that fold is the deletion this PR is about, so its test now asserts the refusal and the untouched list. The selection tests from `main` pass unchanged.

`make test-js`, `make test-shell` and `make test-qml` pass on the branch. #108 can be closed in favour of this one.

gw7523:
Rebased onto `062f0f0`; no changes beyond the rebase. `make test-js`, `make test-shell` and `make test-qml` pass on the branch.

huacnlee (approved):

PR #117, opened 2026-09-06
Add a JMAP provider, so a mailbox that offers it threads and searches on the server

ArdonisLLC:
Fastmail already worked here through the `imap` provider, so this is not about
reaching it. It is about three things `Imap.js` documents that IMAP cannot do,
each of which JMAP answers directly.

`Imap.js` declares `threads: false` and falls back to `References`, which is what
every IMAP client has to do. Its search is a `TEXT` criterion the file itself
calls only "the closest standard equivalent". And it refuses `spam` on principle,
because moving a message to a folder teaches a generic server nothing and "a
'Report spam' button that quietly means 'move' is a promise the provider cannot
keep". JMAP gives every message a server-side `threadId`, filters `Email/query`
on the server, and — on a host known to learn from its Junk mailbox — makes that
button honest.

It is also faster in a way that is structural rather than tuned.
`GmailApiClient.qml` records the cost of the alternative: a list call plus one
metadata call per message, where "25 sequential round trips to Google is most of
a second of staring at an empty panel". A JMAP request is a list of method calls
in which a later call may name an earlier one's result, so a page is one round
trip — `Mailbox/get`, `Email/query`, and an `Email/get` that names the query's
ids. Measured against a real 2,171-message inbox: 25 rows with headers in 203 ms.

**The decision worth arguing about, made three times.** Everything above a
provider reads the shape Gmail's API returns, so a JMAP message is adapted into
it, with headers synthesised from the structured fields JMAP already sends. But
wherever actual MIME is needed, this uses the raw message rather than mapping
JMAP's `bodyStructure`: a body is downloaded by `blobId` and handed to
`Message.parseRfc822`, and an outgoing message is the one
`Message.buildRawMessage` already produced, uploaded as a blob and imported.
The alternative was building structured `Email` objects in both directions —
a second implementation of MIME that could disagree with the first. This costs
a round trip when opening a message, which is what IMAP costs too.

Sending is therefore upload → `Email/import` → `EmailSubmission/set`. The
message is imported *without* `$draft`, because a send that fails should not
leave something behind that looks like a draft nobody wrote, and
`onSuccessUpdateEmail` files the sent copy and clears `$draft` in the same
request, so there is no window where a sent message is still in Drafts.

**Push.** The session's `eventSourceUrl` is held open by `scripts/jmap-push.sh`
through curl, read a line at a time by `Process` + `SplitParser`. It does not
replace the poll: a server may expose no event source and a held connection
drops, so `pollTimer` is untouched and this only makes its work happen sooner.
`MailAccount` connects with `ignoreUnknownSignals`, so nothing asks which
provider it holds — the clients that cannot push simply never emit.

**The one change to a shared contract**, and the part most likely to be wrong.
`Registry.can()` answers per provider, but two of these answers are per
*account*: only the session knows whether a token was scoped to send, and only
the server knows whether it has an Archive mailbox or a Junk one that learns. So
the provider states a ceiling and the client withdraws what an account cannot do,
in `unsupported`, read by `MailAccount.accountRefuses()`. A client without that
property refuses nothing, so gmail, hey and imap are unaffected. If you would
rather that seam looked different, say so — it is the piece I am least sure of.

`web`/`webBox` stay false even for a known host: `Registry.webMessageUrl` takes a
provider id and a message id, with no account and therefore no host in scope, so
a generic provider cannot answer it for one host and not another.

Secrets follow the existing rule. The token lives in GNOME Keyring under
`kind=jmap-token`, `accounts.json` holds only host and username, and every curl
invocation takes its config on curl's own stdin through `scripts/curl-config.sh`
— which matters more here than elsewhere, because the push process lives for
hours.

## Verification

`make validate` green on d724154, run on Arch with Qt 6.11.2.

`tests/test_jmap.js` is new — 182 assertions over session parsing, the query
grammar, filter building, the message adapter, action plans, the send requests
and the push stream. Run against unmodified fd713ad it fails immediately
(`ENOENT: providers/JmapProtocol.js`), and `tests/test_provider.js` fails there
too on the provider list (`['gmail','hey','imap']` vs the expected four).

Live, against a real Fastmail account: a 2,171-message inbox paged in one
request in 203 ms with 23 distinct threads across 25 rows; a message body
downloaded raw and parsed to 835 chars of text and 12,339 of HTML with its
`List-Unsubscribe` found; a message sent end to end (upload → import → submit,
`undoStatus: final`) landing in Sent with `$draft` cleared; and a server-side
change producing a push wake in the same second, against a 120 s poll.

Driven by hand on an Omarchy 4.0.2 desktop: the whole add flow, from the
provider picker through JMAP to the address, server and token fields, to
Connect the mailbox — in a light theme and in a dark one, at three columns and
at mini. In each combination the setup page took the address, the server and
the token, and the mailbox connected and listed on the first attempt.

No screenshot is attached to this body. Say if you want one and I will add it.

A fresh agent reviewed the diff against `AGENTS.md` with no context from the
work. It found one high-severity fault and four smaller ones, all fixed in
0555e24:

- **The capability ceiling had a hole in it.** `Jmap.js` declares `archive`,
  `spam` and `send` as a ceiling and lets the client withdraw them per account,
  but `refuseUnavailableAction` — the guard `act()` runs *before* the optimistic
  update — consulted only `Provider.can`. For a JMAP account it could never
  refuse anything, so `e` and `s`, bound in every mail context, did what the
  hidden button would not have. Downstream it was worse: archiving on a server
  with no Archive mailbox left `moveTo` empty, `planIsEmpty` said true, and
  `applyPlan` reported success having sent nothing — the row left the list and
  the note said "Archived" for a request no server saw. That is the fault
  `AGENTS.md` records against HEY, reintroduced through the ceiling. Fixed in
  both places: a plan now records the role it asked for, so an unresolvable move
  is distinguishable from an empty one and is refused with a note naming the
  missing mailbox; and the guard consults `accountRefuses` as well. Both are
  covered by tests that fail without the change.
- `getAttachment` created a handle for its own callback guard but never threaded
  it into the fetch, so aborting suppressed the answer while the message
  download and blob fetch ran on. It now drives the real request, as
  `ImapClient.getAttachment` does.
- `scripts/jmap-push.sh` had no test. It is now a case in
  `tests/test_curl_config.py`; removing its `validate_config_fields` call makes
  that test fail on an accepted config injection.
- The token field set `echoMode` directly, overriding the `TextField`
  component's own `password` binding, and its `tokenVisible` property was never
  wired to anything. It now uses the component's `password` property and has the
  reveal control `ImapSetupPage` has.
- A `children` list on the request handle was never populated — every chain here
  is sequential — so it implied a cancellation shape this client does not have.
  Removed.

It also reported that the commit message carries no `## Verification` or
`## Release Notes` sections. Those belong in this body, which is where they are;
the reviewer had the diff but not this text.

What it checked and found sound, stated because a reviewer will ask: the
`inFlight` bookkeeping and the timeout-versus-abort race in `send()`; that the
token reaches no argv, log or error string; and that none of the seven shared
files regress gmail, hey or imap — `accountRefuses` is a no-op for clients that
declare no `unsupported`, and the new `Connections` block is
`ignoreUnknownSignals`.

## Release Notes

- A mailbox that speaks JMAP can now be added directly, instead of over IMAP.
  Fastmail is the one this was built and tested against.
- Conversations on such a mailbox come from the server rather than being
  reconstructed from `References`, and search runs on the server.
- New mail appears as it arrives rather than at the next poll, where the server
  offers a push connection.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -222,15 +222,25 @@ key. What matters while working:
 
 ## Providers
 
-- A mailbox is a **provider**: `gmail`, `hey`, or `imap`, listed in that order
-  because IMAP is the answer for a server the other two do not name and a
-  chooser that opened with it would ask the question backwards. `Provider.js` is
+- A mailbox is a **provider**: `gmail`, `hey`, `jmap`, or `imap`, listed in that
+  order because IMAP is the answer for a server the others do not name and a
+  chooser that opened with it would ask the question backwards. JMAP sits ahead
+  of it for the same reason: fewer servers speak it, so it is the more specific
+  answer. `Provider.js` is
   the only place that knows the differences — which mailboxes exist, what a query
   string means, what the service can be asked to do, and how it signs in.
   Nothing above it branches on a provider id.
 - Two objects make a provider work: something that signs in (`AuthManager`,
-  `HeyAuth`, `ImapAuth`) and something that fetches (`GmailApiClient`,
-  `HeyClient`, `ImapClient`).
+  `HeyAuth`, `JmapAuth`, `ImapAuth`) and something that fetches
+  (`GmailApiClient`, `HeyClient`, `JmapClient`, `ImapClient`).
+- A capability a provider declares is a **ceiling**, not a promise. `Imap.js`
+  declares `archive` true and lets the client find out whether the server has
+  anywhere to put it; `Jmap.js` does the same for `send`, `archive` and `spam`,
+  and its client withdraws what this account cannot do in `unsupported`, which
+  `MailAccount.accountRefuses` reads. Anything gating an action on a capability
+  must consult both — `refuseUnavailableAction` checked only the provider once,
+  and a key bound in every mail context then did what the hidden button would
+  not have.
   `MailAccount` builds one pair through a `Loader` and drives them through an
   identical interface — same method names, same arguments, same callback shape.
   Adding a provider is those two files and a registry entry.
--- a/App.qml
+++ b/App.qml
@@ -931,6 +931,21 @@ Item {
   }
 
   Component {
+    id: jmapSetupPage
+
+    JmapSetupPage {
+      service: root.service
+      textColor: root.foreground
+      dimColor: root.dim
+      dangerColor: root.danger
+      accentColor: root.accent
+      panelFontFamily: root.fontFamily
+      accountCount: root.service ? root.service.accountCount : 1
+      onRemoveRequested: root.removeCurrentAccountFromEditor()
+    }
+  }
+
+  Component {
     id: imapSetupPage
 
     ImapSetupPage {
@@ -1645,7 +1660,8 @@ Item {
             sourceComponent: root.showPicker
               ? providerPickerPage
               : (setup.kind === "imap" ? imapSetupPage
-                : (setup.kind === "hey" ? heySetupPage : gmailSetupPage))
+                : (setup.kind === "jmap" ? jmapSetupPage
+                : (setup.kind === "hey" ? heySetupPage : gmailSetupPage)))
           }
           }
         }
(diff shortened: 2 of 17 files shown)


huacnlee:
Jmail support is already have merged in main. Thanks for PR.

ArdonisLLC:
Excellent! Thank you for this amazing app. It is my favorite mail client, ever.

PR #119, opened 2026-09-06
Read the IMAP search ceiling with a FETCH curl passes through, and take from:, to: and subject:

gw7523:
## What changes

**Every search on an IMAP account answered nothing.** The interactive search opened with `UID FETCH *:* (UID)` to learn the newest UID before searching its first window. curl treats a FETCH of one message as a body fetch and delivers its untagged answer through the protocol-header callback, where the transport never looks, so the ceiling came back empty and the search never got a range. Verified against iCloud through `mail-transport.sh`: that one command returns only the greeting, while a numeric range returns the row.

The ceiling is now read in two short steps — the message count from STATUS on an unselected connection, as the unread badge already does, then `FETCH n:* (UID)`, which curl returns intact and which a message delivered between the two commands raises rather than escapes. The client takes the highest UID it returns. A STATUS with no count, a FETCH with no row, or a BAD to a range past the end (an expunge between the two) each fall back to the complete snapshot rather than answering "nothing" or raising a search error that would discard the cached preview.

**The typed search takes `from:`, `to:` and `subject:`.** With or without a space after the colon, with a quoted value keeping its spaces; other words stay one TEXT criterion. A `*` wildcard is dropped, since IMAP criteria already match substrings. An operator with no value is searched for as typed — which matches nothing — rather than dropped, which would have listed the whole inbox as a result and put Ctrl+A one keystroke from acting on it.

## Verification

- `make test-js`, `make test-shell` and `make test-qml` pass on this branch alone. New cases in `tests/test_imap.js` (`topUidCommand`) and `tests/test_provider.js` (the operators, quoting, wildcards, and empty values); `tests/test_source.sh` now refuses any `*:*` FETCH literal in the IMAP client and protocol.
- Exercised live against iCloud and an Office 365 mailbox over IMAP: a plain word, `from: name@example.com`, `from:*@example.com`, and `subject:` phrases all return rows; each was returning nothing before.

## Security

No new inputs reach the transport. The typed text is still quoted with `JSON.stringify`, which escapes exactly the two characters IMAP escapes, and `normalizeCriteria` folds the control characters it always did; an operator value is quoted the same way. Non-ASCII search terms still go as a quoted string, which is the pre-existing limit of this path and unchanged here.

## Release Notes

- Search works again on IMAP accounts; it had been answering nothing.
- `from:`, `to:` and `subject:` work in the search box on IMAP accounts, with or without a space after the colon.

### DIFF
--- a/providers/Imap.js
+++ b/providers/Imap.js
@@ -59,12 +59,37 @@ var MAILBOXES = [
 ]
 
 // IMAP SEARCH has no free-text operator that means what a user means by typing
-// words into a search box, so the text becomes a TEXT criterion — headers and
-// body, the closest standard equivalent. JSON.stringify is used for the quoting
-// because it escapes exactly the two characters IMAP escapes.
+// words into a search box, so plain words become a TEXT criterion — headers and
+// body, the closest standard equivalent. The three operators people bring from
+// webmail, from: to: and subject:, become the IMAP criteria of the same name;
+// a space after the colon is allowed because that is how they get typed, and a
+// quoted value keeps its spaces. IMAP matches a criterion as a substring, so a
+// `*` wildcard is dropped rather than searched for. An operator with nothing
+// after it, or nothing but wildcards, is searched for as typed — which
+// matches nothing — rather than dropped, which would either list the whole
+// inbox as if it were a result or quietly widen the search.
+// JSON.stringify is used for the quoting because it escapes exactly the two
+// characters IMAP escapes.
 function searchQuery(text) {
   var value = String(text === undefined || text === null ? "" : text).trim()
-  return value === "" ? "" : "folder:INBOX TEXT " + JSON.stringify(value)
+  if (value === "") return ""
+  var pattern = /(from|to|subject):\s*(?:"([^"]*)"|(\S*))|(\S+)/gi
+  var words = []
+  var parts = []
+  var match
+  while ((match = pattern.exec(value)) !== null) {
+    if (match[1] === undefined) {
+      words.push(match[4])
+      continue
+    }
+    var term = (match[2] !== undefined ? match[2] : match[3]).replace(/\*/g, "")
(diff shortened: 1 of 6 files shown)


gw7523:
Rebased onto `062f0f0`; no changes beyond the rebase. `make test-js`, `make test-shell` and `make test-qml` pass on the branch.

huacnlee (approved):
Reviewed head `35c4b4362e8ab27165757907ac7d72d3fef98a0d`. No blocking findings in the search-ceiling and operator changes.

`QT_QPA_PLATFORMTHEME= make validate` passed on a fixed copy of this commit: 445 QML results plus 4 Outlook HTTP results, JS/shell checks, qmllint and plugin validation. Seven additional real-ImapClient scenarios passed (9 QtTest results including setup/cleanup), covering numeric ceiling/progress, missing count, empty/BAD FETCH fallback, failure before/after a settled prefix, and cancellation. The numeric first search remains bounded to 4096 UID numbers; snapshot continuation retains bounded message windows on a reused connection.

Real curl 8.21.0 against a controlled loopback IMAP server returns the new STATUS/FETCH/search responses correctly. Twelve operator/quote/backslash/control-character/Unicode cases passed through the actual query builder and transport as a single SEARCH each, without an extra mutation command. Qt/QtTest 6.11.2; synthetic credentials only.

Security: PASS, scoped to the changed query and response-to-command boundaries. Fixed operator names and quoted values flow through existing criteria normalization, numeric UID bounds, base64 stdin transport and pre-curl config validation; no new origin or credential routing is introduced. This is not a complete transport audit or proof of live-provider compatibility; existing non-ASCII/control-escape search limitations remain.

Non-blocking follow-up: keep the new control-flow scenarios as maintained runtime tests. Also qualify the description's universal claim: the old command returned its UID on my controlled server, so I did not independently reproduce the reported iCloud/Office 365 failure. The author's live-provider evidence and this replacement-path verification are distinct.

PR #126, opened 2026-09-07
Move a row at the keystroke while its send waits for the slot

crwilson31108:
Deleting down a list on an IMAP account behind Proton Bridge emptied the list at the server's pace rather than the user's.

## Reproduction

1. Open a mailbox on an IMAP account served by Proton Bridge.
2. Press `d` on a row, then `d` on the next, then the next.

The first row leaves at once. Every row after it stays on screen, under a cursor that has already moved on, until the server has answered for the row before it. Bridge answers a `UID MOVE` to Trash in about two seconds, because it asks Proton before it says OK; reads settle in milliseconds. Measured against a local Bridge: a move took 1.9 s, an append 2.4 s, a header fetch of fifty rows 30 ms.

## Why it happens

`act` allows one mutation in flight, and since #107 a second action waits rather than being refused. But what waited was the whole action: the request went into the queue as a verb and an id, and the optimistic edit ran only when `runQueuedAction` handed it back to `act`. So the row moved when the slot freed, not when the key was pressed.

## The change

The optimistic edit is taken at the keystroke whatever is in flight. What the queue holds is the send: `dispatch` closes over the call and the rollback, and `runQueuedAction` runs it and nothing else. A callback that frees the slot sends the next queued action before it acts on its own answer, so no revalidation settles a state the next edit is not in yet. The `Qt.callLater` net in `onPendingActionChanged` is therefore unreachable and goes. `Model.enqueueAction` keeps the first send when a true repeat coalesces and no longer coalesces a quiet press with an explicit one: they differ in scope, note and rollback, so they queue in turn.

Every optimistic edit is an intent, held per summary until its server answers: the row, each member the rail draws, and the reader's copy, each with the state it started from and the function that made the edit. A success drops its intent and nothing else, since the summaries the later ones started from already hold it. A failure drops its intent and replays the ones behind it from where the failed one started, so what stays on screen is exactly the edits still waiting. `Model.rebaseIntents` is that replay, and `rowAfterMemberEdit` is the row recompute lifted out of `act` so it can run from more than one starting state. A true repeat that coalesces into an earlier send lets go of what it held, since nothing will ever answer it.

A refused removal goes back where the settled order says: `restoreRow` anchors on the lists as they stood before the first edit still in flight on that query, held while any is — the list the failed edit itself saw is not enough, because with two removals queued the second saw a list the first had already shortened.

`account/Intents.qml` holds that bookkeeping and the one restore path every caller shares, beside the account file the way `BatchAction.qml` and `Rsvp.qml` are: `MailAccount.qml` was a few bytes under the 128 KiB ceiling in `tests/test_source.sh` and now has room.

## Since the last review (head `4a279dc`, rebased on main `7f7c1f0`)

- **Batch actions are on the same queue.** `BatchAction.run` takes its optimistic edit at the keystroke whatever is in flight, queues one send for the lot when a send holds the slot (its rows and verb coalesce like any other request), and its completion sends the next queued action before acting on its own answer. Both interleavings — an action then a batch, a batch then an action — are covered.
- **Bulk refusals settle through the intents.** Mark-all and the batch hold an intent per row, member and reader's copy under one token. A refusal replays each row over whatever was taken on it since, so a star pressed while mark-all or a batch was still deciding stays. A batch answered per row settles only its refused rows; one refused as a whole settles every row and reads the list again, as before. The list is built once and assigned once, on screen or in the cache.
- **The cached query is rebased, not overwritten.** A refusal answered after navigating away repairs the cached copy of the query the edit was taken on with the same rule as the list on screen — the row replaced if listed, put back in the settled order if this edit removed it — rather than writing the list that edit saw. Two removals refused in turn leave the cache holding both rows, in order.

## Screenshots

Synthetic mail (example.com senders), the real `App` and `Service` rendered offscreen at 1100×700 with the test stubs' theme, against a client whose answers are held until the harness releases them. Each pair is the same moment of the same key sequence on main `7f7c1f0` (before) and on this head (after), so the two halves differ only in what the change does. Hosted on a branch of my fork (`crwilson31108/omamail` at `pr126-screenshots`), not in this tree; I can move them to the attachment host if you prefer.

**Pending rapid actions.** Cursor on the first row, `d` three times, nothing answered yet. Before: one row gone, the other two still listed under a cursor that has moved on to the fourth row. After: all three gone.

Pending rapid actions, before and after

Once every answer is in, both end the same way:

Rapid actions settled, before and after

**Failed-action restoration.** `d` on "Notes on the analytical engine", then `d` on "Your invoice #20419"; the server will refuse the first. Nothing answered yet — before: only the first row gone; after: both gone.

Two presses pending, before and after

The server refuses the first. Both put "Notes…" back in its place, between "Compiler retreat…" and "Re: Machine intelligence…", and show the refusal in the status row. What differs is what the second press has done by then: before, its row leaves only now, as its turn comes; after, it left at the keystroke and its send is already in flight.

First refused, before and after

The second answer accepted: "Your invoice…" gone, "Notes…" where it was, in both.

Restoration settled, before and after

The list before any press, for reference:

The list before any press, before and after

## Verification

`make validate` green on `4a279dc` rebased on `7f7c1f0`: node, shell (portable and libcurl), 497 QML tests, 4 native HTTP checks, 64 sidebar checks, qmllint exit clean, `omarchy plugin validate`, `git diff --check`.

`tests/qml/tst_queued_actions.qml`, against the real `MailAccount` with the provider replaced through its Loader, now runs 23 cases. The seven added for this round, of which six fail on the previous head `4b30e5b`:

- mark-all refused with a star queued behind it: the read comes off, the star stays, and the accepted star is what the row says; nothing held afterwards.
- mark-all refused in the Unread view: every row comes back, in order.
- two removals refused after navigating away: the cached copy of the original query holds `[one]` after the first and `[one, two]` after the second; the list on screen is untouched.
- a batch taken behind an action: its row leaves at the keystroke, its send waits, and the action's answer sends it (this threw on the previous head).
- an action taken behind a batch: the batch's answer sends it (this never drained on the previous head).
- a queued batch trash answered per row, one refused: that row is back in the settled order, the accepted ones gone, and the note counts it.
- a batch mark-read refused as a whole with a star queued behind it: the read comes off, the star stays.

`tests/test_model.js` covers `listAfterRestore` beside `restoreRow` and `previewAfterRestore`. `tests/test_source.sh` pins the new shape: the batch queues its own send and drains the queue from its callback, both bulk paths settle through `intents.restore`, and mark-all no longer assigns a snapshot of the list.

The refused-request paths are exercised by the runtime regressions rather than by hand, since a live server does not refuse on demand. No live Proton Bridge test was run for this revision.

## Release Notes

- Deleting, archiving or moving several messages in a row now moves each row the moment you press the key, even on a server that takes seconds to confirm. The presses still reach the server one at a time, and a refused one puts its row back where it was, keeping any change you made to it since.


account/MailAccount.qml
```diff
@@ -1696,7 +1633,7 @@ Item {
           && !root.deferredLoadCleared(actionQuery)) {
         root.nextPageToken = actionToken
         root.messages = removed
-          ? root.messages.slice(0, index).concat([before], root.messages.slice(index))
+          ? Model.restoreRow(root.messages, before, beforeMessages, index)
           : Model.replaceById(root.messages, before)
```
huacnlee: [P2] Preserve later optimistic changes when an earlier action fails

Every queued edit now runs immediately, but this rollback still restores the entire pre-action row. In the real MailAccount fixture: star `one`, immediately mark it read, fail star, then succeed markRead. The read request is dispatched, but the first rollback restores the old unread snapshot; the successful callback does not reapply read state. The final row remains unread. The inverse sequence—failed markRead followed by successful star—loses STARRED. Both regressions pass the base and fail this head.

For surviving rows in ordinary Inbox there is no automatic list revalidation to repair this, and rememberList can retain the stale state. Roll back only the failed operation while preserving/rebasing later intents; cover member summaries and selected-message rollback as well as the list. Add runtime regressions for both overlapping-operation sequences.

account/Model.js
```diff
@@ -591,6 +594,20 @@ function removeById(list, id) {
...
+// Its index is stale once the rows queued behind it have gone.
+function restoreRow(list, row, beforeList, index) {
+  var source = Array.isArray(list) ? list : []
+  var beforeRows = Array.isArray(beforeList) ? beforeList : []
+  var held = Number(index) || 0
+  for (var i = held + 1; i < beforeRows.length; i++) {
+    var at = indexById(source, beforeRows[i] ? beforeRows[i].id : "")
+    if (at >= 0) return source.slice(0, at).concat([row], source.slice(at))
+  }
+  var clamped = Math.max(0, Math.min(held, source.length))
```
huacnlee, line 773: [P2] Preserve ordering across multiple optimistic removals

Start with `[one, two]`, trash `one`, immediately trash `two`, then fail both requests. The first failure restores `[one]`. The second action captured `[two]` at index zero after the first optimistic removal, so it has no surviving old follower and this fallback inserts it before `one`. Final order is `[two, one]` instead of `[one, two]`.

This is the PR's rapid-delete workflow when moves are refused. The controlled QML regression passes the base and fails this head. Preserve ordering information from before overlapping removals, or reconcile against an authoritative ordering; a follower anchor from each already-shortened snapshot cannot recover the original order.

huacnlee (changes requested):
Reviewed head `739279e55fbe3c05b080b7988d8ff8a7d9109575`. Immediate optimistic edits improve the rapid-action path, but two rollback regressions remain, detailed inline. Three focused QML probes pass base `062f0f0` and fail this head; I independently reran the head probes and confirmed the same failures.

Validation: `QT_QPA_PLATFORMTHEME= make validate` passes (446 QML results plus 4 native Outlook HTTP results, JS/shell suites, qmllint and plugin validation). The focused head harness reports 9 passed / 3 failed; the base comparison reports 5 passed / 0 failed, including setup/cleanup. Tests use real MailAccount with controlled provider completions and the repository fixture's list/count/cache hooks, Qt/QtTest 6.11.2, offscreen. No live mailbox, real credential or Proton Bridge session was used. These demonstrate state transitions, not live-provider delivery or persistent disk contents.

The request queue remains serialized and in memory; this review does not independently confirm the reported live Bridge timing. The main functional gap is reconciliation after requests fail while later edits have already changed the same state.

Security: PASS, scoped to this change. Existing message/member IDs and source-label context are captured into dispatch closures and passed through existing capability guards and provider methods. No new credential, URL, shell, HTML or persisted executable boundary is introduced. This is not a complete application/transport audit. The functional failures still block approval.

account/MailAccount.qml
crwilson31108: Fixed in 4b30e5b. Every optimistic edit is now an intent held per summary — the row, each member the rail draws, and the reader's copy — until its server answers. A failure drops only its own intent and replays the ones behind it from where the failed one started (`Model.rebaseIntents`), so the star comes off and the read stays, in either order, and the rail and reader say what the list says. The row's recompute is lifted into `Model.rowAfterMemberEdit` and reads the members as the rail holds them at replay time, which is what keeps a rail action's block in step after a refusal.

Runtime regressions in `tests/qml/tst_queued_actions.qml`: both of your sequences, the same with the row in the rail's summaries and open in the reader, two rail actions on one member with the first refused, a coalesced repeat, and a refusal answered after navigating away. All fail on 739279e and pass on this head.

account/Model.js
crwilson31108: Fixed in 4b30e5b. `restoreRow` now anchors on the lists as they stood before the first edit still in flight on that query, held by the account while any is, rather than on the list the failed edit itself saw: before the first surviving follower, else after the last surviving predecessor, else at its index. `[one, two]`, trash one, trash two, both refused, comes back as `[one, two]`; covered in `tst_queued_actions.qml` and, for the anchoring rules, in `tests/test_model.js`.

crwilson31108:
@huacnlee ready for another look at `4a279dc`, rebased on main `7f7c1f0`. The three findings are addressed and covered by seven new regressions in `tests/qml/tst_queued_actions.qml` (six fail on `4b30e5b`), and the description now carries labeled before/after screenshots for pending rapid actions and failed-action restoration, rendered offscreen with synthetic mail. Details in the updated description under "Since the last review".

huacnlee (changes requested):
Reviewed 4b30e5b2f2ff8350cc0fa2a4520bca2855901780. The previous findings are repaired, but the following blockers remain.

- **P1 — Integrate the dispatch queue with current main's batch actions** (`account/MailAccount.qml:1418–1449`). The clean merge with main 174846d still has BatchAction enqueue requests without `dispatch`, so completing a preceding individual action throws. In the opposite order, batch completion never drains an individual action queued behind it. Both synthetic probes pass current main and fail the merge. Update every queue producer/completion and cover both interleavings.
- **P2 — Preserve later edits when bulk-read fails** (`account/MailAccount.qml:1982–1987`). Start markAllRead, immediately star a message, reject bulk-read, then accept star: the star disappears. This probe passes the PR base and fails this head. Bulk-read's snapshot rollback must participate in intent reconciliation.
- **P2 — Reconcile the original cached query after navigation** (`account/MailAccount.qml:1681–1689`). Trash one and two, navigate away, then reject both: the cached original list contains only two. The second shortened snapshot overwrites the row restored by the first failure. Rebase or invalidate the cached query instead. This was verified against the real in-memory CacheStore with mocked disk/process boundaries, not a real disk write.

**UI evidence: BLOCK.** The PR description must include clearly labeled, comparable before/after screenshots for pending rapid actions and failed-action restoration, with synthetic/redacted data and a visible-difference explanation. A short clip can additionally demonstrate the timing.

`make validate` passes on the standalone head (453 QML tests, native HTTP/sidebar checks, JS/shell/lint/plugin validation), and the existing focused suite passes. The additional reproductions above expose uncovered failures. No live Proton Bridge/mailbox tests were run.

**Security: PASS** for the affected in-memory selection/cache and existing provider boundaries. This is not a complete security audit. Product decision: not applicable; the responsiveness change restores intended behavior.

Use GitHub-hosted screenshots with synthetic/redacted data, matching window size, theme, scale and corresponding interaction state. Include a concise visible-difference explanation in the PR description. Screenshots must reflect the current PR head and be updated after UI fixes. Missing screenshots or difference explanations block approval.

Publication check: main has advanced to 68e4243. The integration evidence above was gathered against 174846d; full validation was run on the standalone PR head. Revalidate after rebasing on current main.

PR #129, opened 2026-09-07
Park every send in a queue with its own undo window, and say what is still in flight

gw7523:
## What changes

**A second send no longer waits for the first.** One parked send was all there was, and a second was refused with "Another message is waiting to be sent". A reply written while the last one is still counting down is the ordinary way a morning's mail gets answered, and the undo window is exactly what makes the refusal likely. Each account now keeps a queue of parked sends, oldest first, each with its own moment to go.

**One at a time on the wire, in order.** One timer per account waits for whichever send is due first. A send in flight holds the rest until it answers, so the account's single `sending` slot still means what it says and a provider sees the messages in the order they were written. A send with no undo window still goes at once when nothing is ahead of it; behind another it waits its turn.

**Undo takes back the newest, and the right draft comes back.** `undoSend` answers with the send's name. The composer parks one draft per send beside that name — `Service.send` hands the names out (`send-N`), so two accounts' sends never share one — and `Alt+Z` restores that draft, not another. A failed send restores its own draft by the same name. The toast reads "3 queued · newest in 9s" and its button "Undo last" past one; the Send button no longer greys out while something is parked.

**Nothing is lost on the way.** Every parked draft rides along in the compose recovery file, so a restart inside several undo windows keeps them all, and recovery goes on holding them until nothing is parked. A removed account hands its parked sends back to the composer with a note instead of dropping them.

**The status line says what is still owed.** "Sending 2 · 1 queued to send · 1 action running · 3 waiting", each phrase only while true, nothing when idle. It reads the row-action line from #107 for the last two.

## How

- `account/SendQueue.qml` is the queue, beside the account (which is within a few hundred bytes of the 128 KiB ceiling): the parked list, the two timers, `park`, `deliverDue`, `deliverAll`, `undoLatest`, `abandon`. `MailAccount` keeps `send` (the payload and its checks), `deliver`, and the readings the window draws: `sendPending`, `sendPendingCount`, `latestSend`, `sendSecondsRemaining`.
- `replySent` and `replyFailed` carry the send's name; `ComposeView.takeParked` finds the draft by it, falling back to the newest for an undo and the oldest for an answer only when a caller did not name the send (the test stubs), and never splicing on a named miss.
- `Model.activityStatus` is the pure strip text; `Recovery.serialize` takes the other parked drafts.

## Verification

- `make test-js`, `make test-shell` and `make test-qml` (366 QML results) pass on this branch alone. New coverage: `tst_send_failures` parks a second send from another account and undoes the newest, leaving the older parked; `tst_undo_send_toast` for the counted wording; `test_model.js` for the strip; `test_compose_recovery.js` for the parked list.
- Exercised live on an IMAP and an Office 365 account: two sends back to back with Alt+Z inside the window, a switch between accounts before the undo, a zero-delay pair, and a run of `Shift+U` then `d` on five ticked rows while the strip reads.

## Security

No new inputs reach a provider or the transport. Send names are the service's own counter and never leave the process; payloads are built exactly as before.

## Release Notes

- Several messages can be sent in a row without waiting for the undo window of the last one; each keeps its own undo, and Undo takes back the newest.
- The status line shows what is still in flight: sends parked or going, actions running or waiting.


gw7523:
Rebased onto `062f0f0` (the All Mailboxes view and preview-on-cursor landed between the branch and the PR). The naming now rides on `main`'s routed `send`: the From address picks the mailbox, that mailbox's line parks the send, and the service hands out the name; a submission naming a mailbox that is gone is still refused with `false`. The counts sit beside the merged-view status readings. `make test-js`, `make test-shell` and `make test-qml` pass on the branch.

huacnlee (changes requested):
Reviewed b85882e658a6ed2611974408895f5d99487f85e2. The queue and activity UI are accepted product scope, but the following blockers remain.

- **P1 — Preserve every draft when removing an account** (`Service.qml:420–422`, `components/ComposeView.qml:695–700`). Removing an account with three queued messages restores each synchronously into a composer with one interrupted-draft slot. Only the newest two drafts survive; the oldest is lost. Keep and persist the complete unsent set before releasing queue ownership.
- **P1 — Persist recovered drafts on every subsequent recovery write** (`App.qml:73–93`). Recovery loads extra drafts into `recoveryDrafts`, but later saves serialize only `parkedDrafts`. Recover two drafts, edit/save the first, then restart again: the second draft has disappeared from recovery. Cover repeated restarts and transitions between recovered drafts.
- **P2 — Relay each failure once and preserve unified account context** (`Service.qml:1670–1671`). The duplicated relay emits two failure signals and unconditionally switches active account even in unified mode. Keep the existing guarded switch and one signal; assert both event count and unchanged active account.
- **P2 — Preserve queue order when undo delay changes** (`account/SendQueue.qml:69–72`). Queue a message with a 30-second delay, change the delay to zero, then queue another: the newer send overtakes the first. Wait on the oldest entry as promised, or explicitly revise that contract.

All four findings were reproduced with actual QML, synthetic drafts and mocked external processes. `make validate` passes on the standalone head (446 QML tests plus native HTTP, JS/shell/lint/plugin checks); these are missing regression cases.

Resolve current-main merge conflicts in App.qml, Makefile, account/Model.js and tests/test_model.js and revalidate the resulting head.

**UI evidence: BLOCK.** The PR description must include clearly labeled, comparable before/after screenshots and visible-difference explanations for queue toast/count/Undo last, Send-button state, and the activity strip, at relevant widths.

**Security: PASS** for affected confidentiality/input/transport boundaries. Draft loss remains a functional data-integrity blocker; this verdict is not a complete security audit. Product decision: accepted.

Use GitHub-hosted screenshots with synthetic/redacted data, matching window size, theme, scale and corresponding interaction state. Include a concise visible-difference explanation in the PR description. Screenshots must reflect the current PR head and be updated after UI fixes. Missing screenshots or difference explanations block approval.

Publication check: main has advanced to 68e4243. The integration evidence above was gathered against 174846d; full validation was run on the standalone PR head. Revalidate after rebasing on current main.

gw7523:
Rebased onto `68e4243` (#116 and #148 underneath); no change of its own. `QT_QPA_PLATFORMTHEME= make validate` passes on this head.

gw7523:
Rebased onto v0.8.2 (`27c1b4a`, after #151); the head is now `45abadc`. The commit's content is unchanged — `git range-diff` shows context lines only. `QT_QPA_PLATFORMTHEME= make validate` passes on the new head.

PR #133, opened 2026-09-07
Find a CalDAV server's calendars from its address

gw7523:
## What changes

**A CalDAV server address is enough to add its calendars.** A calendar could only be added by its exact collection address, and the calendar query runs against whatever address a source holds. Typing the server — `https://caldav.icloud.com/`, the address every provider's help page gives — saved a source that asked the server's root for events, got none, and left the calendar empty with nothing to say why.

Adding a calendar now starts with **Find calendars**: a PROPFIND at the typed address for `current-user-principal`, one at the principal for `calendar-home-set`, and one at depth 1 on the home for what it holds. A typed address that is itself a calendar answers the first step by saying so and is added as it is, under the typed name if one was given. The page lists what was found, a switch beside each, and adds the ticked ones in one write with the one password stored under each calendar's own id. Left out, because none of them answers a calendar query with events: a subscribed feed (`CS:subscribed`), a collection whose component set holds only tasks (a reminder list), the server's inbox and outbox, and the home itself.

## How

- `scripts/calendar-transport.sh` takes an optional fourth field naming the method: `report` (the default, unchanged), `propfind-0` or `propfind-1`. It is one of three words, never a field curl is handed; anything else is refused.
- `calendar/Calendar.js`: the three request bodies, `davResponses`, `resolveDavHref` (absolute as given, a path from the asked address's origin, anything else beside it), `isCalendarResource` (the CalDAV element under any prefix, self-closed or not), `caldavSelf`, `davHome`, `caldavCollections`. All pure and covered in `tests/test_calendar_feed.js` with iCloud's own shapes: prefixed elements, a relative principal, an absolute home on a per-user host, and a listing that mixes the home, a reminder list, the inbox, a subscribed feed and two calendars.
- `CalendarController`: `discoverCalDav(url, username, password)` runs the steps through a process of its own and answers with `calendarsDiscovered(list, error)`; the credentials are held for those requests and no longer. `addCalDavCalendars(list, password)` saves several sources in one write and stores the password under each id in turn; `addCalDavCalendar` is now the one-item case of it.
- `CalendarSettings`: the add form's button is **Find calendars**; the found list appears with a switch each and an **Add N calendars** button. The existing single-address path still works: an address that is one calendar is added the way it always was.

## Verification

- `make test-js`, `make test-shell` and `make test-qml` pass on this branch alone. New coverage in `tests/test_calendar_feed.js` (the parsers, against iCloud-shaped answers) and `tests/test_calendar_transport.sh` (the PROPFIND methods and the refusal of any other).
- Exercised live against iCloud: `https://caldav.icloud.com/` with an app password finds the principal, the per-user home on `pNN-caldav.icloud.com`, and the calendars; the five real ones were added and answer this week's events, the reminder lists and subscribed feeds were left out.

The maintainer also ran `QT_QPA_PLATFORMTHEME= make validate` on cbc1c69: 445 main QML tests and four native HTTP tests passed, alongside JS/shell checks, qmllint and plugin validation. These general checks do not cover all discovery and save-failure boundaries. Comparable before/after screenshots and a visible-difference explanation are still missing.

## Security

The typed address still has to be HTTPS. The three PROPFIND bodies are fixed strings; the method is one of three words checked before anything is written to curl's config. Every field still passes `validate_config_fields` as before. Credential-destination and server-provided display-content handling still require corrections and verification; this revision does not have a passing security review. The password is held for the discovery requests and cleared when they answer.

## Release Notes

- Adding a CalDAV calendar takes the server address: Omamail finds the calendars it holds and adds the ones you pick.

### DIFF
--- a/calendar/Calendar.js
+++ b/calendar/Calendar.js
@@ -121,6 +121,107 @@ function caldavResponses(xml) {
   return out
 }
 
+// ------------------------------------------------------------ discovery
+//
+// A CalDAV server's calendars are found in three steps, each a PROPFIND: the
+// address the user typed says who they are (current-user-principal), the
+// principal says where their calendars live (calendar-home-set), and the home
+// lists them one level down. A typed address that is itself a calendar
+// answers the first step by saying so, and needs no other.
+
+function propfindPrincipalBody() {
+  return '<?xml version="1.0" encoding="utf-8"?>'
+    + '<d:propfind xmlns:d="DAV:"><d:prop>'
+    + '<d:current-user-principal/><d:resourcetype/><d:displayname/>'
+    + '</d:prop></d:propfind>'
+}
+
+function propfindHomeBody() {
+  return '<?xml version="1.0" encoding="utf-8"?>'
+    + '<d:propfind xmlns:d="DAV:" xmlns:c="urn:ietf:params:xml:ns:caldav"><d:prop>'
+    + '<c:calendar-home-set/>'
+    + '</d:prop></d:propfind>'
+}
+
+function propfindCollectionsBody() {
+  return '<?xml version="1.0" encoding="utf-8"?>'
+    + '<d:propfind xmlns:d="DAV:" xmlns:c="urn:ietf:params:xml:ns:caldav"><d:prop>'
+    + '<d:displayname/><d:resourcetype/><c:supported-calendar-component-set/>'
+    + '</d:prop></d:propfind>'
+}
+
+function davResponses(xml) {
+  var input = String(xml || "")
+  var pattern = /<(?:[A-Za-z0-9_-]+:)?response(?:\s[^>]*)?>([\s\S]*?)<\/(?:[A-Za-z0-9_-]+:)?response>/gi
+  var out = []
+  var match
+  while ((match = pattern.exec(input)) !== null) out.push(match[1])
+  return out
+}
+
+// An href the way the server wrote it, made absolute against the address it
+// was asked at: a full URL as it is, a path from that address's origin, and
+// anything else beside it.
+function resolveDavHref(baseUrl, href) {
+  var base = String(baseUrl || "").trim()
+  var value = String(href || "").trim()
+  if (value === "") return ""
+  if (/^https?:\/\//i.test(value)) return value
+  var origin = base.match(/^(https?:\/\/[^\/]+)/i)
+  if (!origin) return ""
+  if (value.charAt(0) === "/") return origin[1] + value
+  return base.replace(/[^\/]*$/, "") + value
+}
+
+// Whether a resourcetype block says "calendar": the CalDAV element, under any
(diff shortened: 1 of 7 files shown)


huacnlee (changes requested):
Changes are required before approval of cbc1c69.

- P1 — Credential destination authorization (calendar/CalendarController.qml:431–439; calendar/Calendar.js:165–173): enforce the configured credential scope throughout discovery and before persisting discovered sources. Correct URL resolution and require explicit, narrowly scoped authorization for any legitimate cross-origin deployment. Add controlled regressions asserting that unauthorized destinations receive no request or credentials. Synthetic local tests confirmed a boundary failure with the actual controller and real TLS/curl transport; sensitive reproduction details are retained locally.
- P1 — Server-provided display content (components/CalendarSettings.qml:283–289): render discovered names explicitly as PlainText, extend the source guard, and add a real Qt regression asserting no resource request. Runtime testing confirmed that the current display boundary is unsafe; no payload is included here.
- P1 — Discovery/form ownership (components/CalendarSettings.qml:saveCalendar and onCalendarsDiscovered): bind results and credentials to the same immutable operation, invalidate pending results when the form changes or is cancelled, and refuse stale saves. A controlled QML test confirmed that current form credentials can be associated with an earlier discovery result. Cover late success/error, changed credentials and repeated attempts.
- P2 — Failed-save cleanup (calendar/CalendarController.qml:sourceWriter.onExited): clear sourcesBeingSaved on configuration-write failure and all terminal paths. The retained queue causes a subsequent unrelated operation to start an empty-password store, report failure after configuration success, and skip refresh. Add failure-then-remove/enable/color/retry regressions.
- P2 — DAV component parsing (calendar/Calendar.js:214): accept legal XML attribute quoting and ordering when checking supported components. A task-only collection using single-quoted attributes is currently offered as an event calendar. Cover valid XML variants and retain the exclusion of task-only collections.
- P2 — UI evidence: add comparable, labeled before-and-after screenshots and a visible-difference explanation to the description. Use synthetic data, matching theme/size/scale/state, and cover the add form, discovery loading/error, selection and saved states on the final revision.

Also bound discovery response bytes and the resulting UI collection count. The existing transport deadline limits elapsed time, but does not bound buffered data or generated delegates; no denial-of-service exploit is claimed.

Local make validate passed: 445 main QML tests, four native HTTP tests, JS/shell checks, qmllint and plugin validation. Targeted controller, transport and real Qt probes nevertheless confirmed the findings above. Tests used synthetic credentials and controlled local endpoints on Qt 6.11.2 and curl 8.21.0; live-provider delivery was not tested.

Security: BLOCK. UI evidence: BLOCK. Product decision: pending independently of these technical findings. Server-address discovery and multi-calendar selection simplify CalDAV setup, but add credential-policy and asynchronous-state costs; I recommend introducing that scope only after these fixes and the product owner's acceptance. Scope acceptance does not replace security verification.

PR #134, opened 2026-09-07
Copy an event's location, and open a place on a map

gw7523:
## What changes

An event's location was shown and, when it was a link, opened. A place — a street address, a venue name — could only be read off the screen. The event detail gains two buttons beside the existing ones:

- **Copy location**, for any location: the text as written goes to the clipboard through `wl-copy` as one argument, with no shell in between, the way the address menu already copies an address.
- **Open in Google Maps**, for a location that is not a link: a maps search for the text as written, opened in the browser. A link keeps "Open location" and does not get this one.

## Verification

`make test-qml` and `make test-shell` pass on this branch alone. Exercised live on a fork carrying this branch, on iCloud events: a street address copies to the clipboard and opens as a Google Maps search; an event whose location is a link keeps Open location and gains only the copy.

## Security

The location text is user data from the calendar. It reaches the clipboard as a single argv entry and the browser as a URL-encoded query on a fixed host; neither passes through a shell.

## Release Notes

- The event detail can copy a location, and open a place on Google Maps.

### DIFF
--- a/components/CalendarEventDetail.qml
+++ b/components/CalendarEventDetail.qml
@@ -1,5 +1,6 @@
 import QtQuick
 import QtQuick.Controls
+import Quickshell
 import qs.Commons
 import qs.Ui
 import "../calendar/Calendar.js" as Calendar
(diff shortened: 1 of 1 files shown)


huacnlee:
Visual evidence required for this UI change. Please add clearly labeled before-and-after screenshots to the PR description, using synthetic or redacted data and GitHub-hosted attachments. Compare the same view, interaction state, window size, theme and scale, and explain the visible differences. Cover event detail with a place and with a URL, including Copy location and Open in Google Maps. Update the images after any further UI changes.

PR #135, opened 2026-09-07
Start a new calendar from a provider card

gw7523:
Stacked on #133 (CalDAV discovery); only the last commit is this PR.

## What changes

Adding a calendar began with an empty form and the question every provider answers differently: which address, which username, and where the password comes from. It now begins with **a card per provider**: iCloud, Fastmail, Nextcloud, Yahoo, Zoho, GMX, mailbox.org, Posteo, any other CalDAV server, or Google.

A CalDAV card prefills the address where the provider has a fixed one, hints it where the user's own server supplies it (Nextcloud's `/remote.php/dav/`), names the username and password in the provider's own words, says where its app password is made with a link to that provider's page, and ends in the same **Find calendars** discovery from #133. Google's card says only that its calendars arrive with a Gmail mailbox, which is what happens. Every address stays editable: a preset is a starting point, not a lock.

## How

- `calendar/Providers.js`, pure and node-tested: the card table, `formFor(id)` for the form's starting values, `isHelpUrl` for the help links.
- `CalendarSettings`: a Flow of cards behind **Add a calendar**; picking one shows the form with the preset applied, a note, the help link, and Back. The form's own placeholders are unchanged, so the existing source guards still hold; a card's words for the fields sit under them as a hint.
- Microsoft 365 / Outlook.com and Proton are not cards yet: each needs a source kind of its own (Graph calendars; read-only subscriptions), and each will bring its card with that kind rather than ship a card that leads nowhere.

## Verification

`make test-js` (new `tests/test_calendar_providers.js`: unique ids, known kinds, HTTPS starting addresses, help links only on the named hosts), `make test-shell` and `make test-qml` pass on this branch alone. The discovery the iCloud card ends in has been exercised live against iCloud (see #133); a run through the card itself on a fork carrying this branch is next, and this line will say how it went before the PR is ready for review.

## Security

The help links open only on the named providers' own HTTPS hosts, checked both in the table's test and again at click. Nothing else changes what the transport is asked.

## Release Notes

- Adding a calendar starts from a provider card that fills in the address and says where the app password is made.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,6 +83,7 @@ test-js:
 	node tests/test_calendar.js
 	node tests/test_calendar_cache.js
 	node tests/test_calendar_feed.js
+	node tests/test_calendar_providers.js
 	node tests/test_calendar_sources.js
 	node tests/test_calendar_palette.js
 	node tests/test_bar_preview.js
--- a/calendar/Calendar.js
+++ b/calendar/Calendar.js
@@ -121,6 +121,107 @@ function caldavResponses(xml) {
   return out
 }
 
+// ------------------------------------------------------------ discovery
+//
+// A CalDAV server's calendars are found in three steps, each a PROPFIND: the
+// address the user typed says who they are (current-user-principal), the
+// principal says where their calendars live (calendar-home-set), and the home
+// lists them one level down. A typed address that is itself a calendar
+// answers the first step by saying so, and needs no other.
+
+function propfindPrincipalBody() {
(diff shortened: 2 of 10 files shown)


huacnlee:
Visual evidence required for this UI change. Please add clearly labeled before-and-after screenshots to the PR description, using synthetic or redacted data and GitHub-hosted attachments. Compare the same view, interaction state, window size, theme and scale, and explain the visible differences. Cover provider cards, the selected provider form, and calendar discovery results. Update the images after any further UI changes.

PR #139, opened 2026-09-08
Suggest the calendar events the agent finds in a message

gw7523:
With **Suggest calendar events from mail** on in Settings — off until it is — a message from a person that names a time or a date is handed to the system AI once when opened, in the background. What it finds folds into one line above the message, "The agent found N events in this message", which opens on a click into a card per event: what, when, where, with **Add** and **Dismiss**. Add opens the calendar's event composer with the fields filled in, so the owner chooses the calendar and looks the times over before anything is written; a written event waves its suggestion away, a dismissed one stays away for the session.

This is the original #139 carried onto the native agent stack that #174 put under the panel: the same behaviour and tests, with the Python runner's part now in `src/agent/events.rs`. Co-authored with @gw7523.

## Before the model

A look is one AI call with the message in it, so the gates come first and cost nothing — the setting; mail from a person rather than a machine or a list (`noreply`/`notifications@`/`mailer-daemon`-style senders, Gmail's Promotions/Updates/Forums/Social categories, and any message with a `List-Unsubscribe` header are skipped); a time or a date in the subject or the text (`Agent.mentionsDate`: a clock time, a month with a day, a numeric date, "tomorrow", or a weekday bound to a plan — a bare "Sunday" in prose is a word); a message from the last two months; no look at that message yet; at most two looks running at once. The worker sends at most the first 8,000 characters. The switch says in three lines that it uses the system AI and spends tokens.

## How it runs

- The suggester reads the message through the same account-bound `agent.context` as an ask and starts the job with `events: true`; Rust records it as kind `events`, runs it at the `haiku` model with its own fenced prompt (rules, then every line of the message behind `| ` between two fence lines and nothing after the closing one), and on completion reads the last JSON array out of the answer into a bounded, validated `events` record on the job — at most ten, strings cleaned and cut to size, times as epoch milliseconds, whole days midnight to midnight in the local zone. No array, or dates that do not parse, is no event and no failure. The rules ask for one event per occasion: a deadline is the moment it falls due, not also the day after or the announcement.
- `agent.jobsProjection` leaves a look out of `byMessage`, attention and the panel's scopes — no row glyph, no glow, no history entry — and returns it under `eventLooks` by account and message with `activeEventLooks`. A failed or cancelled look is not held, so the message may be looked at again. That is the reason for **API revision 2** in `backend-api.json`, with a contract case for the look projection.
[...]

PR #140, opened 2026-09-08
Type into the account switcher to find a mailbox by name or address

gw7523:
Stacked on #139 (suggested events), which sits on #115. The account switcher (`Alt+A`) listed its rows and answered digits and the cursor keys; with four mailboxes that is a scroll and a count. It now opens with a search line on top: letters narrow the rows to the mailboxes they name — by the name they were given or their address, matched loosely and best first — the arrows (or Ctrl with `j`, `k`, `n` or `p`) walk what is left, and Enter opens the row the cursor is on, reported by its place in the full list so a narrowed list still opens the right thing. A digit is a letter here, as is a bare `j`, `k` or `o` — it has to be, for "jack" to find a mailbox; the combined "every mailbox" row stands aside while anything is typed; typing the query away rests the cursor on the mailbox in use again; a key that reaches the popup's column is forwarded to the search line. docs/KEYS.md says so.

`Model.fuzzyScore` and `Model.filterRows` hold the matching, and `SwitcherSearch` is the line, kept as their own pieces so a second popup can take them.

(An earlier head of this PR gave the mailbox switcher the same line; #148 removed that popup, so this is the account switcher alone now.)

## Verification

`QT_QPA_PLATFORMTHEME= make validate` passes on this head (rebased onto `68e4243`). Tests: `tests/test_model.js` (the scoring — letters in order, the start of a word beats the middle, an address is searched and a name outranks an address that only contains it — and the filter), `tests/qml/tst_account_switcher.qml` (Down and Ctrl+K move; a name and an address both find the mailbox; the combined row stands aside while typing; a digit and a space are typing, and clearing rests the cursor on the mailbox in use).

## UI evidence

Captured on the same synthetic mailboxes (two IMAP accounts, Ada and Bob), the app's own window at 1280×800, the same Omarchy theme and scale; "before" is the tree of #139's head, "after" this head. Click an image for full size.

### `Alt+A`, nothing typed

| Before | After |
| --- | --- |
| before-switcher-empty | after-switcher-empty |

After: a search line at the top of the popup, the rows beneath it as before (every mailbox, then each account), the cursor on the account in use, the popup centred. Before: the rows alone, anchored at the address in the status bar.

### "bob" typed

| Before | After |
| --- | --- |
| before-switcher-match | after-switcher-match |

After: the rows narrow to the mailbox the letters name, by its name or its address; the combined "All mailboxes" row stands aside while anything is typed. Before: there is nowhere to type.

### "zzz" typed

| Before | After |
| --- | --- |
| before-switcher-none | after-switcher-none |

After: no mailbox row at all, and the Add and Manage rows still there. Before: as above.


gw7523:
Since opening: a digit is a key only where a row on show carries its number (a folder called by numbers no row carries is typed like any name; in the account switcher a digit is always a letter); the filtered row's origin is `sourceIndex` rather than `index`, which a Repeater owns; typing the query away rests the cursor on the row in use rather than the combined row; a key that reaches the popup's column is forwarded to the search line. Tests cover each. `QT_QPA_PLATFORMTHEME= make validate` passes on this head.

gw7523:

gw7523:
Rebased onto `68e4243` with #139's new head underneath. #148 removed the mailbox switcher, so this PR is now the account switcher's search line alone (title and body rewritten; `tst_mailbox_switcher.qml` is gone with the popup). `QT_QPA_PLATFORMTHEME= make validate` passes on this head.

huacnlee:
Visual evidence required for this UI change. Please add clearly labeled before-and-after screenshots to the PR description, using synthetic or redacted data and GitHub-hosted attachments. Compare the same view, interaction state, window size, theme and scale, and explain the visible differences. Cover the account switcher with empty, matching and no-result queries. Update the images after any further UI changes.

gw7523:
The before-and-after screenshots are in the description now — the switcher with nothing typed, with letters that name a mailbox, and with letters that name none — one pair per state on the same two synthetic mailboxes at 1280×800, with the visible difference under each. Please take another look.

gw7523:
Rebased onto v0.8.2 (`27c1b4a`, after #151); the head is now `7c53697`. The commit's content is unchanged — `git range-diff` shows context lines only — so the evidence in the description still stands. `QT_QPA_PLATFORMTHEME= make validate` passes on the new head.

PR #130, opened 2026-09-07
Sign in to a work or school Microsoft mailbox, and send through Graph where the tenant has SMTP switched off

gw7523:
## What changes

**Microsoft 365 mailboxes can sign in.** The Outlook provider addressed every sign-in to the `consumers` tenant, so a work or school mailbox could not sign in at all. The tenant is now a reading off the account entry (`imap.tenant`), kept by `Microsoft.normalizeTenant` to `consumers`, `organizations`, `common`, or a tenant id or domain; anything else — a path, a host, a query — is the consumer tenant. It is one path segment of Microsoft's own URL, never a URL of its own, so a hand-edited `accounts.json` cannot steer a device code or a refresh token to another host. A work or school mailbox submits through `smtp.office365.com`; both kinds read through `outlook.office365.com`, fixed as before.

**Sending where the tenant has SMTP switched off.** Most Microsoft 365 tenants disable authenticated SMTP, and a send fails with "SmtpClientAuthentication is disabled". With `imap.send = "graph"` the same message goes to Graph's `sendMail` — one fixed address, POSTed as base64 MIME with a bearer token of Graph's audience — and Graph files the sent copy in Sent Items itself, so the client appends nothing. The Graph token is obtained with the refresh token already in the keyring, exchanged for `https://graph.microsoft.com/Mail.Send` the way the mail token is refreshed (`OutlookAuth.withGraphToken`); nothing new is signed in to, a rotated refresh token is stored back, and a registration without `Mail.Send` gets a plain-words error rather than a 401.

**Setup.** Two switches on the Outlook page: *Work or school account (Microsoft 365)*, and, under it, *Send through Microsoft Graph*. The README says when each applies.

## How

- `MicrosoftOAuth.js`: `normalizeTenant`, `authorityFor`, `deviceUrlFor`, `tokenUrlFor` (the consumer tenant answers with the existing constants, which is what keeps `tests/test_outlook_http.py`'s redirection working), `GRAPH_SCOPES`, `graphRefreshBody`, `missingGraphScope`.
- `Outlook.js`: `settings(address, tenant, send)`; `ImapProtocol.normalizeSettings` carries `send`; `Accounts.makeImapSettings` carries `tenant` and `send`, neither a secret.
- `OutlookAuth.qml`: takes the entry's settings for those two readings only; a second, purpose-tagged keyring lookup feeds the Graph exchange through the same `postForm`, `isCurrent` and session-generation gates, so logout and identity change discard a late Graph answer as they do a mail one.
[...]

PR #141, opened 2026-09-08
Name a mailbox where it is set up

gw7523:
A mailbox could be named only afterwards, from the settings list (#102), though the moment anyone wants a name is the moment they add a second mailbox that differs from the first only in its domain. The IMAP, JMAP and Gmail setup pages now open with a name field above the rest of the form (on the Gmail page above the steps, since the shared OAuth client's step folds away once it exists, which is exactly when a second mailbox is being added) — optional, and empty is no name, which puts the address back — through one `AccountNameField`. IMAP and JMAP save it with the rest of the form and on the way to a sign-in; the Gmail page writes it to the entry when the client is saved and when the sign-in starts, whichever is pressed. One fix underneath: a draft mailbox is addressed by its position until it has an id, and `configureAccount` used to let go of that position on any change, so a name typed before the address turned the setup page into the previous mailbox's — it now keeps the position until the row has an id. The field shows the name the entry already has, which a new `Service.accountName` says — `accountLabel` cannot, because it falls through to the address's local part. The HEY page has no form to put it in; the settings list still names a HEY mailbox.

## Verification

`QT_QPA_PLATFORMTHEME= make validate` passes on this head. New test: `tests/qml/tst_setup_name_field.qml` (IMAP: the name comes from the store, goes back trimmed with the form and with a sign-in, and empty clears it; Gmail: a second mailbox with the client step folded still has the field, and the name goes with a sign-in and with saving the client; the real Service: naming a draft keeps it current). Not on the HEY page, which has no form; Outlook's page is in #130 and will take the field there.

## UI evidence

Captured on the same synthetic mailboxes (an IMAP account and a Gmail account already set up), the app's own window at 1280×800, the same Omarchy theme and scale; "before" is this PR's base (`85eb6c2`), "after" this head. Click an image for full size.

### Add a mailbox → IMAP

| Before | After |
| --- | --- |
| before-setup-imap | after-setup-imap |

After: a **Name (optional)** field above the address, with the hint of what it is for. Before: address and password alone.

### Add a mailbox → JMAP

| Before | After |
| --- | --- |
| before-setup-jmap | after-setup-jmap |

After: the same field, in the same place. Before: none.

### Add a mailbox → Gmail, with a Gmail account already set up

| Before | After |
| --- | --- |
| before-setup-gmail-second | after-setup-gmail-second |

After: the name field sits above the client step, so a second Gmail mailbox can be told apart from the first before it is signed in. Before: the client step alone, and two Gmail mailboxes list by the same address shape.


huacnlee:
Visual evidence required for this UI change. Please add clearly labeled before-and-after screenshots to the PR description, using synthetic or redacted data and GitHub-hosted attachments. Compare the same view, interaction state, window size, theme and scale, and explain the visible differences. Cover Gmail, IMAP and JMAP setup, including a second Gmail account with the shared-client step folded. Update the images after any further UI changes.

gw7523:
The before-and-after screenshots are in the description now — the IMAP and JMAP setup pages with the name field, and the Gmail page for a second Gmail account with the field above the client step — one pair per state on the same synthetic mailboxes at 1280×800, with the visible difference under each. Please take another look.

gw7523:
Rebased onto v0.8.2 (`27c1b4a`, after #151); the head is now `887db28`. The commit's content is unchanged — `git range-diff` shows context lines only — so the evidence in the description still stands. `QT_QPA_PLATFORMTHEME= make validate` passes on the new head.

PR #147, opened 2026-09-08
Load the next page as the list is scrolled to its foot, and start with fifty

gw7523:
The list stopped at a "Load more" button after twenty-five rows, so reading a mailbox was a scroll, a press, a scroll. The next page now comes as the list reaches its foot — never while a page is on its way — and a page that lands still short of the foot (a small mailbox, a merged list held back by a mailbox that has not caught up) is followed by the next once the list settles, until the foot is out of view or there is no more. The ask is made a moment after the signal rather than inside it, so a cached page painted on the way into a live load cannot race that load. The button stays for the page that did not come. The first page is fifty rows rather than twenty-five, in the manifest, the service and the account alike, so the cache key and the page agree.

The list takes the panel's scroller as `scroller` and watches its foot itself, so the panel says one word and the list does the rest.

## Verification

`QT_QPA_PLATFORMTHEME= make validate` passes on this head. New test: `tests/qml/tst_load_more.qml` (reaching the foot asks once, not again while the page is on its way, again when the page has landed and the foot is reached anew, a short page is followed until the foot leaves the view, never when there is no more; the button still asks).

Not changed here: in the merged All Mailboxes view, rows past the page watermark stay hidden until every mailbox with more has loaded past it, so the cursor stops at the last visible row after a load; scrolling now asks every such mailbox for its next page, which moves the watermark, but a mailbox that answers slowly still holds the rest back.

## UI evidence

Captured on the same synthetic mailbox (thirty messages with more on the server), the app's own window at 1280×800, the same Omarchy theme and scale; "before" is `main` (`68e4243`), "after" this head. Click an image for full size.

### The list, at the top

| Before | After |
| --- | --- |
| before-list-top | after-list-top |

Nothing changes here: the first page as it lands. (The first page is fifty rows on this head, twenty-five before; the mailbox here holds thirty.)

### The list scrolled to its foot, with more on the server

| Before | After |
| --- | --- |
| before-list-foot | after-list-foot |

After: the foot reaching view asks for the next page by itself — the button already reads **Loading**, with nothing pressed. Before: the foot waits with **Load more** until it is pressed.

### The page that did not come

| Before | After |
| --- | --- |
| before-list-failure | after-list-failure |

After: the failure is said in the status bar, and the button is back to **Load more** for a press — the page is asked for again by hand, not on a loop. Before: the same button, since it was never pressed.


gw7523:
Rebased onto `68e4243`: the list inset #148 changed is kept as upstream has it, with the scroller hooked up beside it. No other change. `QT_QPA_PLATFORMTHEME= make validate` passes on this head.

huacnlee:
Visual evidence required for this UI change. Please add clearly labeled before-and-after screenshots to the PR description, using synthetic or redacted data and GitHub-hosted attachments. Compare the same view, interaction state, window size, theme and scale, and explain the visible differences. Cover the list before and after automatic paging, including loading, failure and short-page states. Update the images after any further UI changes.

gw7523:
The before-and-after screenshots are in the description now — the list at its top, scrolled to its foot with more on the server (the next page asking for itself, against a button waiting to be pressed), and the page that did not come — one pair per state on the same synthetic mailbox at 1280×800, with the visible difference under each. Please take another look.

PR #149, opened 2026-09-08
Move a message to a label from the reader and the row menu, through a picker that draws the tree

gw7523:
Stacked on #140. `v` opened a picker over a flat list of labels, filtered by a typed substring, and nothing else did. The reader now carries a button beside Archive and the row menu a row, both opening the same picker on the message they are about — archived under the label on Gmail, moved into the folder on IMAP. The picker draws the labels as the rail does: the tree, a child stepped in under its parent by its own name; typed letters narrow it to the best matches, flat, by the whole path, matched loosely (`Model.fuzzyScore` from #140) so "wk" finds Work and everything under it.

`Model.movableLabels` returns the tree with a `depth` and `leaf` per row when nothing is typed, and scored matches otherwise; the labels given are not written on. One upstream expectation widened: "an" finds "banana" first and "Archive notes" after it, where before it found only the substring.

## Verification

`QT_QPA_PLATFORMTHEME= make validate` passes on this head. Tests: `tests/test_model.js` (tree order and depth, the label the list is in left out, typed letters), `tests/qml/tst_move_to_label.qml` (real App/Service with a controlled client: `v` opens the tree, typing narrows it and Enter moves the message under the label chosen; the row menu and the reader's button open the picker on their own message).

## UI evidence

Captured on the same synthetic IMAP mailbox (one account, five folders, four messages), the app's own window at 1280×800, the same Omarchy theme and scale; "before" is the tree of #140's head, "after" this head. Click an image for full size.

### The reader

| Before | After |
| --- | --- |
| before-reader | after-reader |

After: the toolbar under the message gains the Move button (the label glyph) between Archive and Trash. Before: Archive and Trash alone.

### The row menu

| Before | After |
| --- | --- |
| before-row-menu | after-row-menu |

After: a **Move to a label...** row under Archive. Before: no way to file from the row.

### `v` on a message

| Before | After |
| --- | --- |
| before-picker-tree | after-picker-tree |

After: the picker, headed **Move to** with the count, its search line, and the labels as the rail draws them — a child stepped in under its parent (Flights under Travel, Invoices under Work). Before: `v` does nothing; the list stays.

### The picker with "inv" typed

| Before | After |
| --- | --- |
| before-picker-query | after-picker-query |

After: the list narrows to what the letters name, best first, each by its whole path (Work/Invoices). Before: nothing, as above.


gw7523:
Rebased onto `68e4243` with #140's new head underneath; no change of its own. `QT_QPA_PLATFORMTHEME= make validate` passes on this head.

huacnlee:
Visual evidence required for this UI change. Please add clearly labeled before-and-after screenshots to the PR description, using synthetic or redacted data and GitHub-hosted attachments. Compare the same view, interaction state, window size, theme and scale, and explain the visible differences. Cover the reader move button, message-row menu, hierarchical destinations and filtered results. Update the images after any further UI changes.

gw7523:
The before-and-after screenshots are in the description now — the reader's Move button, the row menu, the picker drawing the tree, and the picker narrowed by typed letters — one pair per state on the same synthetic mailbox at 1280×800, with the visible difference under each. Please take another look.

gw7523:
Rebased onto v0.8.2 (`27c1b4a`, after #151); the head is now `4710d9d`. The commit's content is unchanged — `git range-diff` shows context lines only — so the evidence in the description still stands. `QT_QPA_PLATFORMTHEME= make validate` passes on the new head.

PR #152, opened 2026-09-08
Let a press on a menu row end at the menu row

sunli829:
The account menu opens from the address in the status line, at the bottom
left, and goes up from there across the mailbox rail. Choosing a mailbox
from it also activated the rail row the menu was drawn over: the account
switched and the mailbox switched with it, in one click, and the second
half of that was never asked for. Pressing a rail row to dismiss the menu
did the same thing — the menu closed and the mailbox changed.

Both halves are the same mistake, and it is not modality. The rows in
these menus were drawn with a `TapHandler`, and so are the rows on the
rail. A handler takes a *passive* grab: it answers the press and lets it
carry on down the stack, so two of them stacked one over the other both
saw the same press and both fired their tap on the release. A `MouseArea`
takes the exclusive grab instead, and the press ends where it landed. The
rows on both sides of this are `MouseArea`s now — every clickable row in
the five popups, and the rail rows underneath them.

Modality is the other half of the story and is here too, because the two
routes into a window are not the same. A non-modal popup closes on an
outside press and lets that press through to whatever *item* is under it,
so dismissing a menu drawn over the message list also selected the row it
was covering. `modal: true` stops that; `dim: false` keeps it from
darkening the window behind a menu, which is not what modality is here
for. The two halves are not interchangeable and both are load-bearing:
with only the `MouseArea`s, pressing the rail to dismiss the menu still
switched the mailbox; with only modality, choosing from the menu still
did. An open popup does not stop a `TapHandler` beneath it from answering
a press — which is the whole reason the rail rows had to change as well.

Both rules are in AGENTS.md now, under *Popups and their triggers*, rather
than pasted into five files that each know half of them.

Two things follow from modality and are worth naming. A control can open
*and* close its own popup now, because the press that closes a modal one
never reaches the control — so the "closed a moment ago" timestamps that
`ComposeView` and `ContactsPicker` kept for that are gone, along with the
250ms window in which they silently dropped a genuine reopen. And while
any of these menus is up, a press outside it puts it away and does nothing
else: right-clicking a different message row no longer moves the row menu
there in one click, and middle-click-archive is swallowed. That is what
dismissing a menu should cost, but it is a change.

## Verification

`make validate` green on this commit — which runs the QML suite: `validate`
depends on `test`, and `test` on `test-qml`. 453 QML tests, 0 failed.

`tests/qml/tst_menu_over_rail.qml` is the reproduction, driven through the
real window: it fills the rail with enough rows to reach the part of the
window the menu is drawn over, opens the menu where the status line opens
it, and watches the rail row under the pointer. Both gestures — choosing
from the menu, and pressing the rail to dismiss it — report one activation
of the covered rail row against `main`'s components and none with the
change. A short rail is the arrangement in which this bug cannot be seen
at all, which is why the test builds a long one.

`tests/qml/tst_account_switcher.qml` gains the item-side pair: a press
outside the menu closes it without reaching the surface behind, and a row
still answers its own mouse. The first fails against the unfixed code with
"and does not reach what it was covering".

`tests/qml/tst_compose_recipients.qml` gains two. A contact's `+ Cc` pill
puts the address in Cc — it fails against the first draft of this change,
which had covered the pills; see below. And picking an address leaves the
keyboard in the field it went into, which is the one thing modality could
have cost the compose view.

`tests/qml/tst_popup_keys.qml`'s fixtures are modal now. AGENTS.md points
at that file to hold the rule that a popup answers its own keys, and it
was holding it for a shape the app no longer ships.

A fresh agent reviewed the diff against AGENTS.md. It found a blocker I
had shipped: the row-wide `MouseArea` in `ContactsPicker` was declared
after the `+ To` / `+ Cc` / `+ Bcc` pills and covered them, so every pill
added to To — the exclusive grab that fixes this bug is exactly what made
that unreachable, where the old handler had let the press through to the
button. It also found a fifth popup with the same defect, `ComposeView`'s
From menu, which the first draft missed; the dead reopen guards; a stale
comment in `AppMenu`; and that the new test file had been copied whole
from `tst_app_navigation.qml`, header comment and 180 lines of unused
fixture included. All are fixed here. Its security verdict was PASS.

Not done, and needed before this merges: driving it on a screen. These are
offscreen tests and they cannot see a menu that lands four pixels off or a
hover that stopped following the pointer. Somebody has to open the account
menu, the message menu, the app menu, the label picker, the contacts
picker and the From menu with the mouse, in a light theme and a dark one
and at both window sizes, and put a screenshot here.

## Release Notes

- Choosing an account from the menu no longer also switches the mailbox
  the menu was covering, and dismissing a menu no longer acts on whatever
  was underneath it.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -161,6 +161,24 @@ different features and only one of them is here.
   bug this rule exists to prevent.
 - A popup that would overflow flips to the other side of its trigger, then
   clamps to the window edge, then clamps to zero. All three, in that order.
+- A popup is `modal: true` with `dim: false`. Modality is what makes a press
+  outside it end at the overlay: a non-modal popup closes on that press *and*
+  lets it through, so putting a menu away also acted on whatever the menu was
+  covering. `dim` is off because the modality is there for the grab and not for
+  a scrim — the window behind a menu does not darken.
+- A clickable row inside a popup is a `MouseArea`, and so is anything the popup
+  may be drawn over. A `TapHandler` takes a *passive* grab: it answers the
+  press and lets it carry on down, so two of them stacked one over the other
+  both fire on the same release — choosing from a menu drawn over the mailbox
+  rail chose from the menu and switched the mailbox under it. Modality does not
+  help here, because an open popup does not stop a handler beneath it from
+  answering a press; only the exclusive grab a `MouseArea` takes does.
+  `tests/qml/tst_menu_over_rail.qml` holds both halves.
+- A control can both open and close its own popup, because the press that
+  closes a modal one never reaches the control. A non-modal popup needed a
+  "closed a moment ago" timestamp to tell the two apart; there is no such
+  timestamp any more, and reintroducing one would be the second mechanism for
+  something modality already does.
 
 ## Keys and focus
 
--- a/components/AccountSwitcher.qml
+++ b/components/AccountSwitcher.qml
@@ -131,7 +131,9 @@ Item {
     width: Style.space(250)
     implicitHeight: rows.implicitHeight + Style.space(8)
     padding: Style.space(4)
-    modal: false
+    // Modal and undimmed; see AGENTS.md, "Popups and their triggers".
+    modal: true
+    dim: false
     focus: true
     closePolicy: QQC.Popup.CloseOnEscape | QQC.Popup.CloseOnPressOutside
     onHeightChanged: root.place()
@@ -246,8 +248,9 @@ Item {
         }
 
         HoverHandler { id: unifiedHover }
-        TapHandler {
-          onTapped: {
+        MouseArea {
+          anchors.fill: parent
+          onClicked: {
             menu.close()
             root.unifiedChosen()
           }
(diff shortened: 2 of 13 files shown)


sunli829:
Closing: #148 already fixed the half of this that matters, and did it better — `TapHandler { gesturePolicy: TapHandler.ReleaseWithinBounds }` takes the exclusive grab on press without giving up the handler idiom. My version replaced those handlers with `MouseArea`s, and one of them covered the `+ To` / `+ Cc` / `+ Bcc` pills in the contacts picker; the upstream fix has no such problem, which I checked by running this branch's pill test against main, where it passes.

Two gaps are left over, in case they are worth their own change later. Both are reproduced by `tests/qml/tst_menu_over_rail.qml` and `tests/qml/tst_account_switcher.qml` on this branch:

1. A press outside a menu closes it *and* lands on the item underneath — the popups are `modal: false`, so dismissing a menu drawn over the message list also selects the row it was covering. `modal: true; dim: false` fixes it, measured.
2. A press outside a menu that lands on a **rail row** switches the mailbox as well. Modality does not stop that one: an open popup does not keep a `TapHandler` beneath it from answering a press, and giving the rail rows `ReleaseWithinBounds` does not help either, because the press is outside the menu rather than under it. I did not find a fix for this one.

PR #156, opened 2026-09-08
Suggest labels for a message in the move-to picker, from counts of what each label holds

gw7523:
Stacked on #149 (move to a label), which sits on #140. The picker that `v`, the reader's button and the row menu open drew the tree and asked to be typed into; on an account with thirty labels that is three letters and Return every time, for a message whose sender alone says where it goes. It now leads with up to three labels the message looks like, headed **Suggested**, each with why underneath — the sender, its domain, a word — and Return takes the first; the tree follows under **All labels**, less the ones suggested, and typing narrows both: a suggestion stays on top while the letters still name it and drops off when they do not.

The judgement is counts, not a model. Per label: how often each sender address, sender domain, recipient, subject word and body word has been seen under it. A message's own tokens are looked up in every label's counts and the labels ranked by how much likelier each makes the message than the mailbox as a whole does — a multinomial naive Bayes over the five kinds of token, read as lift. Each kind counts once whatever its length (the mean lift of its tokens, each capped so one rare token cannot own the kind), weighted so who sent it counts for more than a word in it and a long body cannot outvote the sender; a word no label has ever seen says nothing about any of them, and one seen elsewhere but not here counts against. A label none of the message's tokens has ever been seen under is not offered, however small it is. The count maps carry no prototype, so a subject that says "constructor" is a word and not a lookup. `account/LabelBrain.js` holds the rules (tokeniser, features, scoring, learning, the file's shape); `account/LabelBrain.qml` holds the files: one profile per account under `$XDG_DATA_HOME/omamail/labels/`, named as the cache names its own so the name cannot leave the directory, and a journal of choices beside it. `docs/LABELS.md` says the rest.

The counts come from two places, and both add to the same file:

- **The build.** Settings → Mailboxes → **Build now**: a job of the agent runner's — `agent-job.py` with `labels: true`, its own kind, a transient systemd user unit like an ask, cancellable, listed in the pane — whose command is `scripts/label-brain.py build` and whose spec is the job's message file. The script finds the himalaya account whose `email` is the mailbox's address in himalaya's config (two of them is a question, not a guess), lists the newest 2000 envelopes of each label — every page 500, so a cap that is not a multiple still reads the newest `cap` and no page starts inside another; a later page failing keeps what was read and says the label is short — reads the newest 150 of those for their text (text parts first, HTML stripped when that is all there is, the first 4000 characters, the id after a `--`), and writes the file once, whole, 0600 in a 0700 directory, and only there: the name must be absolute, of the shape the window gives a profile, in that directory itself, with no link on the way, or the spec is refused before anything is read. A label that cannot be listed keeps the counts the previous file held for it, the choices made since riding along in that entry and read back in the shape a fresh one has, so one bad folder on a rebuild does not forget a label; every label failing is a failure and writes nothing. One progress line per label, which the settings page shows while it runs; the last line is the summary. No agent, no model.
- **Choices.** A label chosen in the picker for one message adds that message's tokens to the label — and the reader's text with them when the message is the one open — twice over when the label was not among the suggestions, since the brain was wrong or knew nothing and that is the stronger lesson. What there is to learn is read before the move, while the row is still listed (its tokens, not the row), and learnt after it, once the move has been sent: a refused move teaches nothing, and a ticked batch teaches nothing at all — it was offered no suggestions, and one label for many messages is a filing, not a lesson. The window never rewrites the profile: each choice is one line appended to a journal beside it (and each message leaving its label is a line too), by one command that makes the directory 0700 and appends the line, 0600 if the file is new. The commands run one at a time in order, each carrying the paths it is for, so an account switch cannot misfile one; Forget removes both files after the append in hand and drops the ones waiting; a read waits for the queue and the queue for the read, so neither sees half of the other; a command that does not finish in ten seconds is told to stop. Reading is the profile plus the journal's lines newer than the build; the build folds the journal in — it read the archive those choices moved messages into — and removes it. Nothing is written while a build runs. A profile never built still learns from choices and the settings page says that is all it knows.

Nothing is unlearnt: a message moved out of its label, or thrown away from inside it — and on a provider that files by folder, archived or marked as spam — is counted, one per listed row of a batch, and at twenty-five since the build the settings page says a rebuild is due. Nothing is filed: suggestions are rows, taken by Return or a click. A merged list and a ticked batch get no suggestions — a suggestion is about one message on one account. Off by default, since on, every move writes a file; the setting's text says what is read and where it is kept, and **Forget** deletes it.

The tokeniser is the one rule the window and the script must share, and the test runs the Python one over the same strings and compares the stop lists.

## Verification

`QT_QPA_PLATFORMTHEME= make validate` passes on this head (on `68e4243`). Tests: `tests/test_label_brain.js` (tokens and their Python twin, features, scoring — sender names the label, words name it when the sender is new, the sender outvotes three hundred words of another label's, one rare word cannot outvote four receipt words, unseen words change nothing, "constructor" is a word, a lesson learns from its stored tokens, no evidence is no suggestion, the label on screen and system labels are never offered, the limit — learning and its weights, the file's shape and what it refuses, maps and labels bounded as they are read, pruning at 6000 to the 5000 most frequent, the rebuild rule at twenty-five, the build's spec), `tests/test_label_brain.sh` (the build against a himalaya that is a shell script: full pages and a cap of 600 over 1200, a later page failing, the body cap, HTML as words, a label that cannot be listed keeping its counts from the file before — made whole, a corrupt one not stopping the build — while a listed one is replaced whole, the journal of choices folded in and removed, a file outside the labels directory, a relative name, a name of another shape, a symlink, or a labels directory that is a link refused, HEY refused, the account found through himalaya's config, two accounts with one address refused, an unknown address failing plainly, every folder failing writing nothing, 0600/0700), `tests/test_agent_job.sh` (the `labels` kind), `tests/test_agent.js` (the kind, the newest build per account, no attention, the finished note, the payload), `tests/test_model.js` (the picker's rows: suggestions first with their headings and reasons, one heading per group, the tree less them, typing keeps or drops a suggestion, three at most), `tests/qml/tst_label_suggestions.qml` (the picker draws the headings and reasons, Return takes the first and says so, typing drops a suggestion the letters do not name, the plain tree when there are none, the cursor following a list that shortens under it; the file-holding object learns at once and forgets; a choice is one line appended by one command with the paths it is for, a departure a line of its own, one command at a time; Forget removes both files after the append in hand and drops the ones waiting; a command that does not finish is told to stop and one that never started is let go; a read waits for the queue and the queue for the read, and what is read is the profile plus the journal's lines newer than the build, once; an account switch leaves the old account's appends bound for its own journal), `tests/qml/tst_move_to_label.qml` (with the setting on, a choice teaches the brain and the next picker leads with it; a ticked batch opens plain; trash from inside a label counts as leaving it, a batch counts each listed row and not a tick on one that is gone, archive and spam count on a folder provider, a move to the label on screen or marking read does not; the lesson is the row's tokens read before the move and learnt after it, a batch has none, and off there is none).

The build was also run by hand against a live iCloud IMAP mailbox through himalaya (three folders with spaces and an apostrophe in their names, caps of 20 envelopes and 2 bodies): ten seconds, the file as described, and unseen messages from two of the three folders scored to their own folder — the third, a forwarded reservation, to another, which a full build should settle.

## UI evidence

Captured on the same synthetic IMAP mailbox (one account, five folders, four messages), the app's own window at 1280×800, the same Omarchy theme and scale; "before" is the tree of #149's head, "after" this head. Click an image for full size.

### The list, nothing open

| Before | After |
| --- | --- |
| before-list | after-list |

Nothing changes: the list and the rail are untouched.

### `v` on a message: the picker, plain

| Before | After |
| --- | --- |
| before-picker | after-picker |

Nothing yet: with the setting off, or nothing learnt, the picker is the tree as #149 drew it.

### The picker after two choices (Receipts for an invoice, Travel/Flights for a flight), opened again on the invoice

| Before | After |
| --- | --- |
| before-picker-suggested | after-picker-suggested |

Before: the plain tree, since there was nothing to suggest. After: a **Suggested** heading with Receipts on top and its reason underneath (“invoice” · “march”), then **All labels** with the tree less Receipts; the cursor rests on the suggestion, so Return takes it.

### Settings → Mailboxes

| Before | After |
| --- | --- |
| before-settings-mailboxes | after-settings-mailboxes |

After: the **Suggest labels from the archive** switch (On here) with its explanation, the status line ("Learning from your choices alone so far: 2 messages…"), **Build now** and **Forget**, and the note on what the build reads. Before: the name field goes straight to the mailbox list.


gw7523:
The body sample is now the newest 150 messages per label (was 50); the script, the spec default, the Settings text and the docs say so. `QT_QPA_PLATFORMTHEME= make validate` is being re-run on this head.

gw7523:
The before-and-after screenshots are in the description now (synthetic mailbox, the app's own window at 1280×800, same theme and scale, one pair per state with the visible difference under it). Please take another look.

gw7523:
Rebased onto v0.8.2 (`27c1b4a`, after #151); the head is now `87b753e`. The commit's content is unchanged — `git range-diff` shows context lines only — so the evidence in the description still stands. `QT_QPA_PLATFORMTHEME= make validate` passes on the new head.

huacnlee:
Thank you for the thoughtful implementation and the unusually thorough explanation and testing.

I’m going to leave this feature out of Omamail. Although it does not use an AI model, it introduces a substantial local classification subsystem: mailbox scanning, token statistics, profile and journal persistence, background jobs, a Python builder, lifecycle coordination, and additional settings UI. That is too much long-term complexity for the simple and maintainable direction I want for Omamail.

My longer-term direction is to expose Omamail capabilities through a CLI or MCP interface. Classification and filing suggestions can then be handled by an external AI agent without embedding another learning system and its maintenance surface into the mail client itself.

This is a product-scope decision rather than a criticism of the care put into the code. I’m sorry we cannot introduce it, and I appreciate the work behind it.

gw7523:
Thank you and understood.
-
Jack W. Holloway

On Sep 11, 2026, at 02:27, Jason Lee ***@***.***> wrote:

﻿

Closed #156<https://github.com/huacnlee/omamail/pull/156>.

PR #160, opened 2026-09-08
Prevent active attachment content from opening directly

ecohash-co:
## Summary

Prevent sender-controlled active content from reaching `xdg-open`, even when its filename pretends to be a passive document.

Opening now checks both the normalized filename and the attachment's decoded bytes before creating a temporary file. HTML, SVG/XML documents, desktop entries, scripts, Windows PE programs, ELF binaries, Mach-O binaries and Windows shortcuts are refused. Saving remains available and preserves the sender's sanitized filename.

Refused attachments create no runtime file and never start `xdg-open`.

## Verification

- `python3 tests/test_attachment_common.py`
- `bash tests/test_attachment_open.sh`
- `make test-js`
- `python3 tests/test_qml_names.py`
- `python3 tests/test_curl_config.py`
- `bash tests/test_source.sh`
- `git diff --check origin/main...HEAD`

The regression suite covers HTML disguised as PDF, SVG disguised as PNG, UTF-16 HTML, shebang scripts, PE and ELF executables, dangerous filename normalization, valid PDF/PNG/plain text, no launcher invocation and no file write on refusal.

Local checks pass without Docker on `79cb66a`. GitHub Actions provides the full Linux portable run for the current head.

## Release Notes

- Prevent active attachment content from opening directly, even when a sender disguises it with a safe-looking filename.


huacnlee (approved):
Reviewed and approved; Security: PASS for the affected active-content launcher boundary.

PR #162, opened 2026-09-08
Fix helper script paths for scoped Omarchy 4.x plugins

ueltyx:
## Summary

Fix Omamail mailbox setup on Omarchy 4.x when installed as a scoped third-party plugin.

Omarchy intentionally removes the private `manifest.__sourceDir` field. Without a fallback, Omamail attempted to start `/scripts/config-store.sh` and `/scripts/mail-transport.sh`, leaving IMAP setup stuck at `Checking`.

The contributor reproduced this with Proton Mail Bridge and confirmed that resolving the helper directory from `Service.qml` itself restores mailbox setup.

Current `main` now contains the final implementation: it resolves `Qt.resolvedUrl(".")` independently of manifest internals, decodes the filesystem URL and removes the trailing slash. After resolving the merge conflict, this PR retains regression coverage for a scoped manifest without `__sourceDir`, including the decoded component directory used by helper commands.

## Verification

- `bash tests/test_source.sh`
- `bash tests/test_service_source.sh`
- `python3 tests/test_qml_names.py`
- `make test-js`
- GitHub Actions on the current head

Local portable checks above pass without Docker on `83d80f7`. The local QML runner is unavailable because its Qt 6.7 binary links to a removed Homebrew ICU 74 library; GitHub Actions runs the QML regression on Linux.

## Release Notes

- Fix mailbox setup for scoped Omarchy 4.x plugin installations by resolving bundled helper scripts independently of private manifest fields.


huacnlee (approved):
Reviewed and approved; Security: PASS for the affected local helper-path boundary.

PR #164, opened 2026-09-09
Find the plugin's own directory when the manifest does not carry it

flyingnobita:
Omamail's helper scripts never ran on an ordinary install. `Service.qml` derived `pluginDir` from `manifest.__sourceDir` alone, and Omarchy's shell deletes that key from the manifest it hands a third-party plugin: `publicPluginManifest()` in `shell.qml` strips `__sourceDir`, `__isFirstParty` and `__hostCapabilities` unless the plugin is first-party. So `pluginDir` was `""`, every command became `/scripts/…`, and none of them started.

Nothing said so, which is why it reads as a dead button rather than a bug. `beginLogin()` sets `loginBusy` before launching `scripts/pkce.sh`, and a Process that fails to start raises nothing the interface can read, so Gmail sign-in showed "Finish the sign-in in your browser…" over a browser that never opened. `config-store.sh`, `keyring-store.sh` and `register-mailto.sh` were dead the same way.

To reproduce: install the plugin the way a user does, with `omarchy plugin add`, enable it, and press **Sign in with Google**. A checkout under `~/.config/omarchy/plugins` reproduces it too, symlinked or not, because `PluginRegistry` scans that directory as third-party regardless of how it got there.

The answer the engine always has is the component's own URL. `Qt.resolvedUrl(".")` from `Service.qml` is the plugin root by construction, since `manifest.json` names `Service.qml` at that path and `scripts/` is its sibling. Measured on Qt 6.11.2, Qt keeps the symlink a file was loaded through rather than resolving it away, so a `make install` checkout and a cloned install name the directory the same way and the mailto handler written from it stays valid under both.

That trades one silent failure for a possible other: a nested service entry point would make this a *wrong* directory rather than no directory, and a wrong one passes every existing `pluginDir === ""` guard. `tests/test_service_source.sh` now holds `entryPoints.service` at the root so that cannot happen quietly.

The decision lives in `PluginPath.js` rather than in QML, per the working agreement that everything which parses or decides is reachable by the node tests without a compositor. An empty source directory is treated as no answer rather than as a decision, so the fix does not depend on whether the shell sanitises by deleting the key or by blanking it.

**This touches `AGENTS.md`.** The Layout table said the root holds the three QML entry points "and nothing else", which `PluginPath.js` contradicts, so the row now names it and records the root invariant above. Amending the working agreement is your call, not mine: if you would rather the library lived under a module directory, say where and I will move it and revert the doc.

## Verification

`make validate` green on this commit: 520 + 4 + 64 passed, 0 failed, exit 0.

Red/green, node: regressing `resolve()` to the original manifest-only behaviour fails `tests/test_plugin_path.js` with `actual: ''`. Red/green, QML: against `origin/main:Service.qml`, `test_a_stripped_manifest_still_finds_the_checkout` fails. Both new guards in `tests/test_service_source.sh` were checked by defeating them: deleting the manifest branch while keeping its comments, and nesting `entryPoints.service`. `tests/test_plugin_path.js` also covers percent-decoding, invalid sequences, and the URL forms that are refused: `file://host/share`, `qrc:`, `/`, empty and null.

`tests/qml/tst_move_wiring.qml` set `__sourceDir: ""` to keep every helper unlaunchable while it drives the real service. Since empty is no longer a decision, its fixture now names a directory that does not exist, which holds the same guarantee without shaping production logic around a test.

Driven by hand on Omarchy 4.0.0.r2090.g5b91db5, Qt 6.11.2: a Gmail account signed in end to end through the loopback flow, which was impossible before this change, and an Outlook account added over device-code IMAP. After restarting the shell on this commit both accounts refetched within ten seconds, no `/scripts/…` failures appear in the log, and `~/.local/share/applications/omamail.desktop` was written for the first time. No interface change, so no screenshots.

Worth knowing when reproducing: `omarchy-launch-shell` sets `QS_DISABLE_FILE_WATCHER=1`, so QML edits are not read live as CONTRIBUTING says; `omarchy restart shell` is needed.

Two fresh agents reviewed this. The first rejected an earlier revision for putting parsing logic in QML against the working agreement, for justifying a fallback order with a claim about `Qt.resolvedUrl` canonicalising symlinks that measurement contradicts, and for guards a comment could satisfy; it was rewritten around those findings rather than patched. The second reviewed this diff and found that the empty-source-directory rule was shaped around a test fixture, that one guard was still comment-satisfiable, and that nothing held `Service.qml` at the root. All three are fixed above.

## Release Notes

- Helper scripts now run when Omamail is installed as an ordinary plugin. Gmail sign-in, saving accounts and calendars, the keyring, attachments, contact suggestions and the `mailto:` handler were all silently doing nothing.
- On first run after upgrading, Omamail registers itself as the desktop's `mailto:` handler and begins reading contacts for recipient suggestions, both of which it had been failing to do.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -23,7 +23,7 @@ three directories away from the client that calls it.
 
 | Module | What it is |
 |-----------------|--------------------------------------------------------|
-| root | `Service.qml`, `BarWidget.qml`, `App.qml`, and nothing else. `manifest.json` names these three and the shell loads them at that path. |
+| root | `Service.qml`, `BarWidget.qml`, `App.qml`, and `PluginPath.js`. `manifest.json` names the three QML files and the shell loads them at that path. The one library here answers where the plugin is installed, which is a fact about this directory rather than about any module under it, and it reads that from `Service.qml`'s own URL: the service entry point stays at the root, or the answer becomes a wrong directory instead of none. `tests/test_service_source.sh` holds that. |
 | `providers/` | Everything that differs between mail services: a description per provider, the registry over them, the protocol each speaks, and the pair of objects — signs in, fetches — that each needs. |
 | `account/` | One mailbox and the list of them. `MailAccount.qml`, `Accounts.js`, and the rules in `Model.js` about what a list does after an action. |
 | `cache/` | What a query result and a message body are kept in, and the two objects that keep them. |
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,7 @@ test-js:
 	node tests/test_oauth.js
 	node tests/test_microsoft_oauth.js
 	node tests/test_credentials.js
+	node tests/test_plugin_path.js
 	node tests/test_secrets.js
 	node tests/test_gmail_api.js
 	node tests/test_message.js
--- a/PluginPath.js
+++ b/PluginPath.js
@@ -1,0 +1,66 @@
+.pragma library
+
+// Where the plugin's own directory comes from, so that pluginDir + "/scripts/…"
+// names a file that is really there.
+//
+// Omarchy's shell hands a third-party plugin a sanitised manifest:
+// publicPluginManifest() deletes the source directory unless the plugin is
+// first-party. A pluginDir read from that field alone was empty on every
+// ordinary install, so each helper resolved to "/scripts/…" and failed to start
+// without raising anything the interface could show.
+//
+// The answer the engine always has is the component's own URL. Measured on Qt
+// 6.11.2, it keeps the symlink a file was loaded through rather than resolving
+// it away, so a checkout linked in by make install and a cloned install name the
+// directory the same way, and the mailto handler written from it stays valid.
+//
+// This depends on Service.qml sitting at the plugin root, which is what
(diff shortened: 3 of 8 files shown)


huacnlee:
Closing: main already carries this fix. #174 moved the UI under `ui/` and made `pluginDir` resolve from `Qt.resolvedUrl("..")` in `ui/Service.qml`, so a manifest stripped of `__sourceDir` no longer leaves the helpers at `/scripts/…`. In a real Omarchy install the directory it produces is the same one `PluginPath.resolve` would, and the extra inputs this PR guards against (`file://host/…`, malformed `%` sequences, `?`/`#` in a directory URL) are ones Qt never emits for a local directory, so there is no behavioural difference left to merge.

Your diagnosis of `publicPluginManifest()` stripping the key was exactly right, and the write-up is what let us confirm the shell behaviour quickly. If you think the QML regression test (`tst_plugin_dir.qml`, pinning `pluginDir` to the checkout root) and the entry-point guard are worth having on their own, a small PR against current main with just those would be welcome.

PR #165, opened 2026-09-09
test: Hold the plugin's own directory to tests, whatever the host says

### LINKED ISSUE #163: Save client / all script-backed writes silently fail: pluginDir is empty (manifest.__sourceDir stripped for third-party plugins)
## Summary

`pluginDir` (and therefore every feature that shells out to a bundled script — saving OAuth credentials, keyring storage, PKCE, contact suggestions, mailto registration, the account/window/calendar config writers) resolves to an empty string when Omamail is installed the normal way, via `omarchy plugin add`. In the setup panel this shows up as "Save client" appearing to do nothing: no error, no saved file, and sign-in never proceeds.

## Root cause

`Service.qml` derives its own directory purely from the manifest the shell hands it:

```qml
readonly property string pluginDir: manifest && manifest.__sourceDir
  ? String(manifest.__sourceDir) : ""
```

But Omarchy's shell (`shell.qml`, `publicPluginManifest()`) strips `__sourceDir` (along with `__isFirstParty` and `__hostCapabilities`) from the manifest of any plugin that isn't first-party:

```qml
function publicPluginManifest(manifest) {
  if (!manifest) return null
  if (manifest.__isFirstParty) return manifest
  var copy = JSON.parse(JSON.stringify(manifest))
  delete copy.__sourceDir
  delete copy.__isFirstParty
  delete copy.__hostCapabilities
  return copy
}
```

Every plugin installed with `omarchy plugin add` is third-party by definition (`firstParty: false` in `omarchy menu keybindings --print` / `omarchy-shell shell listPlugins` output), so this stripping always applies to Omamail as installed. `PluginRegistry.trustedCapabilities()` also only grants capabilities to first-party plugins, so there's no opt-in path today for a third-party plugin to get its own source directory back.

Confirmed live via `journalctl --user -t omarchy-shell`:

```
WARN: Process failed to start, likely because the binary could not be found. Command: QList("/scripts/config-store.sh", "credentials.json")
```

The leading `/` shows `pluginDir` was `""` at call time — every `pluginDir + "/scripts/..."` call collapses to a path off the filesystem root.

## Environment

- Omarchy version: 4.0.3-1
- Omamail commit: 7f7c1f071001cfaec7342c25b35ede1fc823b330 (latest `main` as of this report)
- Installed via: `omarchy plugin add https://github.com/huacnlee/omamail.git --enable`

## Suggested fix

Have `pluginDir` fall back to resolving its own file location when the manifest doesn't carry `__sourceDir`, since that doesn't depend on any shell-granted capability:

```qml
readonly property string pluginDir: manifest && manifest.__sourceDir
  ? String(manifest.__sourceDir)
  : String(Qt.resolvedUrl(".")).replace(/^file:\/\//, "").replace(/\/$/, "")
```

This idiom (`Qt.resolvedUrl(".")` / `Qt.resolvedUrl("relative/path")`) is already used elsewhere in Omarchy's own first-party plugins for the same purpose (e.g. `shell/plugins/agents/Panel.qml`, `shell/plugins/bar/widgets/Indicators.qml`), so it should behave the same way inside Omamail's own QML files. I patched my local install with exactly this change and confirmed it fixes the issue: toggling the Omamail window now successfully writes `~/.config/omamail/window.json` (previously nothing was written and the log showed the same "Process failed to start" warning for every `config-store.sh` invocation), and the setup panel error path also resumed working end to end for credential saving.

Happy to open a PR with this change if useful.

gw7523:
Closes #163.

This started as a fix. While it sat, #115 landed the same reading of
`Qt.resolvedUrl(".")` on `main`, so the fix is yours already and this branch is
now just the regressions for it, rebased onto it.

## Why it is worth covering

`pluginDir` is load-bearing in a way that fails silently. Every script this
plugin runs is named against it — the mail transport, the config store, the
keyring, `pkce.sh`, contact suggestions, the mailto registration, the calendar
writers. When it reads empty, the commands go out as `"/scripts/..."`, every
one of them fails to start, and nothing says why: the mailbox reports "Not
connected", "Save client" appears to do nothing, and an Outlook sign-in cannot
keep its refresh token because the keyring command never runs. #163 describes
the same thing arriving as a Gmail sign-in that never proceeds.

It is also a boundary the host controls and can change again. That is the case
for pinning the behaviour rather than the implementation.

## The tests

`tests/qml/tst_plugin_dir.qml`, six cases:

- a service with no manifest at all works out the checkout's own directory;
- so does one given the public manifest a stripping host passes;
- so does one whose manifest still carries `__sourceDir` — the directory is
  deliberately independent of the host, so all three agree;
- it is absolute, carries no trailing slash and no `file://`;
- a script command built from it is rooted under the plugin rather than at `/`,
  which is the shape the bug actually produced.

Five of the six fail against the previous `manifest.__sourceDir` reading,
including the `"/scripts/mail-transport.sh"` symptom itself. The third is the
one that does not: it passed before because a manifest carrying the field was
the only case that used to work.

## Verification

`QT_QPA_PLATFORMTHEME= make validate` passes on this head. Qt 6.11.2,
Omarchy 4.0.3.

Driven by hand as well, on a third-party install: before the fix the shell
logged a `Process failed to start` for every `"/scripts/..."` command and the
window read "Not connected"; after it, those stop and the mailbox refreshes.
@tlehman confirmed the same independently on #163, including the Gmail path.

## A separate snag, not fixed here

`tests/qml/tst_agent_ownership.qml:94` asserts
`pluginDir.indexOf("/omamail") >= 0`, so `make validate` fails whenever the
checkout directory is not itself named `omamail` — a git worktree, a fork
cloned under another name, or a CI path. It passes from a directory of that
name and fails from any other, which is how I met it. Left alone here since it
is not this branch's business, but it is worth loosening to something that does
not depend on what the directory is called.

### DIFF
--- a/tests/qml/tst_plugin_dir.qml
+++ b/tests/qml/tst_plugin_dir.qml
@@ -1,0 +1,76 @@
+import QtQuick 2.15
+import QtTest 1.3
+import "../.." as Omamail
+
+// Where the plugin's scripts are found. A host that hands the service a public
+// copy of the manifest strips its private fields, and the service must still
+// know its own directory: without it every command is built as "/scripts/..."
+// and nothing that shells out can start.
+Item {
+  Omamail.Service {
+    id: withoutManifest
+    manifest: null
+  }
+
+  // What a shell that strips private fields actually passes: the public
+  // manifest, with `id` and `name` but no `__sourceDir`.
+  Omamail.Service {
+    id: publicManifest
+    manifest: ({ id: "omamail", name: "Omamail", version: "0.8.2" })
+  }
+
+  // A manifest that still carries the private field, which the directory is
+  // deliberately independent of.
+  Omamail.Service {
+    id: hostSaidWhere
+    manifest: ({ id: "omamail", __sourceDir: "/opt/somewhere-else" })
+  }
+
+  TestCase {
+    name: "PluginDir"
+
+    // The repository root, which is where Service.qml lives and so what the
+    // service should work out for itself.
+    function root() {
+      var url = String(Qt.resolvedUrl("../.."))
+      var path = decodeURIComponent(url.substring("file://".length))
+      return path.length > 1 && path.charAt(path.length - 1) === "/"
+        ? path.substring(0, path.length - 1) : path
+    }
+
+    function test_a_service_with_no_manifest_still_knows_its_directory() {
+      compare(withoutManifest.pluginDir, root())
+    }
+
+    function test_a_manifest_without_the_private_field_falls_back_to_the_file() {
+      compare(publicManifest.pluginDir, root(),
+        "a host that strips __sourceDir must not leave the scripts unrooted")
+    }
+
+    function test_the_directory_is_absolute_and_has_no_trailing_slash() {
+      var dir = publicManifest.pluginDir
+      compare(dir.charAt(0), "/")
+      verify(dir.charAt(dir.length - 1) !== "/", dir)
+      verify(dir.indexOf("file://") < 0, dir)
+    }
+
+    // The command every script is named against is rooted, rather than the
+    // "/scripts/mail-transport.sh" an empty directory produced.
+    function test_a_script_command_is_rooted_under_the_plugin() {
+      var command = publicManifest.pluginDir + "/scripts/mail-transport.sh"
+      compare(command, root() + "/scripts/mail-transport.sh")
+      verify(command.indexOf("//scripts") < 0, command)
+      verify(command !== "/scripts/mail-transport.sh")
+    }
+
+    // The directory is worked out from the file and nothing else. A manifest
+    // that still carries `__sourceDir` does not move it, so the plugin behaves
+    // the same whether the host keeps the private field, strips it, or passes
+    // no manifest at all — which is three different hosts today.
+    function test_the_directory_does_not_depend_on_the_manifest() {
+      compare(hostSaidWhere.pluginDir, root())
+      compare(hostSaidWhere.pluginDir, publicManifest.pluginDir)
+      compare(hostSaidWhere.pluginDir, withoutManifest.pluginDir)
+    }
+  }
+}


tlehman:
Tested this on Omarchy 4.0.3 with Omamail 0.8.2 (#163). I had independently patched `pluginDir` to the same `Qt.resolvedUrl(".")` fallback before finding this PR, and it fixes Gmail sign-in — `scripts/pkce.sh` runs, the browser opens, and `config-store.sh` writes again.

The `decodeURIComponent` handling here is a genuine improvement over what I had; my version would have broken on a path with a space. Keeping the manifest as the winner when it carries the field seems right too.

+1 from me.

gw7523:
Rebased and repurposed. #115 landed the same `Qt.resolvedUrl(".")` reading on `main` while this sat, so the fix is yours already — I have dropped mine and this branch is now only the regressions for it.

@tlehman thank you for the independent confirmation, and for the Gmail/`pkce.sh` half; I had only traced the Outlook and transport paths.

One test changed meaning in the rebase. Mine asserted that a manifest carrying `__sourceDir` should win. `main` deliberately does not do that — "independently of host internals" — so it now asserts the opposite: the directory is the same whether the host keeps the field, strips it, or passes no manifest, which is three different hosts today. Five of the six fail against the old `manifest.__sourceDir` reading.

Separately, and not touched here: `tests/qml/tst_agent_ownership.qml:94` asserts `pluginDir.indexOf("/omamail") >= 0`, so `make validate` fails whenever the checkout directory is not named `omamail` — a git worktree, or a fork cloned under another name. It passes from a directory of that name and fails from any other, which is how I met it. Happy to send a separate patch loosening it if you would like one.

`make validate` passes on this head: 588 QML, 4 native Outlook HTTP, 64 sidebar text.

huacnlee:
Closing: main already holds this behaviour to tests. `ui/tests/qml/tst_agent_ownership.qml` (`test_public_manifest_resolves_local_helper_directory`) and `ui/tests/qml/tst_service_settings.qml` (`test_scoped_manifest_resolves_bundled_helpers_without_private_metadata`) both pin `pluginDir` for a manifest stripped of `__sourceDir` to the checkout root, absolute, decoded and without a `file:` prefix. Since #174 the QML suite also lives under `ui/tests/qml/` and the Makefile only runs that directory, so a file added at `tests/qml/tst_plugin_dir.qml` would merge cleanly but never execute.

The snag you noted about `indexOf("/omamail")` tying the test to the directory's name is gone too: that assertion now compares against `Qt.resolvedUrl("../../..")` exactly, so worktrees and differently named clones pass.

The two cases main does not cover — `manifest: null`, and a manifest that still carries `__sourceDir` being ignored — are marginal, but if you would like them in, a small PR adding them to one of the two existing tests under `ui/tests/qml/` would be welcome.

PR #167, opened 2026-09-10
Count unread from listed ids and show the number on the bar icon

nixfred:
The bar badge and the Unread tab treated Gmail's `resultSizeEstimate` as a total. An unread poll of three ids comes back with estimate 201 and a continuation token even when four messages match, which is how a mailbox with three unread conversations in Gmail reported 201 in Omamail.

The unread poll now counts listed ids. A finished page is exact. A truncated page uses the listing estimate only when the provider says that number is a real total (IMAP SEARCH, JMAP query, HEY). Gmail's estimate is never exact, so Gmail lists up to 100 ids and keeps paging when more remain, hydrating only the three preview rows. The envelope stays in the bar; the unread total sits on its corner, capped at 99+.

Gmail's web UI counts conversations. Omamail lists messages, so a thread with two unread replies is two in the badge and one in Gmail. That is the remaining difference after 201, and it matches the Unread mailbox.

## Verification

`make validate` green on 38f61594d92f5c2e383307d05704d09ad1d31c2c.

`tests/test_model.js` covers a truncated Gmail page whose estimate is 201 (counts 3, keeps listing), a truncated Gmail page whose estimate is 12 (counts 2, keeps listing), a finished page of four ids, an IMAP SEARCH total of 10, and an IMAP SEARCH total of 201 (finishes at 201, does not keep listing). Replacing `listedUnread` with `return page.estimate` fails: `201 !== 3`.

`tests/test_provider.js` asserts Gmail's unread count page is 100 and its estimate is not exact, and that IMAP, Outlook, HEY, and JMAP mark theirs exact.

`tests/qml/tst_bar_widget_icon.qml` asserts the envelope's overlay is empty at 0, `4` at 4, and `99+` at 201.

`tests/test_source.sh` refuses `inboxUnread = page.estimate`, requires `Model.listedUnread`, `Provider.unreadCountPage`, `Provider.unreadEstimateExact`, a `slice(0, 3)` preview, `getMessages(previewIds`, and `root.countHandle`.

Live Gmail API on this mailbox, unread query `in:inbox is:unread -category:promotions -category:social -category:forums`: maxResults=3 returns estimate 201, 3 ids, continuation token; maxResults=100 returns estimate 4, 4 ids, no continuation. INBOX `threadsUnread` was 3. I did not drive the desktop to screenshot the overlay after the first pass, because that stole focus.

A fresh agent reviewed the first draft and found the 201 special-case sitting in Model.js. That is gone: whether an estimate is exact is a provider flag, default false.

## Release Notes

- The unread badge no longer reports Gmail's dummy 201 when only a handful of messages are unread.
- The bar envelope keeps its shape and shows the unread total on the corner.


huacnlee (changes requested):
Two changes here, and they get different answers.

**Counting unread from the listed ids: yes.** Gmail's `resultSizeEstimate` on a three-id page is a placeholder, and `inboxUnread = page.estimate` is still what main does, so the tooltip can read "201 unread messages" over three and the `estimate > before` arrival check compares noise against noise. Counting the ids a finished page lists, and trusting the estimate only where the provider says it is exact, is the right shape, and keeping the provider flag out of `Model.js` was the right call.

**The number on the envelope: no.** The dot is deliberate, not a placeholder for a count that had not been made exact yet. A count on a mail icon is never zero on a busy mailbox, climbs while you are doing something else, and asks to be cleared; the dot says whether and leaves when to you, so the mailbox can be worked asynchronously rather than as a queue. #187 writes that down in `docs/DESIGN.md` so it stops looking like a gap. Please drop the `GmailIcon.qml` and `BarWidget.qml` changes and `badgeText` with them: the bar says whether, the tooltip says roughly how many, the window says what.

The branch also predates #174, which moved the UI under `ui/` and the provider logic into Rust (`src/providers/*.rs`, with `ui/providers/Registry.js` as the thin QML side), so it conflicts as it stands. The unread poll is now in `ui/account/MailAccount.qml` (`api.listMessages(Provider.unreadQuery(providerId), 3, …)`), and the provider capability flags would go alongside the other per-provider facts there. A rebase that keeps the counting change and leaves the icon alone would be welcome; keeping `UNREAD_LIST_CAP` so a Gmail poll is at most a handful of requests is worth carrying over too.

huacnlee:
Superseded by #188, which carries the counting half of this change onto current main. The poll has lived in the Rust checker (`src/sync/mod.rs`) since #174, so the fix is there now: list a hundred at a time, count the ids handed back, follow the continuation to a cap of 500, hydrate three for the preview. You are credited as co-author on that commit — the 201 and the estimate-versus-count distinction are yours.

The envelope keeps its dot, for the reasons in #187.

PR #169, opened 2026-09-10
Attach the file when another app opens a new message

Ruegen:
Another app can ask Omamail to start a new email and include a file (for example a scanned PDF). Omamail opened the new message, but the file was missing. That is especially easy to miss because mail stays running from the bar, so the window is often already open.

This change puts that file on the draft. A link inside an email that pretends to attach a file is ignored, so a message cannot quietly attach something from your disk. Only a real file path from the desktop mail helper is accepted.

## Verification

`make validate` was green on this commit. The new tests fail on current main (the file never reaches the draft) and pass with this change. On a running desktop, opening a new message with a file attached should show that file on the draft even if Omamail was already open.

A first version of this patch would have trusted attach names inside email links, which is unsafe, and it would have listed some files twice. This commit only takes a checked list of local files from the desktop helper.

No before-and-after screenshots yet. What you should see: the attachment row on the new-message screen gains the file. The rest of the form is unchanged.

## Release Notes

- A file handed over when opening a new message now appears on the draft, including when Omamail is already open from the bar.

### DIFF
--- a/components/ComposeView.qml
+++ b/components/ComposeView.qml
@@ -572,6 +572,11 @@ DropArea {
       fromWasChosen = true
     }
     if (mode === "draft") loadDraftAttachments(messageId, attachments)
+    var files = Array.isArray(values.attachments) ? values.attachments : []
+    for (var a = 0; a < files.length; a++) {
+      var path = String(files[a] || "")
+      if (path !== "") enqueueAttach("read", path)
+    }
   }
 
   // Where the keyboard goes when composing becomes the context. A reply starts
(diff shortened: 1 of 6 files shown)


Ruegen:
Follow-up: pressing Back on a new message that still had the file attached crashed Omamail. The window was holding the whole file in memory when it tried to leave. It now keeps only the file path, stops any attach still in progress, and Back can close the draft.

Please try Email PDF again, leave the file on the message, and press Back.

PR #173, opened 2026-09-11
Read a page through a window and wait out a throttled read

vyacheslav-voloshyn:
## What went wrong

Two Gmail accounts, both over the Gmail API, `maxMessages` at its default of 50. Opening the window — or switching accounts, or any poll that finds a page the cache does not hold — put `Gmail is rate limiting this account. Try again shortly` in the bar, with no `(retry in Ns)` because Google sent no `Retry-After`. The count and the list then stayed as they were until the next poll two minutes later.

Reproduction, one account, cold cache:

1. `maxMessages` at 50.
2. Delete `~/.cache/omamail` and open the window on a mailbox with 50+ messages.
3. `Alt+A` to the second account and back, twice.

`getMessages()` fetched the page with every id in the air at once — its own comment said *"Fetches every id at once"*. Gmail allows one user 250 quota units a second and charges five for a `messages.get`, so a 50-row page is exactly the ceiling, and the `messages.list` that found the ids shares that second. Google answers the overflow with `403 rateLimitExceeded` (the legacy envelope, `error.errors[0].reason`), and `request()` retried nothing but `401` — so the refusal went straight to the mailbox as a failed page.

## What changed

- **A page is read through a window.** `getMessages()` keeps `Api.MAX_PARALLEL_READS` (10) reads in the air and starts the next as each lands: 50 units a round rather than 250 at once. The reads still overlap, which is why they were sent together in the first place; they are just no longer all of them. One flat loop, not recursion — a read can answer synchronously (no credential yet, or an id no URL can be built from) and its callback re-enters the pump, which on a large page would otherwise be one stack frame per id.
- **A throttled read is waited out instead of handed to the mailbox.** `GmailApi.retryDelayMs()` uses a `Retry-After` Google names, capped at 15s, and otherwise doubles from 1.2s with jitter, three attempts. `GmailApi.rateLimited()` decides what may be sent again: `429` always, `403` only for `rateLimitExceeded`/`userRateLimitExceeded` or `status: RESOURCE_EXHAUSTED` — because `403` is also how Gmail reports a missing scope, a disabled API and a suspended account, none of which waiting fixes.
- **Reads only.** A `401` can be re-sent because it says Google never processed the request; a throttling refusal can come from a front end with the write already committed behind it, and a re-sent `messages/send` is a second message nothing can take back. The retry is gated on `GET`.
- **The in-flight count moved from the call to the handle** (`handle.counted`, `release()`). A request that is sent again is one request as far as `busy` is concerned, and the pause holds the count so no caller is told the page failed mid-wait. `abortRequest()` releases it: a handle abandoned during a wait has no reply left to arrive — `abort()` already drove the previous one to DONE — and every mailbox switch and poll abandons the page before it (`MailAccount.qml:704, 723, 1510, 1892`). `release()` counts once per handle, so the ordinary reply releasing it again is not a second release.
- A waiting retry is part of the request and dies with it. If its Timer cannot be created the request fails instead of being re-sent with no wait at all — that would be the very burst this exists to prevent.
- `responseError()` now routes every reason the client treats as throttling to the one sentence a reader can act on; `userRateLimitExceeded` used to fall through to Google's own prose about quota metrics and timestamps.

Deliberately not done: a client-level FIFO like `JmapClient`'s `Jmap.makeQueue`. The window caps the one fan-out that earns the refusal and the retry covers the residual case where a list or a label read shares the same second. A queue over every Gmail request is the fuller answer and would want `makeQueue` to stop belonging to a JMAP file — glad to do that instead if you would rather have it.

## Tests

`tests/test_gmail_http.py` — loopback fixture, production `request()`, `getMessages()`, retry Timer and XMLHttpRequest; wired into `test-qml` beside the Outlook fixture. Six cases: a 30-id page never exceeds the window while every read is answered at once; a read refused once with `Retry-After: 1` succeeds and the mailbox is told nothing during the wait; a read that stays refused fails after four attempts and says why; a `403 insufficientPermissions` fails on the first answer in under a second; a page abandoned mid-wait gives its in-flight slot back; a throttled `POST` is refused once and left refused.

Against a pristine `634b4a3`, three of them fail:

```
FAIL!  : test_1_a_page_is_read_through_a_window() 'at most undefined reads in flight, saw 30'
FAIL!  : test_2_a_throttled_read_is_waited_out() the mailbox is not told anything while the wait runs
FAIL!  : test_5_an_abandoned_wait_gives_its_slot_back() 'the read is waiting rather than finished'
```

`tests/test_gmail_api.js` covers the two decisions directly: both `403` envelopes and the reasons that are not throttling, `Retry-After` winning and being capped, an HTTP-date and a negative falling through to the backoff, the attempts running out, and the reader-facing sentence for each throttling reason.

## Verification

- `make test-js` — green. `make test-shell` — green. `make qml-check` — no warning on either changed file. `omarchy plugin validate .` — clean.
- `make test-qml` — 645 passed, 0 failed, plus this fixture's 8 and the other two, **with #172 applied**. Without it, `tst_load_more::compile()` fails at `634b4a3` on its own and stops the target before the Python fixtures; that is why it is a separate one-line pull request rather than part of this one.
- By hand, two real Gmail accounts, `maxMessages` back at 50, cache deleted: mailbox opened and both accounts switched through repeatedly. No rate-limit message since.
- This diff was reviewed by a second agent with no part in writing it; the gating of the retry to reads and the release on abort are both its findings. One it raised is left standing and is pre-existing: `MailAccount.fetchSummaries()` discards the handle `getMessages()` returns, so those reads cannot be aborted, and the retry now extends how long such an orphan page can live. It is worth a separate change either way.

No interface change, so no screenshots.

## Release Notes

- Gmail mailboxes no longer report `Gmail is rate limiting this account` when a page of mail loads: a page is now fetched a few reads at a time, and a read Google throttles is waited out and retried rather than failing the page. Sending and other writes are never retried.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -180,6 +180,7 @@ test-qml:
 	QT_QPA_PLATFORM=offscreen QT_QUICK_BACKEND=software \
 		$(QMLTESTRUNNER) -import $(CURDIR)/tests/qml/imports -input tests/qml
 	python3 tests/test_outlook_http.py "$(QMLTESTRUNNER)"
+	python3 tests/test_gmail_http.py "$(QMLTESTRUNNER)"
 	python3 tests/test_sidebar_text.py "$(QMLTESTRUNNER)"
 
 # Both engines on the same fixtures. The QML column is the one that decides
--- a/providers/GmailApi.js
+++ b/providers/GmailApi.js
@@ -90,12 +90,11 @@ function responseError(status, payload, fallback) {
     reason = String(payload.error.errors[0].reason || "")
 
   if (status === 401) return "Google rejected the session. Sign in again"
-  if (status === 403 && reason === "rateLimitExceeded") return "Gmail is rate limiting this account. Try again shortly"
+  if (rateLimited(status, payload)) return "Gmail is rate limiting this account. Try again shortly"
   if (status === 403 && /Gmail API has not been used/i.test(detail))
     return "The Gmail API is not enabled for this Google Cloud project"
   if (status === 403) return detail ? redact(detail) : "Google refused this request"
   if (status === 404) return "That message is no longer in the mailbox"
-  if (status === 429) return "Gmail is rate limiting this account. Try again shortly"
   if (status === 0) return "Could not reach Gmail. Check the network connection"
   if (status >= 500) return "Gmail is having trouble right now. Try again shortly"
   if (detail) return redact(detail)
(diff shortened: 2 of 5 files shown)


huacnlee (changes requested):
The problem is real and still present on main, but the fix no longer lands where the code is. #174 moved Gmail HTTP out of QML into the Rust backend, so this branch cannot be rebased mechanically: four of the five files it touches are gone (`providers/GmailApi.js`, `providers/GmailApiClient.qml`, `tests/test_gmail_api.js`, `tests/test_gmail_http.py`), and the `XMLHttpRequest` layer that `request()`, `scheduleRetry()` and the loopback fixture are built on does not exist any more. `ui/providers/GmailApiClient.qml` now calls `backendRequest("gmail.read", …)` and the HTTP happens in `src/providers/gmail_http.rs`.

What main does today, so you can see what survives of this:

- `getMessages()` in `ui/providers/GmailApiClient.qml` still fires every id at once, as one `gmail.read` each. The only cap is the protocol's global `MAX_IN_FLIGHT = 32` in `src/backend/protocol.rs`, which is not per account and, with round trips well under a second, still lets a 50-row page burn far more than 250 units in one second.
- `execute()` in `src/providers/gmail_http.rs` sends once and never retries. It maps `429` to `gmail_rate_limited` and `403` to `gmail_forbidden` without reading the body or the `Retry-After` header, so a `403 rateLimitExceeded` is now indistinguishable from a missing scope or a suspended account. That is actually a step back from the pre-#174 QML, which at least parsed `error.errors[0].reason`.
- `backendError()` in the QML client turns both codes into a bar message and the page fails, exactly as your "What went wrong" describes.

So please rework rather than rebase. A map of where each piece goes now:

1. **Window**: the `pump()` loop ports almost verbatim into `getMessages()` in `ui/providers/GmailApiClient.qml`, and `MAX_PARALLEL_READS` into `ui/providers/GmailApi.js`. Keep the flat-loop rationale; a backend reply can still arrive on the same turn.
2. **Retry**: `rateLimited()` and `retryDelayMs()` belong in `src/providers/gmail_http.rs`, since that is the only place the `403` body and `Retry-After` are still visible. Read the body on a non-2xx before mapping the code, gate the retry on `Method::GET` for the reason you already wrote, and `tokio::time::sleep` between attempts inside the existing `deadline`. `gmail_rate_limited` should then only reach QML when the attempts run out, and it would be worth carrying the reason through so `backendError()` can keep the one sentence a reader can act on.
3. **Tests**: the decision tests move to `src/providers/gmail_http_runtime_tests.rs` (it already has a `(429, …, "gmail_rate_limited")` table) or `tests/gmail.rs`; the window test to `ui/tests/test_gmail_api.js` or a `ui/tests/qml/` case against a fake backend. `tests/test_gmail_backend.js` shows how the backend is driven over stdio if you want an end-to-end fixture.
4. Drop the Makefile change; `test-qml` now runs `ui/tests/qml` via `tests/run_qml_native.py`.

The design itself reads well and I would take it as described: window of 10, three attempts, `Retry-After` capped at 15s, reads only, the pause holding `busy`. The one thing I would add in the Rust version is that a per-account cap is cleaner there than in QML, since two accounts' pages and their `messages.list` calls all pass through the same client.

PR #179, opened 2026-09-11
Save an explicit authority for personal Microsoft accounts

MitchellMonaghan:
Personal Outlook/Hotmail setup currently saves `imap.tenant` as an empty string when the work/school switch is off. Device sign-in normalizes that value to `consumers`, so Microsoft consent can succeed, but the native account-bound credential resolver reads the saved empty string literally and rejects the token destination. Subsequent mail requests fail with `invalid_params`, shown as “Mail request failed”. Signing in again saves the same invalid value.

Persist `consumers` explicitly for personal accounts, keeping `organizations` for work/school accounts. Both are already supported by the pinned backend; no backend API change or release is needed. This changes the saved value only, not the setup page's appearance or the OAuth permissions and credential destinations.

To reproduce on main, configure a personal Outlook/Hotmail account with the work/school switch off, complete sign-in, then request its folders. The setup values contain an empty tenant and the account-bound request fails. The new regression checks the real setup component's personal → work → personal settings.

Existing entries with an empty tenant are not automatically migrated by this patch. Saving the affected account through the corrected setup page, or signing in again, writes the valid value.

## Verification

`make validate` passes locally with Rust/Cargo 1.98.1 and Qt 6.11.2 on Omarchy 4.0.3: Rust unit/integration tests, JavaScript and shell checks, 807 main QML cases, 6 native Outlook HTTP cases, 64 sidebar cases, the native HTML resource-request security test, qmllint (with warnings), plugin validation, and `git diff --check`.

The new `OutlookBoundaryReview::test_personal_account_saves_explicit_authority` test was observed failing against the original code (empty string instead of `consumers`) and passing with the fix. The targeted suite passes all 12 cases.

Live: after correcting the previously saved personal-account values locally and installing the setup fix, read-only folder requests succeeded for two Hotmail accounts, and the user confirmed both accounts work in Omamail. This live recovery included correcting existing saved settings; it is not evidence of an automatic migration.

A fresh, empty-context agent reviewed the diff and affected settings/credential path. No blocking findings; scoped security verdict PASS. It noted the existing-entry limitation documented above. Work/school behavior is covered by the regression test, not a live organizational account. No layout, label, theme or other visual changes are included.

## Release Notes

- Save valid sign-in settings for personal Outlook and Hotmail accounts so signing in again no longer leaves mail requests failing.

### DIFF
--- a/ui/components/OutlookSetupPage.qml
+++ b/ui/components/OutlookSetupPage.qml
@@ -55,7 +55,9 @@ Column {
     if (address === "") return null
     var clientId = validatedClientId()
     if (clientId === "") return null
-    var tenant = workSwitch.checked ? "organizations" : ""
+    // The native backend needs an explicit authority; saving an empty string
+    // lets device sign-in succeed but breaks subsequent account-bound refresh.
+    var tenant = workSwitch.checked ? "organizations" : "consumers"
     var send = workSwitch.checked && graphSwitch.checked ? "graph" : ""
     var imap = Outlook.settings(address, tenant, send)
     imap.tenant = tenant
--- a/ui/tests/qml/tst_outlook_boundaries.qml
+++ b/ui/tests/qml/tst_outlook_boundaries.qml
@@ -122,6 +122,18 @@ Item {
       compare(label.textFormat, Text.PlainText,
         "Server error messages must not be interpreted as resource-bearing HTML")
     }
+    function test_personal_account_saves_explicit_authority() {
+      var host = readyHost()
+      var service = createTemporaryObject(serviceFactory, parent, { auth: host.auth })
+      var page = createTemporaryObject(pageFactory, parent, { service: service })
+      verify(page !== null)
+      compare(page.accountValues().imap.tenant, "consumers")
+      var work = findChild(page, "outlook-work-switch")
+      work.checked = true
+      compare(page.accountValues().imap.tenant, "organizations")
+      work.checked = false
+      compare(page.accountValues().imap.tenant, "consumers")
+    }
     function test_cancel_matches_google_signin_and_allows_retry() {
       var host = readyHost()
       var service = createTemporaryObject(serviceFactory, parent, { auth: host.auth })


danvasi:
Independent confirmation of the diagnosis, from a separate clean install — Omarchy 4.0.3, plugin `main`, backend pinned `0.9.0`, Qt 6.11.2, one personal `@hotmail.com` mailbox. Nothing here contradicts the patch; two details below are things the pull request does not mention.

**Reproduction, from the user's side.** A personal Outlook account added with the work/school switch off signs in completely — device code, consent accepted, refresh token in the keyring — and then every request fails. Its saved entry held `"tenant": ""`.

**The failure is at the JSON-RPC handshake, not in IMAP.** Captured by temporarily logging the callback in `OutlookAuth.qml`:

```
auth.token({provider:"outlook", accountId:"outlook:…@hotmail.com", resource:"mail"})
  -> {"code":-32602,"message":"Invalid params"}
```

That is `destination()` in `src/auth/mod.rs` refusing the empty tenant, reached from `src/auth/credentials.rs:127` via `entry["imap"]["tenant"].as_str().unwrap_or("consumers")` — `unwrap_or` never fires for a present-but-empty string. QML turns the unrecognised code into `temporarily_unavailable` at `OutlookAuth.qml:540` (that is the text in the status row); the IMAP path reports the same failure as "Mail request failed".

**Why it looks like it half-works**, which cost me some time and may be worth a sentence in the release note: the in-memory access token from sign-in lives long enough to list the mailbox's folders once, and the plugin asks for a fresh token about every minute (`auth.token` reports `expires_in: 60`). So the mailbox appears to work, then dies roughly a minute after a successful sign-in, with nothing the user changed.

**Two things your patch does not cover:**

1. It is not only the window. The same empty tenant breaks the account-bound CLI path, because `resolve_account()` in `src/providers/imap/mod.rs` asks for a token before touching the socket:
   `omamail --json call imap.folders` with that `accountId` → the same `-32602`.
2. The IMAP settings path already tolerates the empty value — `outlook_settings()` in the same file treats a blank tenant as `consumers`. Only the token destination is strict. That asymmetry is what lets a broken account list folders once before it fails.

**The fix direction verified live**, for what it is worth as evidence: setting that single field to `"consumers"` in `accounts.json` and reloading made the account work immediately — `auth.token` returned a token carrying `openid offline_access https://outlook.office.com/IMAP.AccessAsUser.All https://outlook.office.com/SMTP.Send`, a live `imap.folders` call returned 19 folders, the inbox fetched (50 messages cached, estimate 1463), and the unread badge counted. As you documented, re-saving the account's settings page writes the empty value back and the failure returns within a minute.

**Optional, entirely the maintainer's call.** If legacy empty values should heal without the user re-saving the account, `credentials.rs` could read the field the way `outlook_settings()` already does — `.as_str().filter(|t| !t.is_empty()).unwrap_or("consumers")`. If the intended path for existing entries is the corrected setup page, then this comment is only evidence and no change is implied by it.

PR #181, opened 2026-09-12
Fix native calendar errors with provider recovery guidance

Dandiccf:
Native calendar requests currently collapse every backend failure to “The calendar request failed.” That leaves people without a useful next step, even though the configured calendar provider is already known.

This change keeps backend diagnostics private and uses only the existing provider kind to offer a safe recovery path: Microsoft users are directed to Graph permissions and reauthentication, Google users to reauthentication and Calendar access, and CalDAV users to their server address and password. Because it does not depend on backend error codes or messages, the pinned backend and API contract remain unchanged.

## Verification

Current head `6d1b7e8` incorporates upstream `3460ac8` (0.10.3/API4). `make validate` passes, including Rust, JavaScript, shell/security, plugin QML, standalone composition, both qmllint checks and plugin validation. The merge preserves upstream's native CalDAV credential-write/refresh test alongside our error regressions (targeted controller suite 13/13). The actual SHA256-verified x86_64 published backend passes its released API4 contract. This integration does not change our production error text or layout; retained screenshots describe the same PR behavior. Cross-platform release certification and hosted fork-CI authorization remain external.

The final JavaScript and QML regression tests were copied onto an unchanged `origin/main` snapshot. The JavaScript test failed because `nativeRequestError` did not exist, and the QML suite failed both new assertions because Microsoft failures still rendered as “The calendar request failed.”

A fresh independent agent reviewed the final diff and found no issues; security verdict PASS. The review confirmed that the mapping uses only `source.kind`, exposes no backend diagnostics, and requires no backend API revision or release.

The affected dark-theme, 1000 × 700 calendar view was rendered before and after with a synthetic Microsoft failure. No live mailbox was contacted, and the screenshots and diff contain no email addresses, messages, account identifiers, tokens, or other personal data.

Before, every provider failure showed only the generic calendar request message:

Before: generic native calendar error

After, the same Microsoft failure identifies the provider and points to Graph permissions and reauthentication:

After: Microsoft calendar recovery guidance

## Release Notes

- Calendar failures now identify Microsoft, Google, or CalDAV and suggest the relevant account or server settings to check.

### DIFF
--- a/ui/calendar/Calendar.js
+++ b/ui/calendar/Calendar.js
@@ -533,6 +533,19 @@ function graphResponseError(status, responseText) {
   return detail !== "" ? "Microsoft Graph answered: " + detail : "Microsoft Graph answered " + status
 }
 
+// Keep native backend diagnostics private while giving each provider an
+// actionable recovery path. This deliberately does not depend on backend
+// error-message semantics, so it remains compatible with the pinned API.
+function nativeRequestError(kind) {
+  if (kind === "microsoft")
+    return "Microsoft calendar request failed. Check Graph permissions in Settings, then sign in again"
+  if (kind === "google")
+    return "Google calendar request failed. Sign in again and check Calendar access"
+  if (kind === "caldav")
+    return "CalDAV calendar request failed. Check its server address and password in Settings"
+  return "The calendar request failed"
+}
+
 function googleEventsUrl(startMs, endMs) {
   return "https://www.googleapis.com/calendar/v3/calendars/primary/events?"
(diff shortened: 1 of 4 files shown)


Dandiccf:
Superseded by #183, merged into main as a5d514cdf25e63b0925102b246b730e6284a1a1a. Verified that main retains the provider-specific nativeRequestError mapping, its CalendarController call site and regression tests; #183 also adds the iCloud-specific guidance. Closing this duplicate PR because its changes have already landed through #183. No branch deletion.

PR #193, opened 2026-09-13
Refuse a call the pinned backend cannot take, and say the backend needs an update

huacnlee:
Stacked on #192. The plugin side of the released/unreleased step.

- `Runtime.js`/`Runtime.qml` carry `latestApiVersion` and `unreleasedMethods` from the runtime status beside the required revision; absent, malformed, or more than one step ahead reads as no step.
- `Backend.qml`: `apiVersion` is what the connected binary speaks (0.9.0 predates the field and is 1), `needsUpdate` whether the checkout is a step ahead of it, and `request()` answers a call to a method only that step has with `{code: -32012, message: "backend_needs_update"}` — the same shape as the backend's own refusals — instead of sending it to a binary that never heard of it. The helpers live in `Compatibility.js` and are unit-tested.
- `Service.backendNeedsUpdate` is the same flag for views: a feature on the step says "the backend needs an update" and waits; `tests/test_source.sh` (in #192) drops the check once the step is released.
- `BackendFixture.markReady` sets the new runtime fields for the QML suites.

Validation: `make test-js` (new cases in `test_backend_runtime.js` and `test_backend_compatibility.js`), QML suite 807/807 through `tests/run_qml_native.py`, `tests/test_backend_process.py`.

https://claude.ai/code/session_01HWHo1fASgShNhwyvn7vpXC


huacnlee:
Folded into #192 as a second commit; one PR for the whole model.

PR #208, opened 2026-09-14
Fix conversation rail flicker during reprojection

Depodra:
## Reproduction

All mailboxes view with a JMAP account whose Inbox has a thread of several messages:

1. Open that thread.
2. The conversation rail appears, vanishes and appears again — about three times — with the reader reflowing under it each time, before it settles.

A single message does not do it. On one mailbox it is milder, but the same mechanism.

## What is happening

The rail is drawn from `Service.conversationProjection`, asked of the backend whenever `conversationSource` changes, and asking began by blanking what was in hand — `showsRail: false`, no stops — until the reply. The source is an object literal, so it is a new object, and its change signal fires, whenever any input is reassigned, whether or not anything in it differs. In a merged view that is constant: `Unified.composeThread` and `composeMembers` build fresh objects on every read, and `mailboxes` is a new array on every unified snapshot, which every member account's list change produces. Opening a thread brings several of those in a row — the selection, the member seed and merge replies, the quiet mark-read's optimistic update — and each one was rail gone, `account.conversation`, rail back.

## The change

- A source that reads the same as the one last asked for, compared as text, asks nothing.
- When the source names the same rail as the projection in hand — the same thread, in the same mailbox, of the same account — that projection stays drawn until the new one lands. Only the ways along it are cleared — its navigation, and the first and last stop that `n` and `p` fall back to without one — so a stale next or previous cannot be followed in between, and no end jumped to; a late reply is refused by the serial as before.
- A different rail starts from a blank, as before, so its predecessor's stops are never drawn for it. The account is part of the identity because All mailboxes composes a message id with its account on the way out but leaves `thread.id` as the provider gave it, so two accounts can each hold a `t1` in their Inbox. Keyed on the thread and mailbox alone, the reader moving from one to the other kept the first account's rail, stops and caption up until the second's landed. `select` in a merged view happens to clear the other mailbox first, but two paths do not: switching mailboxes in the one-mailbox view while each has a thread open (`switchTo` leaves every mailbox's selection where it was), and turning All mailboxes back on while the mailbox that last answered the reader still holds a thread. The key's parts are JSON-encoded so no id can run into the next.
- A failed ask forgets that the source was asked, so the next change of the same source asks again.

The key and the keep-or-blank rule are `Model.projectionKey` and `Model.pendingProjection`, so node can prove them; `Service.qml` only wires them, and names the reading account in the source it keys (`accountId`, a field the backend's projection does not read).

## Before and after

Both frames: the reader open on a three-message conversation, at the moment a second projection of that same conversation is in flight — which is what a member merge, the quiet mark-read and a list refresh each produce while a thread opens. Same view, same state, same 1400×900 window, same theme (this machine's Omarchy theme, Ristretto, dark), scale 1, synthetic mail.

**Before** (main at `f2774a9`): the rail has collapsed and the body has reflowed to the full width. This is the frame that flashes.

Before: the rail collapsed while the projection is in flight

**After** (this branch): the rail, its stops and its caption stay up; only the ways along it — navigation, first, last — are cleared underneath, which draws nothing.

After: the rail stays up while the projection is in flight

Once the projection lands the two are byte-identical ([the settled frame](https://github.com/Depodra/omamail/releases/download/pr-208-screenshots/settled.png)): the change alters when the rail is drawn, not how it looks.

How they were made: rendered offscreen by the QML engine from the plugin's own `App.qml`, through the repository's QML test stubs of the shell singletons with this machine's theme colours and font size laid over them, with a real `Service` deciding the projection at each moment through the same `BackendFixture` the QML tests use. One harness file produced all four frames; only `ui/Service.qml` and `ui/account/Model.js` differ between before and after. A recording of the live shell cannot hold this frame still — it lasts one backend round trip — and the live mailbox is real mail, so the frames are synthetic and rendered. The images are release assets on my fork, the attachment host having no way in from the command line.

## Verification

`TZ=UTC LC_ALL=C make validate` green on `8d42c80` (Arch, Qt 6.11.2, cmake 4.4.3). `TZ=UTC` because `agent::events::tests::a_whole_day_runs_midnight_to_midnight_and_an_end_defaults_to_an_hour` assumes a UTC machine and fails on an Australia/Brisbane one with or without this change; `LC_ALL=C` because `tests/test_publish.sh` compares a `sort` of file names and en_AU collation orders `app/CMakeLists.txt` before `Cargo.lock`, again with or without this change.

New tests this round, each run against the previous head of this branch (`2b20961`, where the key was thread and mailbox alone) and watched fail:

- `ui/tests/test_model.js`: `projectionKey` keys the same thread id in two accounts differently, and no id can run into the part beside it.
- `ui/tests/qml/tst_unified_routing.qml`, `test_a_thread_with_the_same_id_in_another_account_is_another_rail`: two JMAP mailboxes in a merged view, each holding a thread `t1` in its Inbox; the reading host moves from one to the other, the projection in hand goes blank rather than keeping the first mailbox's stops, and what lands is the second's. Without the account in the key it fails with "Bob's t1 is not Ada's: nothing of Ada's rail stays up while Bob's is fetched — Actual true, Expected false".
- `test_model.js` and `tst_native_account_projections.qml` (`test_same_thread_keeps_the_rail_while_a_projection_is_in_flight`) now also assert that a kept projection's `first` and `last` are cleared with its navigation. Both fail on the first commit of this round (`afa06de`), where they survived.

From the first round, still in place and still failing on `main`: `ui/tests/qml/tst_native_account_projections.qml` — the same thread keeps the rail, stops and caption while a projection is in flight and loses only its navigation; an unchanged source does not ask again — and `test_model.js` for `pendingProjection`.

By hand (first round): the Omarchy shell on Hyprland, All mailboxes over a JMAP (Stalwart) mailbox and an IMAP one, opening multi-message threads from the JMAP account — the rail appears once and stays, where before it drew and vanished about three times per open. Driven at the three-column size in one theme. This round's change does not alter that path — one account's thread is keyed as before, with its account in front — and the cross-account case it adds is the QML test above and the frames.

Two fresh agents reviewed this round's diff with empty context, one against `AGENTS.md` and one against the findings above. What they found, all fixed in the second commit: the in-flight window still let `n` and `p` reach an end of the rail, because the kept projection carried the backend's `first` and `last`, which `App.stepMember` falls back to when the open message has no navigation entry (`main` never met this: its blank projection took `showsRail` with it); `projectedThread` and `sameThread` had outgrown what they held, now `projectedRail` and `sameRail`; the Node assertion for the encoded key did not discriminate a plain join, now it does; a blank projection shared its empty lists with every other blank, now built afresh; and the QML test now pins the mailbox it runs in and lets go of the service's host however it ends. Both traced `account.conversation` in `src/account/conversation.rs`: it indexes named fields of a `Value`, so the `accountId` the source now carries is ignored there, and no contract fixture pins the params. One asked whether the account should stay off the wire. It stays in the source on purpose: the text comparison must see an account switch too, or a source from another mailbox that happened to serialise identically would be neither re-asked nor blanked.

## Release Notes

- In All mailboxes, opening a conversation no longer makes the rail flash and the reader jump several times before it settles.


huacnlee (changes requested):
Requesting changes for two blocking findings:

1. `ui/account/Model.js:1371` does not account-qualify the projection identity. `Unified.composeThread` deliberately leaves `thread.id` raw, so two accounts in the unified view can both produce (for example) `t1\ninbox`. Switching between those accounts is then treated as the same conversation and retains the previous account's rail, stops, and caption until the new projection arrives. Please account-qualify this key and add Node and native-QML regressions covering a unified cross-account switch where the raw thread IDs are equal.

2. This changes visible rail flicker/reflow behavior, but the PR description has no comparable UI evidence. Please add clearly labeled before/after screenshots of the same view and interaction state, at the same window size, theme, and scale, using synthetic or redacted mail, plus a concise explanation of the visible difference.

The reviewed head is `2b20961c5e501f734bcc1810806dff55850533ee`. Security: PASS for the affected in-memory state scope. The PR is also currently conflicting with the latest base branch, so please resolve that conflict and retain the regressions above.

Depodra:
@huacnlee Thanks for the review. All three are addressed; the head is now `8d42c80`.

1. **Account-qualified key.** `Model.projectionKey` is now `JSON.stringify([accountId, thread.id, mailboxKey])`, and `Service.qml` names the reading account in the source it keys (`accountId`, which `account.conversation` in the backend ignores: it indexes named fields of a `Value`). Node: `ui/tests/test_model.js` — the same `t1` in two accounts keys differently, and the parts are encoded so no id can run into the next. Native QML: `ui/tests/qml/tst_unified_routing.qml`, `test_a_thread_with_the_same_id_in_another_account_is_another_rail` — two JMAP hosts in a merged view, each holding `t1` in Inbox; the reading host moves from one to the other and the projection in hand goes blank rather than keeping the first host's stops, and what lands is the second's. Both fail on `2b20961`. The switch is made by moving the reading host directly, because `select` clears the other host first and that clear happened to hide the fault; two real paths do not clear (switching mailboxes in the one-mailbox view with a thread open in each, and turning All mailboxes back on while the last reading host still holds a thread), and the key now covers them.
2. **Before/after screenshots** are in the description: same view, state, 1400×900 window, theme (Omarchy Ristretto, dark) and scale 1, synthetic mail, with the settled frame shown byte-identical between the two. The description says how they were made.
3. **Conflict** resolved by merging `main` (`f2774a9`) into the branch: `ui/account/Model.js` and `ui/tests/test_model.js` had #209's reload-depth block landing at the same spot as this branch's projection block; both are kept, upstream's first.

A fresh review of this round also found that a kept projection still carried `first` and `last`, which `App.stepMember` falls back to when the open message has no navigation entry, so `n`/`p` during the in-flight window jumped to an end of the rail; the second commit clears them with the navigation, with Node and QML assertions that fail on the first commit.

`TZ=UTC LC_ALL=C make validate` green on `8d42c80`.

Depodra:
Was waiting for this as I found another bug - however, that's addressed in https://github.com/huacnlee/omamail/pull/257 and is unrelated.

PR #226, opened 2026-09-15
chore: Fix calendar TLS fixtures across native test runners

Dandiccf:
The calendar discovery TLS peer fails before serving a request on the macOS jobs for #183: Python's `SSLContext.load_cert_chain` reports `[X509V3: INVALID_CERTIFICATE] invalid certificate`, then Rust reports JSON EOF because the peer exited before announcing its port. This is visible in the [standalone CI run](https://github.com/huacnlee/omamail/actions/runs/34983290042) and [native credential contract run](https://github.com/huacnlee/omamail/actions/runs/34983290043).

Use a checked-in synthetic CA and server certificate, following the existing mail-provider TLS fixtures, instead of generating a self-signed certificate with the runner's OpenSSL executable and configuration. The server certificate covers the existing test names, including the deliberately forbidden origin, so a credential-origin regression cannot hide behind a hostname mismatch. The fixture expires in September 2036 and is documented as public, test-only material.

Only test infrastructure changes. The peer still binds to loopback; the Rust test client still uses real TLS and pins DNS to that peer. Production TLS verification, credential destinations, redirect policy, API contract and release pin are unchanged. The new startup regression is part of `make test-shell-portable` and `make validate`.

## Verification

`make validate` passes locally on Linux, including Rust, JavaScript, shell/security tests, plugin and standalone QML, lint and plugin validation. Both `cargo test --locked --lib calendar::` and `cargo test --locked --lib --no-default-features --features standalone calendar::` pass all 20 tests, including the real HTTPS iCloud/Graph redirect and foreign-origin connection/credential regressions.

`python3 tests/test_calendar_tls_fixture.py` was observed failing both subcases against the unfixed peer: no OpenSSL executable on `PATH`, and a nonexistent `OPENSSL_CONF`. Both now pass, from a non-repository working directory. This reproduces the unwanted host-generator dependency, not the exact macOS certificate error. It also passes with the system Python.

Manually exercised the real peer with Python TLS: the trusted Graph and foreign-origin test names connect, while a wrong hostname and the untrusted chain are rejected. Strict OpenSSL chain, server-purpose and hostname checks pass. Local versions: Python 3.14.7 linked to OpenSSL 3.5.8, OpenSSL CLI 3.6.4. No real accounts or credentials were used; no UI changed. Native macOS confirmation remains pending the hosted checks; the precise platform-specific certificate-generation defect was not reproduced on Linux.

An independent fresh-context agent reviewed the complete seven-file diff and found no actionable functional, performance or security defects. It independently repeated fixture startup, strict chain checks, all four valid TLS hostnames, wrong-hostname rejection and rejection without the test root. Security: PASS for the scoped test-fixture boundaries on Linux; native macOS remains unverified. UI evidence and product decision: not applicable.


Dandiccf:
The calendar TLS fix now passes both previously failing macOS jobs. There is one remaining failure in the Windows backend job: five process tests time out or fail to observe child startup within the deadline.

The process code is unchanged, and the same eleven tests passed in the previous run with the same Windows runner image and Python version: https://github.com/huacnlee/omamail/actions/runs/34983290042/job/104441952659. This suggests an intermittent issue, but does not establish its cause.

Could you please rerun failed job 104473870899? https://github.com/huacnlee/omamail/actions/runs/34996474374/job/104473870899

I attempted to rerun it, but GitHub denied the request because I lack repository permissions. If it fails again, I can help investigate further with targeted process-start and pipe diagnostics rather than changing timeout behavior without evidence.

Dandiccf:
Thanks for fixing the calendar TLS fixture in #228. Its merged CSR/signing approach addresses the original macOS certificate failure, so I am closing this overlapping alternative rather than asking you to merge a second solution. No branch is being deleted.

For completeness, we also investigated the earlier Windows process-test failure in a local Windows Server 2025 VM. Using the unchanged process module from this PR's head (`ee100c2dc0c0e1f7c2fed23aeebf33df24502827`), Python 3.14.7 and Rust 1.98.1 targeting `x86_64-pc-windows-gnu`, all nine process-module tests passed in 20 consecutive runs (180 individual test passes). This was a scoped GNU harness, not the full MSVC standalone-backend CI job; it did not reproduce the failure or establish its cause.

The subsequent main CI run also passed its Windows backend job: https://github.com/huacnlee/omamail/actions/runs/35052219677. There is no need to rerun this superseded PR solely for the earlier request above. Thanks again!

PR #231, opened 2026-09-16
Preview a message on move instead of opening it, and make opening one fast

R037:
Moving the cursor with j/k now renders the row it lands on in the reader pane already on screen, without marking it read. Opening a message — through the cache, through Gmail, or on a repeat visit — is also substantially faster: a message that previously took most of a second to render now renders in single-digit milliseconds once it has been seen before, and far less even the first time.

### Why it was slow

Opening a message ran three costs that had nothing to do with how big the message actually was:

- `decodePart`'s UTF-8 branch and `decodeBase64Url` decoded base64 and rebuilt UTF-8 by hand, one bit at a time — up to 840ms for a 144KB message, almost all of it in that loop rather than in parsing the markup.
- `extractBody` and `extractHtml` each walked the message's MIME tree on their own, so a message with no text/plain part of its own — most of what falls back to html in the first place — had its one html part decoded twice.
- A confirmed repeat of an already-cached message still recomputed its attachments, its invite, and its unsubscribe offer from the live Gmail payload, still rewrote the (unchanged) body cache file to disk, and — because Gmail always describes a calendar part by reference rather than inline — still re-fetched an already-resolved invitation's ICS file on every single reopen, flickering its card in the process.

None of this scaled with message size in a way a bigger inbox or a faster machine would fix; it was pure repeated work.

### What changed

- `decodePart` and `decodeBase64Url` now decode UTF-8 through `Qt.atob` directly, after a one-time probe establishes what this build's `Qt.atob` actually returns (it is undocumented and, per its own deprecation notice, not the same as the browser API) — falling back to the byte-by-byte path only when the probe finds nothing usable, or when the result contains a replacement character, which is the one case Qt's own decode and this file's manual one disagree on. This also fixes a real bug: the old code assumed `Qt.atob` returns raw bytes and re-decoded that as UTF-8 a second time, mangling every non-ASCII character that reached `decodeBase64Url` (a calendar invite's summary, a HEY or IMAP raw message).
- `extractBody` and `extractHtml` are now thin wrappers around a shared `extractParts`, which decodes each MIME part at most once.
- A new bounded, in-memory LRU (`cache/RenderCache.js`, 12 entries, pure and immutable like `cache/Cache.js`) remembers what `Html.sanitize` did with a message's markup, so revisiting an already-seen message costs no parse at all.
- `select()` now remembers the record the disk cache painted a selection from. When the live Gmail read that follows confirms the cache was already right, recomputing the attachments/invite/unsubscribe and rewriting the cache file are both skipped, and an already-resolved invite is no longer re-fetched — closing the invite-card flicker as a side effect.
- Moving the cursor (j/k) restarts a 70ms timer instead of previewing immediately, so a held j scrolls at full speed and renders once at rest instead of once per row passed through.

### Release Notes

- Opening a message you've already read is now close to instant.
- Newsletters and other large HTML mail open noticeably faster the first time too.
- Fixed: reopening a message with a calendar invitation no longer flickers the invitation card or re-fetches it from Gmail.
- Fixed: non-ASCII text decoded through `decodeBase64Url` (calendar invitations, HEY and IMAP raw messages) no longer comes out mangled.

### Verification

- `make test` (node, shell, and the 81-test qmltestrunner suite) passes.
- `qmllint` on every changed `.qml` file diffs identical to the pre-change run, aside from line numbers the new code shifted.
- Measured against the real QML engine and a real inbox's cached mail: a 144KB message's `extractBody`+`extractHtml` went from 836ms to 9ms; a revisit of an already-parsed 144KB body went from a 293ms miss to a 0ms hit against the render cache.
- Tried live against a real mailbox after a shell restart: no new warnings in the Quickshell log, and the fixes above were confirmed by hand.

### Known limitations

Selecting a message still writes to several distinct properties the reader's body binding depends on, so Qt's `TextEdit` still lays its text out more than once per selection — bounded by Qt's own layout cost (roughly 5-40ms measured), not by anything this PR touches. Fixing it fully would mean restructuring `MessageReader.qml` so its body binding depends on one `Qt.callLater`-coalesced property instead of the ~15 chained ones it reads today, which is a much larger, higher-risk change for a payoff an order of magnitude smaller than what's here; left for a follow-up if it's ever worth it.


RogerHuston:
This was the reason I began looking for other mail clients. Omamail was just very slow. Personally, I would hope that the read view was created on download vs first visit, but this will help.

Issue #239, opened 2026-09-18
Secret Service reconnect pattern (per-op force-close) triggers gnome-keyring-daemon SIGABRT


FaintentDev:
**Title:** Secret Service reconnect pattern (per-op force-close) triggers gnome-keyring-daemon SIGABRT

**Summary**
`src/credentials/secret_service.rs` opens a new D-Bus connection + DH session for every credential operation, then force-closes it with `connection.close()` under a 250 ms deadline and `runtime.shutdown_timeout(Duration::ZERO)`. When this races the `aes_negotiate` handshake, gnome-keyring 50.0 receives a client/session that has already vanished, passes NULL into `g_variant_new`, and aborts the entire daemon (SIGABRT).

The relevant code:

```rust
if let Some(connection) = connection {
    let _ = tokio::time::timeout(CLOSE_DEADLINE, connection.close()).await;
}
runtime.shutdown_timeout(Duration::ZERO);
```

with the comment "This is close, never graceful_shutdown…" acknowledging the non-graceful teardown.

**Impact**
`gnome-keyring-daemon` crashes intermittently (a few times a day) as mail accounts refresh tokens and read passwords. systemd restarts it, so nothing is lost, but it's noisy and — combined with gnome-keyring's own NULL-session assertion bug — is a real daemon crash.

**Suggested fix**
Reuse one long-lived `SecretService`/zbus connection across operations (store it behind a `OnceCell`/mutex, reconnect only on `Error::Unavailable`), and use `graceful_shutdown()` or send an explicit `CloseSession` before dropping, rather than force-closing per op. The current per-op teardown is what's provoking the handshake race.

**Cross-reference**
This is one half of the problem — see the gnome-keyring issue for the daemon-side NULL-session abort.

karamble:
Fixed in #243.

Your diagnosis was exact: the per-operation connection plus the force-close is what races `aes_negotiate`. Measured against a real daemon with `dbus-monitor`, ten `credentials.get` calls inside one backend process cost ten `OpenSession` calls before and one after, five samples each way.

[...]

PR #243, opened 2026-09-18
Fix the Secret Service session being reopened for every credential operation

karamble:
`src/credentials/secret_service.rs` built a tokio runtime, a D-Bus connection and a DH-encrypted Secret Service session for every credential get, put and delete, then tore all three down at once: `connection.close()` under a 250 ms deadline, followed by `runtime.shutdown_timeout(Duration::ZERO)`. The file's own comment conceded the shape of it, "This is close, never graceful_shutdown…".

That teardown races the daemon's handshake. gnome-keyring 50.0 handles a client that vanishes mid-negotiation by passing NULL into `g_variant_new`, which GLib treats as programmer error, so the daemon holding every credential on the machine aborts. It has done so 13 times on my machine since 15 September, always with the same four lines in the journal inside the same second:

```
gkd_secret_service_get_pkcs11_session: assertion 'client' failed
aes_negotiate: assertion 'session' failed
gkd_secret_service_publish_dispatch: assertion 'client' failed
GLib: g_variant_new: expected GVariant of type 'v' but received value has type '(null)'
```

systemd restarts the daemon, but the login keyring comes back locked, so the visible result is mail that does not load. Omamail's own `errors.json` recorded `credentials.get` failing 13 times over those days, with `mail.check`, `imap.list`, `imap.trash`, `imap.modify` and `reader.open` failing behind it; at the abort on 17 September at 19:48:19, Omamail's calls failed at 19:48:22 and 19:48:23.

The daemon is wrong to abort — a vanished client is a recoverable condition, and the same crash is reported against several other projects — but Omamail is the one provoking it, once per credential operation.

So the connection and the session are now opened once and kept. `Shared` holds a process-lifetime runtime and the session behind a mutex; `run_on` takes it, runs the operation, and leaves it alone. Measured against the real daemon with `dbus-monitor`, ten credential reads inside one backend process:

```
omamail 0.10.5:  OpenSession = 10, 10, 10, 10, 10   (five samples)
this branch:     OpenSession =  1,  1,  1,  1,  1
```

Five things were needed to make that safe rather than merely fewer sessions.

**The runtime outlives the operation.** One worker thread, so credential operations still run alongside each other as they did when each built its own runtime; a current-thread runtime would serialise every one of them behind whichever holds it, and a stuck prompt would then block them all for the whole deadline. I checked the obvious worry and it is not one: a session left idle for three seconds between operations is still usable, because a call is always awaited inside `block_on`.

**A write is never repeated.** The session is dropped and the operation retried when a call fails, because a session the daemon has forgotten fails like an absent one, and that is what makes a daemon restart invisible to a read. Only a read is repeated. A request already written to the socket cannot be retracted, and the daemon may have committed it, so repeating `CreateItem` — called with `replace = false` — would leave two items carrying identical attributes, after which `find` answers `Ambiguous` and Omamail can no longer read *or* delete that credential. A sign-out would silently leave a live refresh token in the keyring.

**A locked keyring is an answer, not a broken connection.** `Error::Unavailable` was doing double duty: a locked item, a locked default collection and a failed `get_secret` all produced it alongside genuine transport failures. Retrying on all of them would have renegotiated a session on every read against a keyring that was merely locked — the exact churn this change removes, in the one state where the daemon is least able to survive it. A private `Fault` enum now separates an answer the daemon gave from a call that did not complete, and only the latter drops the session.

**An evicted session is not closed.** Operations run on their own `spawn_blocking` threads and share one connection, so closing it on one thread's failure would terminate another thread's in-flight call — and that call would then look like a transport failure. Eviction now only removes it from the slot; the last `Arc` to be released closes it, once nobody is using it. This is also why `Session` no longer holds a connection handle of its own: `SecretService` owns the connection it was built on.

**One deadline covers the whole operation.** Opening the session and running it now share a single budget computed once, rather than each carrying its own `timeout(deadline, …)`. Without that, one credential operation could wait two deadlines on the happy path and four with a retry, which would have quietly eaten the headroom in `calendar/mod.rs`'s 24 s CalDAV timeout.

A poisoned mutex is also no longer fatal. The slot is written only after the connection is built, so it guards nothing a panic could break, and refusing it would have disabled every credential read for the life of the process. `shared()` likewise no longer remembers a failed runtime build for ever.

The timeout path still drops the session, so `credentials_native_linux_never_completing_prompt_releases_worker_and_connection` keeps its guarantee that a stuck prompt leaves nothing registered on the bus. That now happens when a session has already failed rather than after every operation, so it no longer races the handshake of the session that follows it. A timeout is not retried: the deadline is the caller's bound and a second attempt would exceed it.

Both pre-existing tests are unchanged and still pass. `run_with` is kept for them as a test-only one-shot that disposes its session immediately, which is what they observe.

This is the Omamail half of #239.

## Verification

`LC_ALL=C make validate` green on this commit (see the locale note below). `cargo test --locked --features integration-test-credentials` green: 533 unit, 32 cli and the smaller suites, 0 failed. The Linux credential gate, `cargo test --lib -- --ignored credentials_native_linux`, passes 8/8 against a real `dbus-daemon` and a real keyring.

I watched every new test fail against the unfixed code rather than assuming it would:

| Test | Against the fault |
|---|---|
| `…_reuses_one_session_across_operations` | fails `left: 3, right: 1` with the session cache bypassed |
| `…_a_failed_write_reaches_the_daemon_once` | fails on 2 `CreateItem` calls where 1 is correct, with the retry allowed on writes |
| `…_a_locked_item_keeps_the_session` | fails `left: 3, right: 1` with a locked item classed as a transport failure |
| `…_a_stalled_connect_costs_one_deadline` | fails "waited 603.7ms for a 300ms deadline" with a per-stage timeout |

Driven by hand against the running daemon on Arch with gnome-keyring 1:50.0-1, using `dbus-monitor` to count `OpenSession` on the session bus while a backend process served ten `credentials.get` calls for an account that does not exist. Five samples each: 10/10/10/10/10 before, 1/1/1/1/1 after. Starting the backend and sending nothing produces 0 on both, so every call counted came from the reads.

**One thing I could not explain, and would rather flag than bury.** Under artificial full CPU saturation — a busy loop pinned to every core — `credentials_native_linux_never_completing_prompt_releases_worker_and_connection` fails on this branch and passes on `main`, reproducibly: 0/5 against 5/5, and 5/5 on this branch once the load is removed. Instrumenting it shows `connect` completing in 3.5 ms and the `OpenSession` negotiation not completing inside the test's 150 ms budget. I tried a current-thread runtime, a single `block_on` covering negotiation and operation rather than two, and both combinations of those with and without the lock held across the negotiation; none of them changed it, so my explanations for it were wrong and I am not offering another. It does not reproduce at any load I can create short of saturating every core, `make validate` and the full gate are green, and production's budget is 5 s rather than 150 ms against a negotiation measured at roughly 50 ms. Still, it is a real difference from `main` and somebody should look at it with fresher eyes than mine.

A fresh agent with no context on the change reviewed the diff against `AGENTS.md` and found five faults, all fixed here: the repeated write, the misclassified locked keyring, the eviction that closed a connection other threads were still using, the poisoned-mutex lockout, and the doubled deadline. It demonstrated the first with its own `dbus-daemon` fixture, showing two `CreateItem` calls for one `Put` and a subsequent `get` and `delete` both answering `Ambiguous`, and the fourth and fifth with runnable probes.

Its security verdict on the revision it read was **BLOCK**, on that repeated write alone: no secret reaches argv, logs or a `Debug` impl, and sharing one session across accounts crosses no boundary that was not already crossed, but a credential that cannot be deleted cannot be revoked. It named three conditions to reach PASS — scope the retry to `Operation::Get`, stop classifying a locked item and a locked collection as transport failures, and add a write regression test that fails if the daemon sees a second `CreateItem` for one call. All three are met here. It has not re-reviewed the result, so that is my claim rather than its verdict.

It also noted that the change lengthens the window in which a DH session key and an open authenticated session exist, from roughly one operation to the process lifetime. That is inherent to sharing a session rather than incidental, and it is written down here because it was written down nowhere else.

Two notes on my environment rather than on this change, both of which reproduce on unmodified `main`. `tests/test_publish.sh` fails under `en_US.UTF-8`: it compares a `sort`ed file list against an ASCII-ordered literal, and `app/CMakeLists.txt` collates before `Cargo.lock` in a UTF-8 locale and after it under `LC_ALL=C`. And `make validate` needs cmake for `test-app-qml`, which `CONTRIBUTING.md` lists as a prerequisite for `make app-build` but not for `make validate`.

No interface change, so no screenshots.

## Release Notes

- The login keyring no longer crashes while Omamail reads or saves an account's credentials, which could leave mailboxes empty until the keyring was unlocked again.


huacnlee (approved):
Reviewed and approved; Security: PASS for the affected Secret Service credential scope.

PR #247, opened 2026-09-19
Fix a credential read that failed at startup never being tried again

karamble:
On this machine every boot left Omamail showing "not connected" and opening on the add-mailbox view, with three configured IMAP mailboxes sitting in `accounts.json` the whole time. Restarting the plugin fixed it, every time, until the next boot.

The cause is a single credential read that fails and is never repeated. Matching `journalctl --list-boots` against the app's own `diagnostics/errors.json`:

| Boot | `credentials.get` failure | Delay |
|---|---|---|
| 09-17 18:54:57 | 18:55:17 | +20s |
| 09-17 21:06:23 | 21:06:42 | +19s |
| 09-18 09:17:04 | 09:17:25 | +21s |
| 09-18 10:24:34 | 10:24:47 | +13s |
| 09-18 11:18:42 | 11:18:54 | +12s |
| 09-19 08:37:52 | 08:38:11 | +19s |

Six boots, six failures, one each. gnome-keyring's daemon was already running thirteen seconds before the failure on the last of those, so the daemon starting late is not the explanation; the collection still being locked is the likely one, and it is the kind of condition that ends by itself a moment later.

What makes it stick is what happens next, which is nothing. The backend sat at 0.44 s of CPU over the following fifteen minutes: it was not retrying and failing, it was not being asked at all. `restoreSession()` has one production caller, `MailAccount.qml`'s `onAuthChanged`, and every fetch path is gated behind `auth.loggedIn`, so a mailbox that failed the read cannot reach the code that would ask again.

Both providers already classify the failure correctly. `tst_credential_rpc.qml`'s `test_store_failure_is_retryable_and_not_reported_as_missing` says so in its name and asserts that a second `withCredentials` does re-read. The missing half is that nothing performs that second call. So this schedules it: a store failure arms a timer and the read is repeated on its own, doubling from five seconds and capped at five minutes. The cap is the part that matters — a keyring unlocked an hour after login should still bring the mailbox back without a restart.

`storeRetryDelay` lives in `Credentials.js` beside the other per-provider credential helpers, and delegates to `OAuth.js`'s `refreshRetryDelay` as `MicrosoftOAuth.js` does, so the cap has one definition rather than three.

**Only a store that did not answer is repeated.** A duplicate keyring entry, a malformed key, or a backend too old to hold credentials are refusals that stay refused, and retrying those forever would hide a mailbox that needs the user behind one that merely looks busy. `AuthManager` draws the same line and says why: "A revoked or expired grant is never coming back."

**A retry that outlives its purpose is the other hazard**, so it is disarmed from four directions. Any read that settles clears it, including one that settles as "no password saved". Signing out clears it, or a timer armed before the credential was deleted would fire afterwards and ask the user to sign in again. Changing the mailbox clears it, which matters more than it looks: `Service.qml`'s accounts `Instantiator` is modelled on a count rather than the array, so removing an account rebinds a surviving delegate to a different one instead of destroying it, and a retry armed for the mailbox that left would fire against the one that took its place.

And signing in clears it, which is the one that bites. The user's answer to "not connected" is to retype the password. A read racing that sign-in is not ordered against the write it races — `rpc.rs` dispatches each request to its own `spawn_blocking` task — so it can hand back the password the store still held and overwrite the one just verified against the server, with nothing reporting it. `OutlookAuth`'s `enqueueKeyringJob` exists for this class of race; `ImapAuth` and `JmapAuth` have no serializer because until now they never issued a read after a write.

Rather than rely on having found every route by which a credential can arrive, the timer itself is guarded: it reads only while the provider holds no credential **and** no sign-in is being verified. The second half is needed because `password` stays empty until the server answers, so the credential-held test alone would let a read land in the middle of a sign-in.

**Scope.** `AuthManager` (Google) and `OutlookAuth` are deliberately untouched. They have the same symptom from a different cause: both discard the error argument entirely — `handleSecretLookup(error ? "" : value)` — so a store failure is indistinguishable from "no token saved", and the account is marked signed-out rather than unread. Fixing that means deciding what those providers should believe when the store is silent, which is a larger change and belongs in its own pull request. This one is confined to the two providers that already draw the distinction and simply never act on it.

## Verification

**Confirmed on a real boot**, which is the reason to trust this rather than the tests. With the change installed, the backoff is legible in the app's own diagnostics:

```
11:47:15  backend starts
11:47:17  credentials.get fails
11:47:22  retried  (+5s)   fails
11:47:32  retried  (+10s)  fails
11:47:52  retried  (+20s)  fails
          retried  (+40s)  succeeds — no further credential errors
11:50:21  all three mailboxes fetched
```

Five, ten and twenty seconds are `storeRetryDelay(0)`, `(1)` and `(2)`. The same boot before the change produced one failure and then silence, with the mailboxes empty until the plugin was restarted by hand. Arch, gnome-keyring 1:50.0-1, three IMAP accounts.

`LC_ALL=C make validate` green on this commit. (The `LC_ALL=C` is for `tests/test_publish.sh`, which compares a `sort`ed file list against an ASCII-ordered literal and so fails under `en_US.UTF-8` on unmodified `main` too — unrelated to this change.) `node ui/tests/test_credentials.js` covers the delay rule, including that a negative or non-numeric attempt is a first attempt rather than a zero-length wait that would spin.

Seventeen QML tests pass in `tst_credential_rpc.qml`. Every new behaviour was watched failing against the code without its fix, not assumed:

| Test | Against the fault |
|---|---|
| `…_store_failure_is_tried_again_with_nobody_asking` | fails on the armed assertion with the scheduling removed; the file also drops from 5.3 s to 16 ms, because the retry never fires |
| `…_jmap_store_failure_arms_the_same_retry` | same |
| `…_signing_in_disarms_a_pending_retry` | fails with the sign-in disarm removed |
| `…_a_permanent_refusal_arms_nothing` | fails with the transient-error test removed |
| `…_changing_the_mailbox_disarms_the_retry` | fails with the `onAccountIdChanged` disarm removed |

`test_a_missing_credential_arms_nothing` is the safety assertion rather than a regression test: if "missing" armed a retry, every mailbox that was never signed in would poll the keyring for the life of the session.

Driven by hand as well: the changed files copied into the installed plugin, `qmllint` clean, shell restarted, all three mailboxes fetching four to six seconds after the backend started, with no errors after startup.

A fresh agent with no context reviewed the diff against `AGENTS.md`. It confirmed the diagnosis by tracing it independently, and found five faults I had not: the sign-in race above, the missing disarm on mailbox change, no terminal state for permanent refusals, the hole in the guard while a sign-in is verified, and `storeRetryDelay` being a third copy of a curve this repo delegates for. All are fixed here. Its security verdict is **PASS** — nothing new reaches argv, no secret reaches a label or log, and the secret-bearing callback is not retained — and it answered the prompt-storm question from the source: `search_items` returns on a locked result and `unlock()` is never called, so this path cannot raise a dialog. It could not run the QML suite in its environment and reported that as NOT RUN rather than as a pass.

No view file changed, so no screenshots — though the user-visible effect is which card the mailbox opens on.

## Release Notes

- A mailbox no longer stays disconnected for the rest of the session when the keyring is not ready as Omamail starts. The credential is read again on its own, so mail loads once the keyring is available instead of only after restarting the plugin.

### Review fix verification

On `a1b1df55f2209c5e8506a5018f6b65dd961c656c`, the credential retry now remains armed when its timer comes due during an in-flight IMAP or JMAP sign-in, so a failed manual sign-in cannot silently discard recovery from the original keyring outage. The new QML regression was observed failing before the fix (`17 passed, 1 failed`) and passing after it (`18 passed, 0 failed`). `node ui/tests/test_credentials.js`, `bash tests/test_source.sh`, `make qml-check`, and the complete QML suite pass; the QML run reports 971 tests passed with none failed.


huacnlee (approved):

PR #249, opened 2026-09-19
Fix new-mail notifications dropped by an undefined bar colour

artemisa81:
## Problem

On the Omarchy shell plugin path (not the standalone app), new-mail notifications never appear, and the shell log fills with one warning per account on every start:

```
WARN scene: .../omamail/ui/Service.qml[2519:7]: Unable to assign [undefined] to QString
```

## Cause

`ui/Service.qml` reads the bar palette from `root.shell.bar.barForeground`. The scoped shell handed to a plugin's service exposes `bar` as `PluginBarStateApi`, a **scalar window-positioning view** with only `barHidden`, `barSize`, `fontFamily` and `position` — it deliberately does not carry `barForeground` (see its own header comment: "The active Bar QObject is never retained here"). `barForeground` lives on `qs.Ui.PluginBarApi`, which is given to *bar widgets*, not to the service.

So `barForeground` is `undefined`, the condition `root.shell && root.shell.bar` is still true, and the undefined value is assigned to `MailAccount.notificationForeground`, a `string` property. The assignment fails, the property stays `""`, and `scripts/notify-mail.py` — which requires `#RRGGBB(AA)?` and `return 2`s otherwise — silently declines to send the notification.

## Fix

Guard the missing property and fall back to a real colour:

```qml
notificationForeground: root.shell && root.shell.bar
    && root.shell.bar.barForeground !== undefined
  ? root.shell.bar.barForeground : Color.foreground
```

With the fallback, `notificationForeground` is `Color.foreground` (a valid colour, which converts to a string the same way `notificationAccent: Color.accent` already does), so `notify-mail.py` accepts it.

## Verification

- Before: the warning is logged on every load, 360 times in one day here; `python3 scripts/notify-mail.py "" "" -- t b` returns 2 (dropped).
- After: no warning is logged on shell restart, and the fallback colour reaches the script.

A longer-term option is for the shell to expose `barForeground` (or the palette) on `PluginBarStateApi`, but the guard is enough to stop dropping notifications and is safe regardless of whether that surface grows.


huacnlee:
Superseded by #256, which fixes the same scoped-shell notification failure with a regression covering the restricted bar API and current before/after visual evidence.

PR #254, opened 2026-09-19
chore: Make the publish file-order check locale-independent

artemisa81:
`tests/test_publish.sh` compared a byte-ordered expected file list with `sort` under the caller's locale. Under a UTF-8 collation, `app/CMakeLists.txt` sorts before `Cargo.lock`, so the check failed even though the release commit contained exactly the intended files.

Pinning this one comparison to the C collation makes the assertion deterministic without changing the publish workflow.

## Verification

`LC_ALL=en_US.utf8 bash tests/test_publish.sh` and `LC_ALL=C bash tests/test_publish.sh` both pass on `13394252ad57184634b6bfe105d2042092578094`.


huacnlee (approved):

PR #256, opened 2026-09-19
Fix mail notifications with scoped shell APIs

Skeptomenos:
Omarchy's scoped shell exposes a bar object without `barForeground`. `Service.qml` selects that missing property and passes an empty foreground to `notify-mail.py`. The helper rejects it, so new-mail notifications never reach the desktop.

Bind the notification foreground to the existing `Color.foreground` singleton, as the adjacent accent binding already does. The regression supplies the restricted bar and verifies the real notification arguments before and after a theme change. On older hosts with a custom bar foreground, notification icons now follow the theme foreground.

To reproduce, enable new-mail notifications under a scoped Omarchy host and receive an unread message. The affected helper exits 2 without displaying a notification.

## Verification

`make validate` passed on ARM64 Linux with Qt 6.11.2, including 602 Rust tests and 961 native-backed QML tests. The focused notification suite passed 10 tests. Against the parent binding, it produced the expected empty-foreground failure, with 9 tests passing and 1 failing.

A fresh independent agent reviewed the diff and found no blocking bugs. Its custom-bar-color observation is documented above.

For the live checks, real Service/account components generated both command vectors using synthetic mail. The unchanged production helper rejected the parent vector and displayed the candidate notification in Retro 82 dark and Catppuccin Latte light, both at scale 1.3333334. Title-specific dismissal completed without a message action. The light-theme check restored the original theme and preserved the user shell configuration.

The dark-theme comparison redacts unrelated desktop content:

Before: no notification. After: the synthetic Omamail notification appears with its themed icon.

The light-theme crop shows only the synthetic notification:

The complete Omamail icon, title, subject, and body in Catppuccin Latte.

Live verification covers the notification command path. End-to-end mailbox arrival remains unverified.

## Release Notes

- Fix missing new-mail notifications when Omarchy provides a scoped shell API.


huacnlee (approved):

PR #257, opened 2026-09-19
Fix a cached read restoring the unread mark on a message read a moment ago

Depodra:
Opening a conversation member the reader had already marked read made the rail draw its stop unread for one round trip, then read again, and sent a second `Email/set` marking it read. Seen on a JMAP account (Stalwart) while traversing the rail: open an unread member, step to another stop, come back, and the stop flashes. It stops once a live read of the message completes without being cancelled by the next step, which is what made it look intermittent and self-healing.

The reader opens a message from its file before the network answers (`reader.open` with `cacheOnly`), and the file is the resource as the server answered the live read that first opened the message — from before the quiet mark-read that opening sends. Nothing after that touches the file: the mark-read, a star, a move all go through the optimistic view into `messages` and `memberSummaries`. But `select`'s callback painted the cached copy's summary as it came — into `selectedMessage`, through `rememberMember` into the member store the rail draws from, and past `marksReadOnArrival`, which saw `unread: true` and marked the message read again. The list row was already protected, since `messages` is rewritten only by the live copy; the member store and the open summary were not.

The fix is one rule in `Model.js`, `cachedDetailSummary`: a detail read answered from disk takes the labels the account already holds for the message — its row or its member summary — and, when the account holds one, its block, since a representative's block is recomputed from its members as they are marked while the file's is from before. The flags then follow `rowWithThread`'s rule, the conversation's as well as the labels', which is what a live read of the same message carries (a JMAP read of a representative gets the block attached in `jmap/read.rs`). A message with nothing in hand, opened from a notification, is painted from the file as before. The live copy that follows is not touched: it is the server's answer and still wins, as it does today. A live copy answered before the first opening's mark-read has settled is a narrower window of the same shape — one round trip, closed by the time another message has been opened — and stays with the existing `keepRead` rule, which covers it only within the opening that sent the mark. The label-to-flag mapping `applyLabelChange` carried inline is now `withLabels`, shared with the new rule and unchanged in what it computes (the Rust parity test in `model_tests.rs` still compares the two).

## Verification

`TZ=UTC LC_ALL=C make validate` green on `ca5616b` (Qt 6.11.2 on Arch, cmake from a venv): 532 Rust tests in the main crate, 965 QML. New `ui/tests/qml/tst_cached_read_member_flags.qml` drives a real `MailAccount` with `reader.open` scripted the way the backend answers it, through opening an unread member from the rail, the quiet mark-read, a step to the representative, and a reopen answered from a cached copy that still says unread. Run against `main` with only the tests applied, both cases fail: the member store and the projected stop read unread on the way (`Actual: [stop,stop,store]`, `Expected: []`) and a second mark-read goes out; and the representative opened from its file, whose block still says a reply is unread, reads unread and is marked read again (`Actual: true, Expected: false` on the open summary's `unread`). `ui/tests/test_model.js` covers the rule itself and fails on `main` (`cachedDetailSummary is not a function`). Live: on the JMAP account (Stalwart) where it was found, running this machine's build with the fix merged, the reported flow — open an unread member of a thread from the rail, step to another stop, come back — no longer flashes; hand-tested 2026-09-20.

Screenshots: rendered offscreen from the plugin's own `App.qml` with synthetic mail, a real `MailAccount` on a scripted reader inside the real `Service`, this machine's theme colours and base font over the repository's test stubs, at 1400×900; the optimistic view of the mark-read held until the frame is taken, as the intent's round trip holds it. Same checkout, the fix stashed for the first frame.

Before: the reply just read is opened again from its cached copy; its stop and the caption go unread until the network answers

After: the same moment; the stop stays read and nothing is marked read again

## Release Notes

- Fixed: a message opened again after being marked read no longer flashes unread in the conversation rail, and is no longer marked read a second time.


huacnlee (approved):

PR #271, opened 2026-09-21
Switch to a Gmail row before opening the shared OAuth client setup page

en3r0:
## Problem

Settings' "Google OAuth client" section ("Shared by every mailbox
above") opens `SetupPage.qml` through `openClientSetup()`, which pushed
straight to the setup page without touching which account was active.
`SetupPage.qml` reads `root.service.auth`, and `Service.qml`'s `auth`
property is `current.auth` — whichever account the window happens to
be showing.

With a saved IMAP mailbox and no Gmail one yet, `current` stayed on the
IMAP row, so the Gmail-branded setup page rendered bound to that row's
`ImapAuth`. `ImapAuth` has no `saveCredentials`, so pressing the button
in the OAuth client form's Save step threw immediately: no browser
opened, no error shown, nothing happened.

Reproduced on a real machine via `journalctl`:

```
TypeError: Property 'saveCredentials' of object ImapAuth_QMLTYPE_517 is not a function
  at SetupPage.qml:72
```

Add mailbox and Edit already avoid this by switching the active account
before pushing a setup page (`chooseProvider`, `editAccount`). This
gives `openClientSetup` the same discipline.

## Solution

- `openClientSetup()` now finds a saved Gmail account and switches to
  it before pushing the setup page, or falls back to
  `chooseProvider("gmail")` to draft one the way Add mailbox would when
  none exists yet.
- The lookup itself (`Service.gmailAccountId()`) lives in `Service.qml`
  rather than inline in `App.qml`, which was already within 158 bytes
  of `tests/test_source.sh`'s 128 KiB per-file ceiling and had no room
  for an inline loop plus explanation.

## Verification

`make validate` green: `test-rust`, `test-js`, `test-shell` (`LC_ALL=C
make test-shell`), `test-qml` (`make test-qml` — 64/64 QML tests plus
the native html-output regression), `test-app-qml` (cmake build of the
standalone app + `tst_standalone_composition`), `qml-check` (`qmllint`
on both changed files — 348 warnings before and after this change,
none new), `omarchy plugin validate .`, `git diff --check`.

The new regression, `tst_app_client_setup_wrong_account.qml`, was
watched failing against the code without this fix — one saved IMAP
account, no Gmail one, `app.openClientSetup()` from Settings — with
`mailService.current` never leaving the IMAP host:

```
FAIL!  ... 'the active account moved off the IMAP row onto a Gmail one' returned FALSE.
```

With the fix, the same test passes: the active account switches to a
fresh Gmail row, `mailService.auth.saveCredentials` is a function (not
ImapAuth's missing one), the Gmail setup page is showing, and the IMAP
mailbox is confirmed untouched — still first in the account list, its
own auth's `loginBusy` still false.

Driven against the real backend and account/service objects (the same
`Omamail.Service` + `BackendFixture` pattern `tst_app_edit_connect.qml`
uses for an adjacent wrong-page bug), not a fully stubbed service, so
the fix is checked against the real per-provider Loader that builds
each account's auth object.

## Release Notes

- "Set up..." on the shared Google OAuth client section in Settings no
  longer silently does nothing when a non-Gmail mailbox is active; it
  switches to (or creates) a Gmail mailbox first, the way Add mailbox
  and Edit already do.


huacnlee:
I pushed `8f98b1a` to cover the existing-saved-Gmail routing branch as part of review; the focused QML test now passes both the create-draft and switch-existing-account paths (4/4).

The remaining review gate is UI evidence: please add clearly labeled, comparable before/after screenshots for the current head, using the same Settings → Google OAuth client flow, window size, theme, and scale, plus a brief description of the visible difference. Once those are attached, please request review again.

